PR #2, opened 2016-03-23
Initial asset rendering command

aaronlevy:


Makefile
```diff
@@ -1,11 +1,22 @@
+export GO15VENDOREXPERIMENT:=1
```
dghubble: You might find parts of https://github.com/coreos/coreos-baremetal/blob/master/test useful.

pkg/assets/assets.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/coreos/bootkube/pkg/assets/internal"
+)
+
+const (
+	AssetPathCAKey           = "ssl/ca.key"
```
dghubble: Can we generate the same asset names between bootkube, bootcfg, and coreos-kubernetes? botocfg follows the same conventions as coreos-kubernetes, which differs from these. https://github.com/coreos/coreos-baremetal/tree/master/examples#assets

aaronlevy: I wanted to change this because I always disliked calling everything `-key.pem`. fwiw `.key` `.crt` are more universal -- but I could be convinced to switch back...

dghubble: I'd be for us changing that everywhere then. The `.pem` suffix seems like a poor choice anyway. Of course its PEM format. One other difference I just noticed is `ssl` dir vs `tls`.

As long as we have the same names, the skew between systemd units / k8s manifests / Ignition/Cloud-config can be minimal.

aaronlevy: Yeah, I'm on board for changing across projects (just hadn't gotten around to it elsewhere -- somewhat low priority). I'm indifferent towards tls/ssl fine to change it (I refer to it as tls everywhere else in code).

cmd/bootkube/start.go
```diff
@@ -0,0 +1,59 @@
...
+	cmdRoot.AddCommand(cmdStart)
+	cmdStart.Flags().StringVar(&startOpts.SSHUser, "ssh-user", "", "Username for the remote ssh host")
+	cmdStart.Flags().StringVar(&startOpts.SSHKeyFile, "ssh-keyfile", "", "Keyfile for the remote ssh user")
+	cmdStart.Flags().StringVar(&startOpts.RemoteAddr, "remote-address", "", "Remote ssh host (ip:port)")
+	cmdStart.Flags().StringVar(&startOpts.RemoteEtcdAddr, "remote-etcd-address", "127.0.0.1:2379", "Remote etcd host (ip:port)")
+	cmdStart.Flags().StringVar(&startOpts.ManifestDir, "manifest-dir", "", "Path to Kubernetes object manifests")
+}
+
+func runCmdStart(cmd *cobra.Command, args []string) error {
+	if err := validateOpts(startOpts); err != nil {
```
derekparker: nit: Could just inline this in a PreRun on the command.

cmd/bootkube/render.go
```diff
@@ -0,0 +1,61 @@
...
+	}{}
+)
+
+func init() {
+	cmdRoot.AddCommand(cmdRender)
+	cmdRender.Flags().StringVar(&renderOpts.OutDir, "outdir", "", "Output path for rendered manifests")
+}
+
+func runCmdRender(cmd *cobra.Command, args []string) error {
+	if renderOpts.OutDir == "" {
```
derekparker: nit: Could do validation in a PreRun. (minor nit)

pkg/assets/assets.go
```diff
@@ -0,0 +1,67 @@
...
+//go:generate gofmt -w internal/templates.go
+
+type Asset struct {
+	Name string
+	Data []byte
+}
+
+type Assets []Asset
+
+func (a Assets) Append(asset ...Asset) Assets {
```
derekparker: This method isn't strictly necessary. You can use the builtin `append` function on the `Assets` type.

pkg/assets/assets.go
```diff
@@ -0,0 +1,67 @@
...
+	return a
+}
+
+func (a Assets) Get(name string) (Asset, error) {
+	for _, asset := range a {
+		if asset.Name == name {
+			return asset, nil
+		}
+	}
+	return Asset{}, fmt.Errorf("asset '%s' does not exist", name)
```
derekparker: style nit: can just use `%q` to quote the string instead of `'%s'`.

pkg/assets/assets.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (a Assets) WriteFiles(path string) error {
+	if err := os.Mkdir(path, 0755); err != nil {
+		return err
+	}
+	for _, asset := range a {
+		f := filepath.Join(filepath.Join(path, asset.Name))
```
derekparker: Accidental nesting of `filepath.Join` calls?

pkg/assets/secrets.go
```diff
@@ -0,0 +1,89 @@
...
+	apiKey.Data = b64Byte(apiKey.Data)
+
+	apiCert, err := assets.Get(AssetPathAPIServerCert)
+	if err != nil {
+		return secret, err
+	}
+	apiCert.Name = filepath.Base(apiCert.Name)
+	apiCert.Data = b64Byte(apiCert.Data)
+
+	config := struct {
```
derekparker: Maybe promote this struct to `TemplateConfig` since it's used in multiple places.

pkg/bootkube/bootkube.go
```diff
@@ -44,16 +45,17 @@ type bootkube struct {
...
 
-func NewBootkube(opts Opts) (*bootkube, error) {
+func NewBootkube(config Config) (*bootkube, error) {
 	apiServer := apiserver.NewAPIServer()
 	fs := pflag.NewFlagSet("apiserver", pflag.ExitOnError)
 	apiServer.AddFlags(fs)
 	fs.Parse([]string{
 		"--etcd-servers=http://" + localEtcdServerAddr,
 		"--service-cluster-ip-range=10.3.0.0/24",
+		"--runtime-config=extensions/v1beta1/deployments=true,extensions/v1beta1/daemonsets=true",
```
derekparker: We should also add `--allow-privileged`.

pkg/tlsutil/tlsutil.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return x509.ParseCertificate(certDERBytes)
+}
+
+func NewSignedCertificate(cfg CertConfig, key *rsa.PrivateKey, caCert *x509.Certificate, caKey *rsa.PrivateKey) (*x509.Certificate, error) {
+	ips := make([]net.IP, len(cfg.IPAddresses))
+	for i, ipStr := range cfg.IPAddresses {
+		ips[i] = net.ParseIP(ipStr)
```
derekparker: nit: I know the IPs are hard coded for now, but this function can return `nil` on an invalid representation of an IP that it cannot parse. Would be useful to print a warning or something on invalid IPs. (Can be done once we start allowing user-provided IP addrs).

pkg/assets/secrets.go
aaronlevy: I think I'm gonna drop this code and just create a secret struct & marshal directly. Too much weirdness here for not much benefit.

cmd/bootkube/start.go
aaronlevy: Didn't even know that existed. Sweet.

pkg/tlsutil/tlsutil.go
aaronlevy: Added a TODO for now because i'm lazy, but agree that we need this before moving away from the hardcoded list.

aaronlevy:
Pretty much a re-write. Mind taking a look @dghubble @derekparker

hack/single-node/user-data.sample
```diff
@@ -0,0 +1,55 @@
...
+        content: |
+          [Unit]
+          Requires=flanneld.service
+          After=flanneld.service
+    - name: kubelet.service
+      enable: true
+      command: start
+      content: |
+        [Service]
+        ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
```
derekparker: Should be able to get rid of this since we're not using static manifests.

derekparker:
LGTM

PR #4, opened 2016-03-30
Add kubelet template

aaronlevy:
/cc @derekparker


dghubble:
I'm trusting kubelet to behave with dev >.>
LGTM

PR #6, opened 2016-04-05
render: control-plane manifest tweaks

aaronlevy:


pkg/assets/templates.go
```diff
@@ -113,12 +99,16 @@ type secret struct {
 	Data       map[string]string `json:"data"`
 }
 
-func newSecretYaml(name, namespace string, data map[string][]byte) ([]byte, error) {
-	b64data := make(map[string]string)
-	for k, v := range data {
-		b64data[k] = base64.StdEncoding.EncodeToString(v)
+func newSecretFromAssets(name, namespace string, assetNames []string, assets Assets) ([]byte, error) {
+	data := make(map[string]string)
+	for _, an := range assetNames {
```
derekparker: You should be able to range over assets directly here.

derekparker: Unless you want to be able to pass in all assets but only create secret for a subset, which upon further inspection seems like what you're doing, so please ignore.

derekparker:
LGTM

PR #12, opened 2016-04-08
cmd/bootkube: Allow config of generated manifests

derekparker:
This patch adds a few flags to configure the generated assets. We may
want to eventually move to something like configMaps or something in the
future, but for now this gets us enough configuration.


aaronlevy:
lgtm

PR #13, opened 2016-04-08
assets: remove nodeSelector on kube-proxy

derekparker:
The kube-proxy daemonset should not be limited to master nodes.


aaronlevy:
lgtm

PR #14, opened 2016-04-08
pkg/bootkube: Add admission-control=ServiceAccount flag

derekparker:
We have all the keys to generate service accounts, but we're not
enabling the plugin.


aaronlevy:
lgtm

PR #15, opened 2016-04-08
hack: Add multi-node hack setup

derekparker:
This patch adds everything needed for a simple self-hosted multi-node
hack/dev cluster. It is opinionated to match the default vagrant VM
setup, but can be optionally configured.


aaronlevy:
LGTM. I like the helpers you've made for the config/render/start stuff -- we should try and do the same with single node too.

PR #18, opened 2016-04-13
hack: tweak helper scripts & add to single-node

aaronlevy:


hack/multi-node/bootkube-render
```diff
@@ -13,7 +13,4 @@ echo "inserting kubeconfig into user-data"
 cat user-data.sample > cluster/user-data && sed 's/^/      /' cluster/auth/kubeconfig.yaml >> cluster/user-data
 cp cluster/user-data{,-worker}
 cp cluster/user-data{,-controller}
-sed -i.bak \
-    -e 's/127\.0\.0\.1/172.17.4.101/' \
-    -e '/^.*master.*/d' \
-    cluster/user-data-worker
+sed -i '' -e '/--node-labels=master=true/d' cluster/user-data-worker
```
derekparker: The `-i ''` is causing errors. Changing it to just `-i` with no backup extension specified works.

aaronlevy: Hmm. This might be an osx vs nix thing. FWIW, allowing the backup file should work in both - so maybe we just go that route.

derekparker: Yeah, I think you're right. I'm with with keeping it to `-i.bak` or something and them `rm`ing that file afterwards.

derekparker:
LGTM

PR #19, opened 2016-04-18
Have bootkube self-shutdown once control-plane running

aaronlevy:


pkg/bootkube/create.go
```diff
@@ -0,0 +1,132 @@
...
+	client, err := clientset.NewForConfig(&restclient.Config{Host: localAPIServerInsecureAddr})
+	if err != nil {
+		return err
+	}
+
+	_, err = client.Discovery().ServerVersion()
+	return err
+}
+
+func reasonForError(err error) unversioned.StatusReason {
```
derekparker: Unused func.

pkg/bootkube/status.go
```diff
@@ -0,0 +1,123 @@
...
+	"k8s.io/kubernetes/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+const (
+	appKey       = "k8s-app"
+	doesNotExist = "DoesNotExist"
+)
+
+func CheckStatus(pods []string, timeout time.Duration) error {
```
derekparker: nit: Just a naming nit, but feels like `WaitUntilPodsRunning` would be more descriptive than `CheckStatus`, since this function doesn't return statuses and just blocks until all pods are running or timeout is reached.

pkg/bootkube/status.go
```diff
@@ -0,0 +1,123 @@
...
+	sc, err := NewStatusController()
+	if err != nil {
+		return err
+	}
+	sc.Run()
+
+	checkFn := func() (bool, error) {
```
derekparker: refactor nit: Maybe we expose `allRunning` as a (possibly exported) method on the status controller that returns `bool, error`, and move this logic there? That way this just becomes `wait.Poll(5*time.Second, timeout, sc.AllRunning)`.

derekparker:
Mostly just a few nits, otherwise looks great.

pkg/bootkube/status.go
aaronlevy: agreed

pkg/bootkube/status.go
aaronlevy: I'll update pr to do this

derekparker:
LGTM -- only thing I noticed during testing was we get a bunch of EOF errors when it shuts down. We should try for a more graceful shutdown if we can, but doesn't have to be part of this PR. I'd post the logs but I closed the terminal window too soon before thinking about it.

PR #20, opened 2016-04-18
README: just use go get

philips:
Is there a reason for the added complexity of telling someone to clone into their GOPATH vs go getting? Seems like a left over from when this was a private repo.


aaronlevy:
lgtm

PR #22, opened 2016-04-20
pkg/asset: set kubeconfig server from config values

aaronlevy:


aaronlevy:
fixes: https://github.com/coreos/bootkube/issues/24

derekparker:
LGTM

PR #25, opened 2016-04-21
hack: Update READMEs regarding auto-shutdown

derekparker:
This patch simply updates the READMEs in the hack directories, now that
the self-killing bootkube code is merged, we no longer should instruct
folks to manually kill the process.


derekparker:
cc: @aaronlevy

aaronlevy:
lgtm

PR #27, opened 2016-04-22
Api endpoint fix

aaronlevy:
This includes a patched hyperkube built from https://github.com/aaronlevy/kubernetes/commits/v1.2.2_runonce.0 which contains a minor fix so that the downward api exposes the podIP when hostnetwork=true.

Upstream issue: https://github.com/kubernetes/kubernetes/issues/24576


pbx0:
lgtm

PR #28, opened 2016-04-26
pkg/asset/templates: Use podIP in kubelet hostname-override

dghubble:
This change removes the reliance on the node hostname, which was picked up by the kubelet as the default routeable kubelet hostname. The idea is that it is better to explicitly assign a hostname by flag than to have a dependency on the host node's hostname configuration.

I've tested this against the network booted libvirt VM cluster in [coreos-baremetal #170](https://github.com/coreos/coreos-baremetal/pull/170). I haven't validated with the bootkube Vagrantfiles since I don't have a VirtualBox/VMWare setup.

### DIFF
--- a/hack/multi-node/user-data.sample
+++ b/hack/multi-node/user-data.sample
@@ -1,7 +1,5 @@
 #cloud-config
 
-hostname: $public_ipv4
-
 coreos:
   flannel:
     interface: $public_ipv4
@@ -28,6 +26,7 @@ coreos:
       command: start
       content: |
         [Service]
+        EnvironmentFile=/etc/environment
         Environment=KUBELET_ACI=quay.io/aaron_levy/hyperkube
         Environment=KUBELET_VERSION=v1.2.2_runonce.0
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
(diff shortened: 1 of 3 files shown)


aaronlevy:
Can you also drop the `hostname:` setting from the hack user-data.samples?

dghubble:
Sure thing

aaronlevy:
In terms of the `./hack` examples, the host kubelet (`/etc/systemd/system/kubelet`) will also need to set the `--hostname-override` flag. One option might be to set `EnvironmentFile=/etc/environment` so we can set`--hostname-override=$COREOS_PUBLIC_IPV4`

aaronlevy:
More or less the equivalent of your change here: https://github.com/coreos/coreos-baremetal/pull/170/files#diff-84506619fb1284a80df9e204f6dc8e82R56

dghubble:
Hmm, testing this out on OSX/Virtualbox/Vagrant, I need curly braces in the unit. `--hostname-override=${COREOS_PUBLIC_IPV4}`, because of cloud-config.

dghubble:
The hack single-node and multi-node examples work for me.

chancez:
This looks great! 👍

aaronlevy:
lgtm

PR #30, opened 2016-04-28
Add hack dirs to readme and some additional details

aaronlevy:
/cc @philips @derekparker @pbx0

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,11 @@
 
 Bootkube provides an entire kubernetes control plane in a single binary, and includes commands to render and consume assets for bootstrapping a self-hosted kubernetes cluster. Bootkube is expected to be used simply for bootstrapping purposes.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,10 +28,17 @@ To start bootkube use the `start` subcommand:
...
 ```
 
-Bootkube expects a directory containing the manifests to be provided as a command line flag, as well as other TLS assets (all of which can be taken from the `render` command). To see the available flags, run `bootkube help start`. 
+Bootkube expects a directory containing the manifests to be provided as a command line flag, as well as other TLS assets (all of which can be taken from the `render` command). To see the available flags, run `bootkube help start`.
 
 When you start bootkube, you must also give it the addresses of your etcd servers, and enough information for bootkube to create an ssh tunnel to the node that will become a member of the master control plane. Upon startup, bootkube will create a reverse proxy using an ssh connection, which will allow a bootstrap kubelet to contact the apiserver running as part of bootkube.
 
+## Hack
+
+There are currently two reference implementations to demonstrate bootstrapping a cluster. These use Vagrant to launch VM(s) and then bootkube to render assets / launch a self-hosted cluster:
```
dghubble: Include? https://github.com/coreos/coreos-baremetal/blob/master/Documentation/bootkube.md

aaronlevy: Yup, I'll add

derekparker:
lgtm

PR #36, opened 2016-05-16
*: Refactor to run bootkube directly on-host

aaronlevy:
- Changes controller-manager & scheduler to be Deployments without
  host-networking.
- Simplify flags needed for render/start steps
- Updates /hack dir installations to use new on-host method.

/cc @derekparker @dghubble


cmd/bootkube/start.go
```diff
@@ -19,25 +21,29 @@ var (
...
-	cmdStart.Flags().StringVar(&opts.SSHKeyFile, "ssh-keyfile", "", "Keyfile for the remote ssh user")
-	cmdStart.Flags().StringVar(&opts.APIServerKey, "apiserver-key", "", "API server private key")
-	cmdStart.Flags().StringVar(&opts.APIServerCert, "apiserver-cert", "", "API server certificate")
-	cmdStart.Flags().StringVar(&opts.CACert, "ca-cert", "", "CA certificate")
-	cmdStart.Flags().StringVar(&opts.ServiceAccountKey, "service-account-key", "", "Service account key")
-	cmdStart.Flags().StringVar(&opts.TokenAuth, "token-auth-file", "", "token auth csv file")
-	cmdStart.Flags().StringVar(&opts.RemoteAddr, "remote-address", "", "Remote ssh host (ip:port)")
-	cmdStart.Flags().StringVar(&opts.RemoteEtcdAddr, "remote-etcd-address", "127.0.0.1:2379", "Remote etcd host (ip:port)")
-	cmdStart.Flags().StringVar(&opts.ManifestDir, "manifest-dir", "", "Path to Kubernetes object manifests")
+	cmdStart.Flags().StringVar(&startOpts.etcdAddr, "etcd-address", "http://127.0.0.1:2379", "Single etcd node to use during bootkube bootstrap process.")
```
dghubble: In render these command line options end in Server, here it ends in Addr?

cmd/bootkube/start.go
```diff
@@ -19,25 +21,29 @@ var (
...
+	cmdStart.Flags().StringVar(&startOpts.assetDir, "asset-dir", "", "Path to the cluster asset directory. Expected layout genereted by the `bootkube render` command.")
```
dghubble: Use the same flag between `render` and `start`?

hack/multi-node/README.md
```diff
@@ -2,31 +2,19 @@
...
-
-This will render all tls assets, manifests, secrets, and user-data files necessary to stand up a local vagrant based development cluster. If you would like to change configuration for any reason, you can tweak parameters found in the `config-env` file.
-
-## Start VM
-
-```
-vagrant up
-```
-## Start Bootkube
+This will generate the default assets in the `cluster` directroy and launch multi-node self-hosted cluster.
```
dghubble: directory spelling

hack/multi-node/bootkube-up
```diff
@@ -1,17 +1,37 @@
...
 
+un="$(uname)"
+local_os="linux"
+if [ ${un} == 'Darwin' ]; then
+    local_os="darwin"
+fi
+
+# Render assets
+if [ ! -d "cluster" ]; then
+  ../../bin/${local_os}/bootkube render --outdir=cluster --api-servers=https://172.17.4.101 --etcd-servers=http://172.17.4.51:2379
```
dghubble: --api-servers=https://172.17.4.101:443?

hack/multi-node/user-data.sample
```diff
@@ -32,8 +30,8 @@ coreos:
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --runonce \
           --runonce-timeout=60s \
-          --api-servers=https://172.17.4.101:6443 \
-          --kubeconfig=/etc/kubernetes/kubeconfig.yaml \
+          --api-servers=https://172.17.4.101:443 \
```
dghubble: reason for the change? I can apply the same to our other examples.

hack/single-node/README.md
```diff
@@ -2,31 +2,19 @@
...
-
-This will render all tls assets, manifests, secrets, and user-data files necessary to stand up a local vagrant based development cluster. If you would like to change configuration for any reason, you can tweak parameters found in the `config-env` file.
-
-## Start VM
-
-```
-vagrant up
-```
-## Start Bootkube
+This will generate the default assets in the `cluster` directroy and launch a single-node self-hosted cluster.
```
dghubble: same, spelling

cmd/bootkube/start.go
aaronlevy: Yeah, I'll switch both to be `--etcd-servers`

cmd/bootkube/start.go
aaronlevy: Sure, can switch both to be `--asset-dir`

pkg/asset/templates/kubeconfig.yaml
```diff
@@ -6,10 +6,11 @@ clusters:
     server: {{ .Server }}
     certificate-authority-data: {{ .CACert }}
```
dghubble: To be clear, two separate certificates are generated and need to be placed on host via this kubeconfig (or in the automated situation by using this generated kubeconfig file directly).

pkg/asset/templates/kubeconfig.yaml
```diff
@@ -6,10 +6,11 @@ clusters:
     server: {{ .Server }}
```
dghubble: Drop .yaml from the template file name to correspond to dropping the extension everywhere else?

hack/multi-node/user-data.sample
aaronlevy: The 6443 was due to not being able to (easily) bind to a port <1024. But it's somewhat non-standard. Users will likely expect to be able to hi "https://kubernetes" for the api-server (and not rely on needing to determine port as well)

pkg/asset/tls.go
```diff
@@ -60,21 +68,39 @@ func newCACert() (*rsa.PrivateKey, *x509.Certificate, error) {
 	return key, cert, err
 }
 
-func newAPIKeyAndCert(caCert *x509.Certificate, caKey *rsa.PrivateKey, apiCertIPs []string) (*rsa.PrivateKey, *x509.Certificate, error) {
+func newAPIKeyAndCert(caCert *x509.Certificate, caKey *rsa.PrivateKey, altNames tlsutil.AltNames) (*rsa.PrivateKey, *x509.Certificate, error) {
 	key, err := tlsutil.NewPrivateKey()
 	if err != nil {
 		return nil, nil, err
 	}
+	altNames.IPs = append(altNames.IPs, net.ParseIP("10.3.0.1"))
```
dghubble: this eliminates the requirement that 10.3.0.1 be passed as the last argument during rendering, correct? e.g. `--apiserver-cert-ip-addrs=172.15.0.21,10.3.0.1`

pkg/asset/templates/kube-apiserver.yaml
```diff
@@ -23,26 +23,25 @@ spec:
         - /hyperkube
         - apiserver
         - --bind-address=0.0.0.0
+        - --secure-port=443
+        - --insecure-port=8080
         - --advertise-address=$(MY_POD_IP)
-        - --etcd-servers={{ .ETCDServers }}
+        - --etcd-servers={{ range $i, $e := .EtcdServers }}{{ if $i }},{{end}}{{ $e }}{{end}}
```
aaronlevy: Just noting that I want to move this logic out of the template soon. I didn't put in a #TODO because it ends up in the rendered templates too.

pkg/tlsutil/tlsutil.go
```diff
@@ -17,6 +17,17 @@ const (
 	Duration365d = time.Hour * 24 * 365
 )
 
+type CertConfig struct {
+	CommonName   string
+	Organization []string
+	AltNames     AltNames
+}
+
+type AltNames struct {
```
dghubble: There is an assumption that these are ordered pairs, I'd add a doc line.

dghubble:
If its too much work, definitely skip this, but it would be nicer to split the build changes (adding darwin, etc.) from the on-host changes.

dghubble:
In hack, I see this scp's bootkube to `/home/core/bootkube`. For automated tools, I think we'd want to provision control plane nodes with a release binary or container image

pkg/asset/templates/kubeconfig.yaml
aaronlevy: I kinda thought the same, but then I was leaving the .yaml on the template to make editors pick up filetype. I removed it from the rendered copy because it's gross to do `kubectl --kubeconfig=kubeconfig.yaml` everywhere...

pkg/asset/templates/kubeconfig.yaml
aaronlevy: I'm not clear on what you mean here.

pkg/asset/tls.go
aaronlevy: Correct. The service ip (10.3.0.1 in this example) will likely need to be plumbed through as a separate configurable at some point (if/when we allow configuring service-ip-cidr).

pkg/tlsutil/tlsutil.go
aaronlevy: This should just be unordered list of domain names and IPs. But I can add doc string if that's unclear.

aaronlevy: If you were thinking of the cli flags `IP=x.x.x.x,DNS=foo.com` those should be parsed prior to using this struct. These values are passed directly to the x509.Certificate object

aaronlevy:
The build changes end up tied with the on-host changes so they can be tested. Added darwin build because render/start could be run on different OSs (render on OSX, start on linux) -- and this was necessary to get hack / on-host tests working.

As far as the scp steps in hack, in any other tooling that is likely going to be "left up to deployer/docs". We will ship images so this could, for example, be run as `sudo rkt run quay.io/coreos/bootkube --exec start` -- but hack will likely continue to use a locally compiled version for dev/test.

pkg/tlsutil/tlsutil.go
dghubble: Perfect, seems fine then

pkg/asset/templates/kubeconfig.yaml
dghubble: Control plane nodes need to be provisioned with the generated kubeconfig file beforehand. In `hack` that's done by appending the entire file to the cloud-config. On physical machines, it was done by distributing the bearer token (secret material) and Certificate Authority certificate as metadata to template the kubeconfig file. With this change, we'll distribute the certificate authority certificate, client certificate, and client-key (secret material) via metadata (alternately, chuck the entire file in metadata, but that can be ugly).

There is no problem to address here, just noting the change and which bits are secret.

pkg/asset/templates/kubeconfig.yaml
dghubble: Ah, sounds reasonable

dghubble:
Ok for on-host changes with build changes. +1 for shipping an image for distribution purposes. lgtm.

derekparker:
lgtm

PR #37, opened 2016-05-17
pkg/asset: Use go-bindata for loading assets

dghubble:
Here is a quick proposal to use `go-bindata` for asset loading, rather than a roll our own approach. It has good support for generating templates in debug (references local files) or prod mode.

The only reason I had to generate `templates.go` into a binassets subpackage is because bootkube uses a type named `Asset` which we might consider renaming. Also, in the Makefile, the bin/bootkube step no longer needs to ensure templates are regenerated each time, for development use debug-templates to pick file changes. We still need to decide about committing generated files. Normally the policy would be to not generate them in applications, but bootkube is also being used as a library and needs to work with vendoring tools.

TODO:

- [ ] Vendor a go-bindata dependency for consistent asset generation
- [ ] Consider whether bootkube's notion of an Asset is should be change. `asset.Asset` is ugly, especially with a subpackage `binassets`. Asset is over-used.
- [ ] Improve the Makefile to represent the debug-template flow
- [ ] Resolve how to import and vendor `bootkube` as a dependency

cc @chancez @aaronlevy


aaronlevy:
If this will help you, I'm all for it. 

I have no strong affinity to the name Asset, so if that needs to change it's fine with me.

aaronlevy:
Do you want this reviewed as-is or is this a WIP?

dghubble:
Currently a WIP

Makefile
```diff
@@ -18,16 +18,21 @@ vet:
...
 	go build -o bin/bootkube github.com/coreos/bootkube/cmd/bootkube
 
 install: all
 	cp bin/bootkube $(GOPATH_BIN)
 
-pkg/asset/internal/templates.go: $(GOFILES)
-	mkdir -p $(dir $@)
-	go generate pkg/asset/templates_gen.go
+templates:
+	go-bindata -pkg binassets -o pkg/asset/binassets/templates.go -prefix pkg/asset/templates pkg/asset/templates/...
```
aaronlevy: I'm not super enthusiastic about requiring a third-party tool to build the project...

Makefile
```diff
@@ -18,16 +18,21 @@ vet:
 VENDOR_VERSION = v1.2.1
 vendor: vendor-$(VENDOR_VERSION)
 
-bin/bootkube: $(GOFILES) pkg/asset/internal/templates.go
+build: $(GOFILES)
```
aaronlevy: This target is a duplicate/unnecessary.

Makefile
```diff
@@ -18,16 +18,21 @@ vet:
...
 	mkdir -p bin
 	go build -o bin/bootkube github.com/coreos/bootkube/cmd/bootkube
 
 install: all
 	cp bin/bootkube $(GOPATH_BIN)
 
-pkg/asset/internal/templates.go: $(GOFILES)
-	mkdir -p $(dir $@)
-	go generate pkg/asset/templates_gen.go
+templates:
```
aaronlevy: Add to .PHONY

Makefile
aaronlevy: It doesn't look like they build release binaries. So would be be confident in `go get -u ...` always working?

dghubble: "Resolve how to import and vendor bootkube as a dependency"

It should be added to vendor. Also, this project doesn't seem to use godeps or glide and might benefit from using a tool.

aaronlevy:
Beyond the -debug flag, what does this provide us over what we currently have?

The debug option seems nice, but it really would only be used in the render command ... and that's not exactly something like wanting to iterate on a website (pretty static for the most part).

This doesn't really resolve the issue of either generating `templates.go` in libraries that want to use this, or just committing the `templates.go` in the first place. (as far as I can tell)

dghubble:
I was hoping `go-bindata` with debug might have an approach to avoid commiting both the sources and binary versions of that assets, but its not clear how that is done. Mainly using `go-bindata` does what `template_gen.go` did but is more mature and featured. 

I'd be happy if we just committed the generated file for now in order to unblock vendoring `bootkube` into other projects, especially as changes are happening.

dghubble:
Discussed offline

aaronlevy:
For posterity, the offline discussion:

Right now, as a quicker solution we should either commit `templates.go`, or add a stub file to `/pkg/asset/internal` so that glide doesn't complain about a missing package.

In the latter option, it would require users of this (as a library) to run `go generate vendor/github.com/coreos/bootkube/pkg/asset/template_gen.go`

Technically if this is being used as a library we should likely just commit the generated assets - I was just hoping to avoid this for now - given that it is changing very rapidly at the moment.

Longer term we can look at using `go-bindata` and `glide`, but not strictly necessary to resolve the issue at hand.

PR #38, opened 2016-05-18
pkg/asset/internal: Add generated templates

dghubble:
* Allows `github.com/coreos/bootkube/asset` to be imported directly

cc @aaronlevy


dghubble:
I'll rebase this and regenerate assets off master soon (~1 day), but currently depend on the named fork internally.

aaronlevy:
Before merging, any thoughts on how we can protect ourselves from builds with mismatched template.go? Off the top of my head:

- add templates.go as a PHONY make target so it is forced to run `go generate` every time we build bootkube target

- have the `clean` target remove template.go and add require a clean-before-build

aaronlevy:
lgtm

PR #41, opened 2016-05-23
pkg/asset/templates: use default service accounts

aaronlevy:
Fixes issue where controller-manager and scheduler were not actually contacting the api-server properly. They should be able to just use the service accounts added to their pods


derekparker:
lgtm

dghubble:
Thanks!

PR #42, opened 2016-05-24
bootkube: Allow render to take an existing CA

### LINKED ISSUE #26: Allow a CA to be provided to bootkube render
Bare metal examples could be simpler if a pre-existing CA cert could be provided to `bootkube render` so that only manifests and other credentials are generated.

Currently, I generate the CA with bootkube, paste it into different machine metadata declarations, then provision nodes to be bootkube-ready. It might also be valuable to pass a desired token to `bootkube`, depending on how long we're going to be using tokens.

derekparker:
Fixes #26


derekparker:
cc: @dghubble @aaronlevy

cmd/bootkube/render.go
```diff
@@ -24,16 +24,20 @@ var (
...
+		etcdServers    string
+		apiServers     string
+		altNames       string
 	}
 )
 
 func init() {
 	cmdRoot.AddCommand(cmdRender)
 	cmdRender.Flags().StringVar(&renderOpts.assetDir, "asset-dir", "", "Output path for rendered assets")
+	cmdRender.Flags().StringVar(&renderOpts.caFile, "ca-file", "", "Path to an existing PEM encoded CA.")
```
aaronlevy: s/ca-file/ca-certificate

Also in the help text, maybe add some behavior notes:

"If set, TLS assets will be generated using this certificate authority"

cmd/bootkube/render.go
```diff
@@ -24,16 +24,20 @@ var (
...
+	cmdRender.Flags().StringVar(&renderOpts.privateKeyFile, "private-key-file", "", "Path to the CA RSA private key.")
```
aaronlevy: s/private-key-file/ca-private-key/

Add behavior notes: "Required if --ca-certificate flag is set."

cmd/bootkube/render.go
```diff
@@ -24,16 +24,20 @@ var (
 	}
 
 	renderOpts struct {
-		assetDir    string
-		etcdServers string
-		apiServers  string
-		altNames    string
+		assetDir       string
+		caFile         string
```
aaronlevy: s/caFile/caCert (or caCertPath)

cmd/bootkube/render.go
```diff
@@ -24,16 +24,20 @@ var (
...
+		privateKeyFile string
```
aaronlevy: s/privateKeyFile/caPrivKey (or caPrivKeyPath)

cmd/bootkube/render.go
```diff
@@ -84,9 +96,11 @@ func flagsToAssetConfig() (c *asset.Config, err error) {
 		altNames = altNamesFromURLs(apiServers)
 	}
 	return &asset.Config{
-		EtcdServers: etcdServers,
-		APIServers:  apiServers,
-		AltNames:    altNames,
+		EtcdServers:    etcdServers,
+		CAFile:         renderOpts.caFile,
```
aaronlevy: s/CAFile/CACert

cmd/bootkube/render.go
```diff
@@ -84,9 +96,11 @@ func flagsToAssetConfig() (c *asset.Config, err error) {
...
+		PrivateKeyFile: renderOpts.privateKeyFile,
```
aaronlevy: s/PrivateKeyFile/CAPrivKey

pkg/asset/asset.go
```diff
@@ -36,9 +36,11 @@ const (
 // AssetConfig holds all configuration needed when generating
 // the default set of assets.
 type Config struct {
-	EtcdServers []*url.URL
-	APIServers  []*url.URL
-	AltNames    *tlsutil.AltNames
+	EtcdServers    []*url.URL
+	APIServers     []*url.URL
+	CAFile         string
+	PrivateKeyFile string
```
aaronlevy: I'm thinking we might want to require these to be x509 objects. This way we're not assuming how we retrieve this data (e.g. from disk), just that it is provided. In the case of other tools using this, we don't know that this data is going to be provided as a path.

dghubble: I think we're providing the contents (or we could provide as types), not the path.

derekparker:
Updated. cc: @aaronlevy @dghubble

derekparker:
@aaronlevy @dghubble data flags removed.

dghubble:
It is handy you've exposed wrappers around the x509 calls you expect to be made, thanks. We'll use `ParsePEMEncodedCACert(pemdata []byte) (*x509.Certificate, error)` and `ParsePEMEncodedPrivateKey(pemdata []byte) (*rsa.PrivateKey, error)` to parse data for `asset.Config` when using this package.

pkg/asset/tls.go
```diff
@@ -8,15 +8,20 @@ import (
...
+
+	if caCert == nil {
+		caPrivKey, caCert, err = newCACert()
+		if err != nil {
+			return assets, err
+		}
 	}
 
-	apiKey, apiCert, err := newAPIKeyAndCert(caCert, caKey, altNames)
+	apiKey, apiCert, err := newAPIKeyAndCert(caCert, caPrivKey, altNames)
```
dghubble: Use the name caPrivKey in `newAPIKeyAndCert` and `newKubeletKeyAndCert`?

pkg/asset/tls.go
```diff
@@ -31,13 +36,13 @@ func newTLSAssets(altNames tlsutil.AltNames) ([]Asset, error) {
 		return assets, err
 	}
 
-	kubeletKey, kubeletCert, err := newKubeletKeyAndCert(caCert, caKey)
+	kubeletKey, kubeletCert, err := newKubeletKeyAndCert(caCert, caPrivKey)
```
dghubble: The private key ends up being used to create new api and kubelet private keys and certificates signed by the CA ala. `tlsutil.NewSignedCertificate` and x509.CreateCertificate https://golang.org/pkg/crypto/x509/#CreateCertificate. This has the stipulation that the provided private key not have a passphrase right. I think that is fine, I just want to be sure I understand this.

dghubble:
Overall this looks great to me, thanks!

pkg/asset/tls.go
derekparker: That is correct, as per the current implementation. I think we could handle passphrase protected keys, but it's not being done currently.

pkg/asset/tls.go
derekparker: Done.

aaronlevy:
lgtm with squash

dghubble:
Same, lgtm

PR #43, opened 2016-05-24
pkg/asset: Update version metadata on manifests

derekparker:
It also seems some weren't prefixed with `v`, either way, it should be consistent.


aaronlevy:
lgtm

PR #45, opened 2016-05-27
Hack build conformance tests

aaronlevy:
Run conformance tests against the hack/single-node and hack/multi-node installs:

`make conformance-single`
`make conformance-multi`


derekparker:
Nice, lgtm.

PR #46, opened 2016-05-27
Simple checks

aaronlevy:
Adds travis tests for template generation, go vet/fmt, and go test (but there are currently no unit tests)

Depends on https://github.com/coreos/bootkube/pull/45

Only last 2 commits are relevant.


derekparker:
lgtm

PR #47, opened 2016-05-31
Add release build process

aaronlevy:
/cc @derekparker @pbx0 @dghubble

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,24 +2,33 @@ export GO15VENDOREXPERIMENT:=1
 export CGO_ENABLED:=0
 export GOARCH:=amd64
 
+LOCAL_OS:=$(shell uname | tr A-Z a-z)
 GOFILES:=$(shell find . -name '*.go' | grep -v -E '(./vendor|internal/templates.go)')
+TEMPLATES:=$(shell find pkg/asset/templates -type f)
 GOPATH_BIN:=$(shell echo ${GOPATH} | awk 'BEGIN { FS = ":" }; { print $1 }')/bin
 
-all: bin/linux/bootkube bin/darwin/bootkube
+all: _output/bin/linux/bootkube _output/bin/darwin/bootkube
+
+release: clean check _output/release/bootkube-linux-amd64.tar.gz _output/release/bootkube-darwin-amd64.tar.gz
 
 check: pkg/asset/internal/templates.go
 	@find . -name vendor -prune -o -name '*.go' -exec gofmt -s -d {} +
 	@go vet $(shell go list ./... | grep -v '/vendor/')
 	@go test -v $(shell go list ./... | grep -v '/vendor/')
 
-bin/%/bootkube: $(GOFILES) pkg/asset/internal/templates.go
+_output/bin/%/bootkube: LDFLAGS=-X github.com/coreos/bootkube/pkg/version.Version=$(shell $(CURDIR)/build/git-version.sh)
+_output/bin/%/bootkube: $(GOFILES) pkg/asset/internal/templates.go
 	mkdir -p $(dir $@)
-	GOOS=$* go build -o bin/$*/bootkube github.com/coreos/bootkube/cmd/bootkube
+	GOOS=$* go build -ldflags "$(LDFLAGS)" -o _output/bin/$*/bootkube github.com/coreos/bootkube/cmd/bootkube
 
-install: bin/$(shell uname | tr A-Z a-z)/bootkube
+_output/release/bootkube-%-amd64.tar.gz: _output/bin/%/bootkube
+	@mkdir -p $(dir $@)
+	tar czf $@ -C $(dir $<) bootkube
+
+install: _output/bin/$(LOCAL_OS)/bootkube
 	cp $< $(GOPATH_BIN)
 
-pkg/asset/internal/templates.go: $(GOFILES)
+pkg/asset/internal/templates.go: $(GOFILES) $(TEMPLATES)
 	mkdir -p $(dir $@)
 	go generate pkg/asset/templates_gen.go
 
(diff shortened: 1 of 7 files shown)


release/version.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
dghubble: This has been done already, why do this again a different way?
https://github.com/coreos/coreos-baremetal/blob/master/git-version
https://github.com/coreos/dex/blob/master/git-version

release/version.sh
```diff
@@ -0,0 +1,45 @@
...
+        GIT_TREE_STATE="dirty"
+        BOOTKUBE_VERSION+="-dirty"
+    fi
+}
+
+ldflag() {
+    echo "-X github.com/coreos/bootkube/pkg/version.${1}=${2}"
+}
+
+ldflags() {
```
dghubble: https://github.com/coreos/coreos-baremetal/blob/master/build#L3

pkg/version/version.go
```diff
@@ -0,0 +1,28 @@
+package version
```
dghubble: https://github.com/coreos/coreos-baremetal/blob/master/bootcfg/version/version.go#L4

pkg/version/version.go
```diff
@@ -0,0 +1,28 @@
...
+
+var (
+	buildDate    string = "1970-01-01T00:00:00Z"
```
dghubble: is this really needed?

dghubble:
This seems overly complicated for what it does.

release/version.sh
aaronlevy: Wasn't aware. I just took this more or less from the kubernetes version logic. But I'll switch to your example.

aaronlevy:
I pulled the version logic almost verbatim from kubernetes, but happy to switch it to the examples you provided.

Besides that is there specific complexity you want to see addressed?

dghubble:
Ah, from Kubernetes. I wondered where this came from. The general approach of setting a Version variable via ldflags seems good. Personally I'd favor just using the `git-version` script from other CoreOS projects, keeping the version, state, build time, and commit version doesn't seem needed.

I like the make target for tar'ing, I might borrow that. But I'd maybe keep using `bin`.

Travis can do the build for you to handle most of `build-release.sh`. Otherwise, you'd need more scripting to get `sudo rkt run` working on Travis, which I haven't tried.

aaronlevy:
My plan is to have travis run `make clean release`. But the actual image building process I'd like to containerize with rkt -- so just assuming we'll leave that to jenkins jobs (`./build-release.sh`)

aaronlevy:
I care less about travis testing if the docker build process is messed up (vs not being able to create release binaries).

aaronlevy:
Also, side-note about the build process using rkt is that I want the only requirement for a build is an empty CoreOS node. Makes portability of this much easier to manage.

build/build-release.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+set -euo pipefail
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+BOOTKUBE_ROOT=${BOOTKUBE_ROOT:-${SCRIPT_DIR}/..}
```
dghubble: Are there real cases where these need to be overridden (except DOCKER_PUSH)?

build/build-release.sh
```diff
@@ -0,0 +1,27 @@
...
+BOOTKUBE_RELEASE=${BOOTKUBE_BIN:-${BOOTKUBE_ROOT}/_output/release/bootkube-linux-amd64.tar.gz}
+DOCKER_REPO=${DOCKER_REPO:-quay.io/coreos/bootkube}
+DOCKER_TAG=${DOCKER_TAG:-$(${BOOTKUBE_ROOT}/build/git-version.sh | tr + _)}
```
dghubble: tr + _ still needed?

build/build-release.sh
aaronlevy: This doesn't really need to be configurable. I'll change it to be static.

Makefile
```diff
@@ -2,24 +2,33 @@ export GO15VENDOREXPERIMENT:=1
 export CGO_ENABLED:=0
 export GOARCH:=amd64
 
+LOCAL_OS:=$(shell uname | tr A-Z a-z)
```
dghubble: why is shell needed everywhere here?

build/build-release.sh
aaronlevy: Right now it's not strictly necessary, because we have no tags. But tags that contain semver comments (e.g. `+foo.1`) would break the docker tags. Mostly a habit from hyperkube releases because there are always comment tags.

I'll remove for now to keep complexity lower, and if we need it later it's easy to add back.

Makefile
```diff
@@ -2,24 +2,33 @@ export GO15VENDOREXPERIMENT:=1
...
 	@go test -v $(shell go list ./... | grep -v '/vendor/')
 
-bin/%/bootkube: $(GOFILES) pkg/asset/internal/templates.go
+_output/bin/%/bootkube: LDFLAGS=-X github.com/coreos/bootkube/pkg/version.Version=$(shell $(CURDIR)/build/git-version.sh)
+_output/bin/%/bootkube: $(GOFILES) pkg/asset/internal/templates.go
 	mkdir -p $(dir $@)
-	GOOS=$* go build -o bin/$*/bootkube github.com/coreos/bootkube/cmd/bootkube
+	GOOS=$* go build -ldflags "$(LDFLAGS)" -o _output/bin/$*/bootkube github.com/coreos/bootkube/cmd/bootkube
+
+_output/release/bootkube-%-amd64.tar.gz: _output/bin/%/bootkube
```
dghubble: I had to lookup the various symbols, but my makefile-fu is not strong. Seems ok.

dghubble:
Left a few notes, but looks much better to me. I'm interested to see how the rkt builds on Jenkins turn out. lgtm when you're happy with it.

Makefile
aaronlevy: Could probably use native makefile globbing/functions instead. But mostly just easier to use well-understood utilities (find / grep / etc).

PR #50, opened 2016-06-07
all: Add "user-space" apiserver checkpointing

derekparker:
This patch adds a new binary: "checkpoint". The checkpoint program will
ensure that the latest apiserver manifest is checkpointed, in case of a
system / apiserver crash.

It is implemented via static manifests. The checkpoint program will
store a pod manifest on disk. When it detects an apiserver is not
running, it will move that file into the directory that was specified as
the kubelet's config dir. From there, the kubelet will see that pod
manifest and run it. Once the program detects both our temporary
apiserver and self-hosted apiserver is running, it will remove the
manifest from the config dir, causing the kubelet to kill it and allow
the self-hosted apiserver to take over.


derekparker:
cc: @aaronlevy @dghubble 

Originally checkpoint was a shell script, but it got outta hand with some of the `jq` parsing I was doing so I converted it into a small go program so it was cleaner / easier to grok. Also makes a smaller image, which is nice since it can just be a scratch with a statically linked binary.

derekparker:
So, to elaborate on the process a bit more: 

Before bootkube dies, it now places a static manifest on disk which describes this checkpointer. The checkpointer is run by the kubelet. This has the benefit of surviving a total reboot (unless the checkpoint manifest is removed from disk). After this, the checkpointer takes over, which I've explained above but can elaborate on.

Makefile
```diff
@@ -7,7 +7,7 @@ GOFILES:=$(shell find . -name '*.go' | grep -v -E '(./vendor|internal/templates.
 TEMPLATES:=$(shell find pkg/asset/templates -type f)
 GOPATH_BIN:=$(shell echo ${GOPATH} | awk 'BEGIN { FS = ":" }; { print $1 }')/bin
 
-all: _output/bin/linux/bootkube _output/bin/darwin/bootkube
+all: _output/bin/linux/bootkube _output/bin/darwin/bootkube _output/linux/checkpoint _output/darwin/checkpoint
```
aaronlevy: _output/bin/linux... (missing bin in path)

hack/multi-node/Vagrantfile
```diff
@@ -127,6 +127,7 @@ Vagrant.configure("2") do |config|
 
       controller.vm.provision :file, source: CONTROLLER_USER_DATA_PATH, destination: "/tmp/vagrantfile-user-data"
       controller.vm.provision :shell, inline: "mv /tmp/vagrantfile-user-data /var/lib/coreos-vagrant/", privileged: true
+      controller.vm.provision :shell, inline: "mkdir /etc/kubernetes/manifests", privileged: true
```
aaronlevy: Can we move this into an ExecStartPre in the kubelet unit. It works here, but at least we're describing what is necessary in the systemd unit in a more portable way.

hack/single-node/Vagrantfile
```diff
@@ -49,5 +49,5 @@ Vagrant.configure("2") do |config|
 
   config.vm.provision :file, :source => USER_DATA_PATH, :destination => "/tmp/vagrantfile-user-data"
   config.vm.provision :shell, :inline => "mv /tmp/vagrantfile-user-data /var/lib/coreos-vagrant/", :privileged => true
-
+  config.vm.provision :shell, inline: "mkdir /etc/kubernetes/manifests", privileged: true
```
aaronlevy: same comment as above, move to user-data

pkg/asset/templates/kube-apiserver.yaml
```diff
@@ -30,33 +30,26 @@ spec:
...
           readOnly: true
       volumes:
       - name: ssl-certs-host
         hostPath:
           path: /usr/share/ca-certificates
-      - name: secrets
-        secret:
-          secretName: kube-apiserver
+      - name: tls
+        hostPath:
```
aaronlevy: This is a pretty big regression if we can no longer distribute tls assets via secrets. It means all master nodes must have tls data manually distributed to them.

Can we not also figure out a way to checkpoint the secrets for the checkpointed pods?

pkg/asset/templates/kubelet-master.yaml
```diff
@@ -0,0 +1,88 @@
...
+        command:
+        - /nsenter
+        - --target=1
+        - --mount
+        - --wd=.
+        - --
+        - ./hyperkube
+        - kubelet
+        - --api-servers={{ index .APIServers 0 }}
+        - --config=/etc/kubernetes/manifests
```
aaronlevy: Is the only difference with "kubelet-master" this flag? It should be fine to add this to all kubelets. We could either just allow static pods from the hosts at that location (but on nodes we don't really have any right now). Or we could rely on /etc/kubernetes/manifests not actually being mounted on the hosts for workers (then it's just a path inside the container and will only be available during the lifetime of the pod

pkg/asset/templates/kubelet-master.yaml
```diff
@@ -0,0 +1,88 @@
...
+        - --allow-privileged
+        - --hostname-override=$(MY_POD_IP)
+        - --cluster-dns=10.3.0.10
+        - --cluster-domain=cluster.local
+        - --kubeconfig=/etc/kubernetes/kubeconfig
+        - --lock-file=/var/run/lock/kubelet.lock
+        - --node-labels=master=true
```
aaronlevy: This also looks to be different, but is a no-op anyway as it only has an effect during node registration (which would be handled by the "on-host" kubelet)

pkg/asset/templates/kubelet-master.yaml
derekparker: No, the difference is that this kubelet has the `master=true` label, so that it will also get "master" components if it is still running when its apiserver goes down.

pkg/asset/templates/kube-apiserver.yaml
derekparker: Yeah, I'd rather continue with secrets as well. I can look into turning those into volume mounts.

pkg/asset/templates/kubelet-master.yaml
derekparker: Ah, so a kubelet coming up on the same host _without_ that flag will not negate it?

pkg/bootkube/bootkube.go
```diff
@@ -106,5 +111,46 @@ func (b *bootkube) Run() error {
...
+		return err
+	}
+
+	// Now that bootkube has finished without an error,
+	// move our "user space" checkpointing pod into place.
+	return ioutil.WriteFile(checkpointPodPath, checkpointPod, 0644)
+}
+
+var checkpointPod = []byte(`
+{
```
aaronlevy: I feel like there is a more portable way of distributing this.

What about the checkpoint pod knows how to install itself as a static pod. For example:

We deploy a checkpoint-install pod the same way we deploy any other object (daemonset template generated by render, injected via api). The checkpoint-install pod is really just using the same checkpoint image.. just referencing internal manifest files rather than the binary itself:

The checkpoint-installer container spec has a command: "/install.sh":

```
#!/bin/bash
cp /checkpoint.yaml /kubernetes/manifests
while true;
  sleep 3600
done
```

Then it has mounts for

host=/etc/kubernetes/manifests
dest=/kubernetes/manifests

This way it basically knows how to update itself.. just drops it's files in place and will survive reboots. If a new version comes along (for whatever reason) it will just replace the existing checkpointer static pod on disk.

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,157 @@
...
+	pods, err = ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		log.Println(err)
+	}
+	return pods
+}
+
+func bothAPIServersRunning(pods []byte) bool {
+	return bytes.Contains(pods, tempAPIServer) && bytes.Contains(pods, kubeAPIServer)
```
aaronlevy: This seems rather fragile. Ideally we would be inspecting the actual api objects (not just raw bytes). And even then we should be trying to match on various criteria (namespace=kube-system, expected labels (kubernetes.io/cluster-service), expected name, etc.)

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,157 @@
...
+`)
+
+var (
+	tempAPIServer = []byte("temp-apiserver")
+	kubeAPIServer = []byte("kube-apiserver")
+)
+
+const (
+	kubeletAPIPodsURL = "http://localhost:10255/pods"
+	ignorePath        = "/etc/kubernetes/_manifests/apiserver.json"
```
aaronlevy: The general convention here is to use /srv/kubernetes/manifests for the inactive assets

pkg/asset/templates/kubelet-master.yaml
aaronlevy: Correct. If node already exists .. doesn't matter what this flag says.

README.md
```diff
@@ -65,6 +65,14 @@ And optionally, to install into $GOPATH/bin:
 make install
 ```
 
+## Building checkpoint image
+
+From the project root, run:
+
+```
+docker build -t $(REPO):$(VERSION_TAG) -f cmd/checkpoint/Dockerfile .
```
aaronlevy: We possibly could just include this tool in the bootkube image also... it's already just a scratch image with static binary. And if so, could be added to build/build-release script.

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,157 @@
...
+			b, err := ioutil.ReadFile(ignorePath)
+			if err != nil {
+				log.Println(err)
+				continue
+			}
+			if err := ioutil.WriteFile(activePath, b, 0644); err != nil {
+				log.Println(err)
+			}
+		}
+		time.Sleep(time.Second)
```
aaronlevy: We should probably give more nuanced delay here. We don't need to checkpoint every second, and we probably want to give an api-server that's not running time to recover on its own (kubelet should try to restart).

pkg/bootkube/bootkube.go
derekparker: Seems reasonable, I like that.

cmd/checkpoint/main.go
derekparker: The reason I'm doing it rather frequently is to stay up to date. Consider if the apiserver pod is changed, we don't get notified, so instead we checkpoint frequently to ensure we're mostly up to date. I'm fine with backing this off a bit, or maybe diffing the manifest against the on-disk manifest, but that seems more complicated for little gain.

aaronlevy: Sure. Ideally we could do something like watch the kubelet-api (like the apis-server) and only act on pod changing -- but I'm not sure that the kubelet-api would provide something like that.

Alternatively, we could do something like hash the spec and if it's changed then re-checkpoint rather than blindly doing it every second. But I'm okay with this being a follow-up optimization.

But I do think we should give the api-server some time to recover on it's own. There already are some concepts of checkpointing via docker labels (not sure about rkt) so that if the container dies, kubelet is responsible for starting it again (e.g. api-server). We just want to fail back to our checkpoint if that doesn't occur.

philips:
Somewhere in here you need to be using a POSIX "atomic move" from a temp file in the ignore directory to the right foo.json filename. THis will protect against partial writes during a crash.

derekparker:
@aaronlevy @philips updated.

Makefile
```diff
@@ -7,7 +7,7 @@ GOFILES:=$(shell find . -name '*.go' | grep -v -E '(./vendor|internal/templates.
 TEMPLATES:=$(shell find pkg/asset/templates -type f)
 GOPATH_BIN:=$(shell echo ${GOPATH} | awk 'BEGIN { FS = ":" }; { print $1 }')/bin
 
-all: _output/bin/linux/bootkube _output/bin/darwin/bootkube
+all: _output/bin/linux/bootkube _output/bin/darwin/bootkube _output/bin/linux/checkpoint _output/bin/darwin/checkpoint
```
aaronlevy: Is there ever a case that we would need a darwin binary?

derekparker: Nah. Will remove.

cmd/checkpoint-installer/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine
```
aaronlevy: We shouldn't need a separate image just for the install script. Just add `install.sh` to the checkpoint docker image, and the runtime command usage can just be different (`command: /install.sh` vs `command: /checkpoint`)

derekparker:
So, the new order of events is:

- Bootkube installs cluster
- API server comes up with a checkpoint-install container, which moves the checkpoint static manifest into place and then sleeps
- checkpoint pod is eventually run by kubelet and starts checkpointing the apiserver + secrets
- system reboots
- on-host kubelet starts checkpoint static manifest
- checkpoint process moves temp-apiserver static manifest into place
- rest of components come up
- checkpoint process removes temp-apiserver static manifest
- kubelet kills temp-apiserver, allowing self-hosted apiserver to take over

cmd/checkpoint-installer/checkpoint.json
```diff
@@ -0,0 +1,39 @@
+{
+        "kind": "Pod",
```
aaronlevy: kind of a nit, but every other manifest we have in this repo is yaml....

cmd/checkpoint-installer/checkpoint.json
```diff
@@ -0,0 +1,39 @@
...
+                                                "mountPath": "/etc/kubernetes"
+                                        },
+                                        {
+                                                "name": "srv-kubernetes",
+                                                "mountPath": "/srv/kubernetes"
+                                        }
+                                ]
+                        }
+                ],
+                "hostNetwork": true,
```
aaronlevy: Is this still necessary?

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/v1"
+	clientset "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_2"
```
aaronlevy: use release_1_3

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+)
+
+func main() {
+	log.Println("begin apiserver checkpointing...")
+	run()
+}
+
+func run() {
+	var tempAPIServerManifest v1.Pod
+	if err := json.Unmarshal(manifestBytes, &tempAPIServerManifest); err != nil {
```
aaronlevy: Is this just be cause it's easier to unmarshall than construct the v1.Pod object itself?

cmd/checkpoint-installer/checkpoint.json
derekparker: Why do you hate me? I kid, I can change it if you'd like...

cmd/checkpoint-installer/checkpoint.json
derekparker: Yeah, we're still talking to the kubelet read-only API via localhost.

cmd/checkpoint/main.go
derekparker: Initially I had that because I wasn't using any upstream objects (as data structures) but now that those are more prevalent here I think it makes sense to use a v1.Pod object, for sure.

cmd/checkpoint/main.go
derekparker: Cool, I took that from elsewhere in `pkg/bootkube` so I'll update that as well.

cmd/checkpoint-installer/Dockerfile
derekparker: Makes sense.

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	cleanVolumes(&apiPod)
+	modifyInsecurePort(&apiPod)
+	return apiPod.Spec
+}
+
+func newAPIClient() clientset.Interface {
+	kubeConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{
+		ExplicitPath: kubeconfigPath},
+		&clientcmd.ConfigOverrides{}).ClientConfig()
```
aaronlevy: style nit:

```
kubeConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
    &clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeconfigPath},
    &clientcmd.ConfigOverrides{}).ClientConfig()
```

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	client, err := clientset.NewForConfig(&restclient.Config{
```
aaronlevy: isn't 'kubeConfig' already a restclient.Config? Can you not pass it directly.

Also, you could use `clientset.NewForConfigOrDie()` since we're treating this as fatal anyway.

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+			// clean it up a bit, and then save it to the ignore path for
+			// later use.
+			tempAPIServerManifest.Spec = parseAPIPodSpec(podList)
+			convertSecretsToVolumeMounts(client, &tempAPIServerManifest.Spec)
+			writeManifest(tempAPIServerManifest)
+			log.Printf("finished creating temp-apiserver manifest at %s\n", ignoreFile)
+
+		default:
+			log.Println("no apiserver running, installing temp apiserver static manifest")
+			b, err := ioutil.ReadFile(ignoreFile)
```
aaronlevy: nit: s/ingoreFile/checkpointManifest -- or something a little more descriptive. So maybe the pair is something like:
`checkpointManifest`
`activeManifest`

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+
+func run() {
+	var tempAPIServerManifest v1.Pod
+	if err := json.Unmarshal(manifestBytes, &tempAPIServerManifest); err != nil {
+		log.Fatal(err)
+	}
+	client := newAPIClient()
+	for {
+		var podList v1.PodList
+		if err := json.Unmarshal(getPodsFromKubeletAPI(), &podList); err != nil {
```
aaronlevy: Nit: have `getPodsFromKubeletAPI` return the PodList object rather than []byte

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+
+func getPodsFromKubeletAPI() []byte {
+	var pods []byte
+	res, err := http.Get(kubeletAPIPodsURL)
+	if err != nil {
+		log.Println(err)
+		return pods
+	}
+	pods, err = ioutil.ReadAll(res.Body)
+	res.Body.Close()
```
aaronlevy: defer() after the initial error check?

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+func bothAPIServersRunning(pods v1.PodList) bool {
+	var kubeAPISeen, tempAPISeen bool
+	for _, p := range pods.Items {
+		switch {
+		case isKubeAPI(p):
+			kubeAPISeen = true
+		case isTempAPI(p):
+			tempAPISeen = true
+		}
+		if kubeAPISeen && tempAPISeen {
```
aaronlevy: Could simplify this to just be `return (kubeAPISeen && tempAPISeen)` after the "for" block

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+func isTempAPI(pod v1.Pod) bool {
+	return strings.Contains(pod.Name, "temp-apiserver") && pod.Namespace == api.NamespaceSystem
+}
+
+// cleanVolumes will sanitize the list of volumes and volume mounts
+// to remove the default service account token.
+func cleanVolumes(p *v1.Pod) {
+	volumes := make([]v1.Volume, 0, len(p.Spec.Volumes))
+	for _, v := range p.Spec.Volumes {
+		if !strings.Contains(v.Name, "default") {
```
aaronlevy: Just checking for "contains default" seems a little risky. Is there any other metadata we can check? or Maybe an exact name match?

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+		for _, vm := range c.VolumeMounts {
+			if !strings.Contains(vm.Name, "default") {
+				volumeMounts = append(volumeMounts, vm)
+			}
+		}
+		c.VolumeMounts = volumeMounts
+	}
+}
+
+func modifyInsecurePort(p *v1.Pod) {
```
aaronlevy: Can you just add a brief comment note why this is necessary

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	writeAndAtomicCopy(m, ignoreFile)
+}
+
+func parseAPIPodSpec(podList v1.PodList) v1.PodSpec {
+	var apiPod v1.Pod
+	for _, p := range podList.Items {
+		if strings.Contains(p.Name, string(kubeAPIServer)) {
```
aaronlevy: Maybe just as a precaution do the same namespace check here. Or also just call `isKubeAPI()`

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
```
aaronlevy: Because you're importing kubernetes tools, which all use glog -- will these work well together? e.g. errors from k8s packages are surfaced properly?

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+		log.Fatal(err)
+	}
+	return client
+}
+
+func convertSecretsToVolumeMounts(client clientset.Interface, spec *v1.PodSpec) {
+	log.Println("converting secrets to volume mounts")
+	for i := range spec.Volumes {
+		v := &spec.Volumes[i]
+		if v.Name == "secrets" && v.Secret != nil {
```
aaronlevy: We happen to call this secrets right now, but there's no guarantee it stays that way & also, we should be checkpointing all secrets anyway.. (anything `v.Secret != nil` I would think)

cmd/checkpoint/main.go
```diff
@@ -0,0 +1,268 @@
...
+			copySecretsToDisk(client, v.Secret.SecretName)
+			v.Secret = nil
+			break
+		}
+	}
+}
+
+func copySecretsToDisk(client clientset.Interface, name string) {
+	log.Println("copying secrets to disk")
+	if err := os.MkdirAll(tlsPath, 0755); err != nil {
```
aaronlevy: Can we try and follow the (sort of) existing secrets convention. Maybe:

`/var/run/secrets/coreos.com/checkpoints/<pod-name>/<secret-name>`

Eventually we could change <pod-name> to maybe the hash of the pod so we kept secrets in sync with pod checkpoints and could be GC'd etc.

Which brings up another side-point: What about cleaning up old secrets that no longer exist for the current checkpoint? This could be a follow-up issue (or just put TODO in code).

pkg/asset/templates/kube-apiserver.yaml
```diff
@@ -17,8 +17,13 @@ spec:
...
       hostNetwork: true
       containers:
+      - name: checkpoint-installer
+        image: quay.io/derek_parker/checkpoint-installer:v1.0.0
+        volumeMounts:
+        - mountPath: /etc/kubernetes/manifests
+          name: etc-k8s-manifests
       - name: kube-apiserver
-        image: quay.io/coreos/hyperkube:v1.3.0-alpha.5_coreos.0
+        image: quay.io/peanutbutter/hyperkube:v1.3.0-alpha.5_derek.1
```
aaronlevy: I'm fine with updating these for consistency, but wouldn't we only need the patched code in the kubelets?

pkg/bootkube/bootkube.go
```diff
@@ -27,6 +27,10 @@ const (
 	// and doesn't end up in a race with bootkube for the insecure port. When bootkube dies, the self-hosted
 	// api-server is using the correct standard ports (443/8080).
 	insecureAPIAddr = "http://127.0.0.1:8081"
+	// The path that the "user space" checkpointing pod will be saved to.
+	// It is assumed that the kubelets running have been configured to
+	// read static manifests from this path.
+	checkpointPodPath = "/etc/kubernetes/manifests/checkpoint.json"
```
aaronlevy: Is this used anywhere?

aaronlevy:
Finished with this review pass

cmd/checkpoint/main.go
derekparker: Cannot do exact name match because it's something like `default-token-dsaf8338h` (e.g. default-token with random junk after) IIRC. I can double check that, but maybe it's better to check against `default-token`?

aaronlevy: Yeah, that's probably enough. We could do regex match, but I think `startsWith(default-token-)` should be sufficient.

cmd/checkpoint/main.go
aaronlevy: nm. I guess if there were a lot of pods you could early exit. But Could be slightly simpler I think

```
for pod := range pods {
    kubeAPISeen = kubeAPISeen || isKubeAPI(p)
    tempAPISeen = tempAPISeen || isTempAPI(p)
    if kubeAPISeen && tempAPISeen {
       return true
   } 
    return false
}
```

cmd/checkpoint/main.go
derekparker: I could go either way, but since we `ReadAll` it seems fine just to close it immediately.

pkg/asset/templates/kube-apiserver.yaml
derekparker: Yeah, that's correct.

derekparker:
@aaronlevy addressed most (if not all) of the review feedback, this PR should be mostly there now.

aaronlevy:
I reworked some of the build process - there were a few different unused files hanging around. Made the release a bit simpler, and moved all of the image stuff under `./image`. I haven't had a chance to test but can you take a look at: https://github.com/derekparker/bootkube/compare/wip-user-space-checkpointing...aaronlevy:checkpoint-updates?expand=1 and incorporate those changes.

So to get this merged: there are a few things I'm going to open an issue about following up on -- but we need to get this released, so we can resolve the outstanding issues later. 

First, We will need to break this into two PRs if possible:

1. (first pr) All changes necessary to incorporate the actual checkpointing mechanism in the bootkube image
1. Then we can tag a release which contains the checkpoint stuff in the image
1. (second PR) that updates bootkube to actually use the checkpoint image (in api-server and such).

derekparker:
@aaronlevy makes sense, I'll split it out this morning.

derekparker:
@aaronlevy should be good for another review. Going to start working on second PR to update templates and actually include the checkpointing just to verify everything still works as expected, and then will updated once this is actually merged and tagged.

hack/single-node/user-data.sample
```diff
@@ -26,14 +26,17 @@ coreos:
...
+        Environment=KUBELET_VERSION=v1.3.0-alpha.5_derek.1
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes/manifests
+        ExecStartPre=/bin/mkdir -p /srv/kubernetes/manifests
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --api-servers=https://172.17.4.100:443 \
           --kubeconfig=/etc/kubernetes/kubeconfig \
           --lock-file=/var/run/lock/kubelet.lock \
           --exit-on-lock-contention=true \
           --allow-privileged \
+          --config=/etc/kubernetes/manifests \
```
aaronlevy: Can you add this same change to `hack/quickstart/kubelet.master` and `hack/quickstart/kubelet.worker`

hack/single-node/user-data.sample
```diff
@@ -26,14 +26,17 @@ coreos:
       content: |
         [Service]
         EnvironmentFile=/etc/environment
-        Environment=KUBELET_ACI=quay.io/coreos/hyperkube
-        Environment=KUBELET_VERSION=v1.3.0-alpha.5_coreos.0
+        Environment=KUBELET_ACI=quay.io/peanutbutter/hyperkube
+        Environment=KUBELET_VERSION=v1.3.0-alpha.5_derek.1
```
aaronlevy: I think we can drop the custom image for now. We can ship a release which has the checkpoint code, but doesn't use it. Then we can update this in the follow up PR (also should have an official `beta.1` soon which contains the hostip fix).

image/checkpoint-pod.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: kube-api-checkpoint
+  namespace: kube-system
+spec:
+  containers:
+  - name: checkpoint
+    image: quay.io/derek_parker/bootkube:v0.1.0
```
aaronlevy: Hmm seems to be a bit of chicken / egg here. Makes me think we might need to build the checkpoint images separately :( -- updates will be difficult otherwise. Can't update this until an image is cut, but can't cut a new image unless tagged / release.

hack/multi-node/user-data.sample
```diff
@@ -25,13 +25,16 @@ coreos:
       content: |
         [Service]
         EnvironmentFile=/etc/environment
-        Environment=KUBELET_ACI=quay.io/coreos/hyperkube
-        Environment=KUBELET_VERSION=v1.3.0-alpha.5_coreos.0
+        Environment=KUBELET_ACI=quay.io/peanutbutter/hyperkube
+        Environment=KUBELET_VERSION=v1.3.0-alpha.5_derek.1
```
aaronlevy: Drop for now until we have official image / we need the fixes contained here explicitly

image/checkpoint-pod.yaml
derekparker: Separate image seems the way to go here.

hack/single-node/user-data.sample
derekparker: Done.

derekparker:
@aaronlevy PTAL.

build/build-release.sh
```diff
@@ -1,27 +1,11 @@
...
-printf "FROM scratch\nCOPY bootkube /bootkube" > ${TEMPDIR}/Dockerfile
-tar xzvf ${BOOTKUBE_RELEASE} -C ${TEMPDIR}
-docker build -t ${DOCKER_REPO}:${DOCKER_TAG} ${TEMPDIR}
-rm -rf ${TEMPDIR}
+    "cd /go/src/github.com/coreos/bootkube && make release"
 
-if [[ ${DOCKER_PUSH} == "true" ]]; then
-    docker push ${DOCKER_REPO}:${DOCKER_TAG}
-fi
+source build-image.sh
```
aaronlevy: Looks like this still exists, but should be changed to build-bootkube-image + build-checkpoint-image?

aaronlevy: Actually, maybe for now just change to `build-bootkube-image` (and drop the old script if no longer used).

Then what we can do is for the checkpoint pod, we build / reference by commit hash (build script will already do that).

Release process becomes:

build-checkpoint-image
ends up with `quay.io/coreos/checkpoint:<githash>`

Update the hash of the image in pod manifest
tag release
build-bootkube-release

Then they can be released in unison

build/build-checkpoint-image.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -euo pipefail
+
+BOOTKUBE_ROOT=${BOOTKUBE_ROOT:-$(git rev-parse --show-toplevel)}
+
+DOCKER_REPO=${DOCKER_REPO:-quay.io/coreos/checkpoint}
```
aaronlevy: Can we pick a less generic name -- given it's the coreos top-level account: maybe something like `pod-checkpointer`

image/checkpoint/checkpoint-pod.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: kube-api-checkpoint
+  namespace: kube-system
+spec:
+  containers:
+  - name: checkpoint
+    image: quay.io/coreos/checkpoint:v0.1.0
```
aaronlevy: Also change this if the repo name changes

pkg/asset/templates/kube-apiserver.yaml
aaronlevy: The official image should work now, can we update these to use: `quay.io/coreos/hyperkube:v1.3.0-beta.1_coreos.0` -- or because the functionality isn't strictly necessary yet (not actually running checkpointing) -- that could be part of the follow up PR (in which case just drop these image changes all together)

aaronlevy:
Just a few more minor comments

pkg/asset/templates/kube-apiserver.yaml
aaronlevy: This still needs to be changed..

pkg/asset/templates/kube-controller-manager.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: kube-controller-manager
-        image: quay.io/coreos/hyperkube:v1.3.0-alpha.5_coreos.0
+        image: quay.io/peanutbutter/hyperkube:v1.3.0-alpha.5_derek.1
```
aaronlevy: remove

pkg/asset/templates/kube-proxy.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       hostNetwork: true
       containers:
       - name: kube-proxy
-        image: quay.io/coreos/hyperkube:v1.3.0-alpha.5_coreos.0
+        image: quay.io/peanutbutter/hyperkube:v1.3.0-alpha.5_derek.1
```
aaronlevy: remove

pkg/asset/templates/kube-scheduler.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: kube-scheduler
-        image: quay.io/coreos/hyperkube:v1.3.0-alpha.5_coreos.0
+        image: quay.io/peanutbutter/hyperkube:v1.3.0-alpha.5_derek.1
```
aaronlevy: remove

pkg/asset/templates/kubelet.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: kubelet
-        image: quay.io/coreos/hyperkube:v1.3.0-alpha.5_coreos.0
+        image: quay.io/peanutbutter/hyperkube:v1.3.0-alpha.5_derek.1
```
aaronlevy: remove

aaronlevy:
After squash, lgtm

PR #54, opened 2016-06-12
Hide log-flush-frequency

### LINKED ISSUE #53: Remove log-flush-frequency flag
bootkube imports indirectly ```k8s.io/kubernetes/pkg/util.logs```, which is creating the global flag:

```
var logFlushFreq = pflag.Duration("log-flush-frequency", 5*time.Second, "Maximum number of seconds between log flushes")
```
As far as I can guess, this flag will not be used in bootkube, but it shows up through the Usage function.

```
$ ./bootkube version -h
Output version information

Usage:
  bootkube version [flags]

Global Flags:
      --log-flush-frequency=5s: Maximum number of seconds between log flushes
```

Can the flag be removed?

odacremolbap:
hides the flag log-flush-frequency from the command line.
this flag is set from k8s libraries and not used in bootkube.

Fixes #53

### DIFF
--- a/cmd/bootkube/main.go
+++ b/cmd/bootkube/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	"github.com/coreos/bootkube/pkg/version"
 )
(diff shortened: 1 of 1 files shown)


aaronlevy:
closed in https://github.com/kubernetes-incubator/bootkube/pull/143

PR #55, opened 2016-06-13
hack: add quickstart scripts & initial GCE guide

aaronlevy:
The quickstart directory contains generic scripts / kubelet.service files which should work for a CoreOS node launched on various providers.

These scripts currently assume `/etc/environment` has been populated with `$COREOS_PUBLIC_IPV4` and `$COREOS_PRIVATE_IPV4`.

The initial guide in the README describes how to use these scripts on vanilla CoreOS nodes using GCE.

ref: https://github.com/coreos/bootkube/issues/32

/cc @philips @pbx0 @dghubble

### DIFF
--- a/hack/quickstart/.gitignore
+++ b/hack/quickstart/.gitignore
@@ -1,0 +1,1 @@
+cluster/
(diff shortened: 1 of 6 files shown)


pbx0:
lgtm

PR #56, opened 2016-06-21
pkg/asset/templates: Name the kubeconfig context

dghubble:
Its useful to give the kubeconfig context a name when managing many k8s clusters with configs along a KUBECONFIG path. Until now, I've manually added a name to be able to switch back and forth (`kubectl config use-context NAME`), so it would be handy if that were done automatically.


aaronlevy:
Not necessarily opposed to this, but if you're specifying a specific kubeconfig (via env var or otherwise) doesn't it only know about the contexts in that file? So wouldn't adding a context to a single-context file not really change anything?

chancez:
@aaronlevy the env var `KUBECONFIG` actually supports multiple files, separated by `:` so you can have multiple files all with unique context names.

dghubble:
Yep, there is a config file per cluster.

aaronlevy:
lgtm

dghubble:
This is still desired. You want a current context to be added to address the concerns of https://github.com/coreos/bootkube/pull/63?

PR #58, opened 2016-06-22
all: Update path for checkpointed secrets

derekparker:


image/checkpoint/checkpoint-pod.yaml
```diff
@@ -15,14 +15,14 @@ spec:
       name: etc-kubernetes
     - mountPath: /srv/kubernetes
       name: srv-kubernetes
-    - mountPath: /var/run/secrets/coreos.com/checkpoints
+    - mountPath: /etc/kubernetes/checkpoint-secrets
```
aaronlevy: /etc is already mounted above, is this necessary?

aaronlevy:
One minor question, but otherwise lgtm

image/checkpoint/checkpoint-pod.yaml
derekparker: Nope, not necessary, will remove.

PR #59, opened 2016-06-22
all: Include checkpointer

derekparker:
This patch simply includes the checkpointer, and bumps the hyperkube
image version to the latest beta release, which carries a patch needed
for checkpointing to work.

Checkpointing is enabled automatically. The API server pod now contains
a "checkpoint-installer" container, which will install the checkpoint
static manifest, at which point the kubelet will run the checkpoint pod.


aaronlevy:
lgtm assuming the travis test passes

PR #61, opened 2016-06-22
pkg/asset/templates: update version metadata

derekparker:


derekparker:
Closing since https://github.com/coreos/bootkube/pull/62 will update versions.

PR #63, opened 2016-06-23
Revert "pkg/asset/templates: Name the kubeconfig context"

aaronlevy:
Reverts coreos/bootkube#56

If a current context is not set, it does not get used. This was surfaced by kubelet complaining about a CA cert mismatch.


dghubble:
LGTM

PR #65, opened 2016-06-23
cmd/checkpoint: Verify API server running

derekparker:
Adds a check to ensure the API is actually running, responding to
requests, when the checkpoint service notices the pod manifest. Came
accross an issue when upgrading a cluster where the API server daemonset
was removed, and the API server brought down, but the kubelet still
returned that pod from its read-only API. This meant that the checkpointer
never enabled the temp apiserver static manifest, meaning the
upgrade could not continue.

Now, if we notice that an API server pod is returned from kubelet (the
actual, self-hosted one, not a temp one) then we also check that it is
up and accepting requests, otherwise we assume it to be dead.

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -69,7 +69,7 @@ func run() {
 			if err := os.Remove(activeManifest); err != nil {
 				glog.Error(err)
 			}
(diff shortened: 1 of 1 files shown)


aaronlevy:
lgtm

PR #67, opened 2016-06-23
Cluster update script

derekparker:
Details are in the commit messages, but:

Includes a script to update a self-hosted cluster. Additionally removes version labels in manifests because it breaks when updating deployments.

Depends on https://github.com/coreos/bootkube/pull/65.


derekparker:
Script is pretty basic, but wanted to get something up for discussion asap.

aaronlevy:
This is a good start - but we don't want to be relying on assets being available locally. In the beautiful self-hosted future -- we should only be interacting with api-objects .. no need for anything local (unless you really really want to).

So we should be able to do something along the lines of:

For deployments use apply and sed magic to change image version:
`kubectl get deployment foo -o yaml | sed 's/\(image: myimage\):.*$/\1:vXYZ/' | kubectl apply -f -`

For daemonsets the initial (somewhat dumb) upgrade can be to apply the change to the existing daemonset (a-la sed process above) -- but then need to kick each pod in that daemonset so they get re-created by the daemonset controller.

There's a lot more nuance we want further down the line... but that should be a good starting place.

derekparker:
@aaronlevy yeah, I was thinking along those same lines of not having any files available locally, interacting completely with the API server. I can either make that change here, or save it for the next iteration of this idea.

aaronlevy:
It kinda defeats the purpose of self-hosted if we're relying on the state to remain valid external to the cluster -- so I don't think we should commit this as-is

derekparker:
@aaronlevy agreed, I'll update it.

aaronlevy:
I should actually clarify on some of this:

In a production cluster, it would behoove you to have your assets committed, and an upgrade would become: kubectl apply -f <foo> (pretty much what you have now). So having "outside" state doesn't negate the benefits of the self hosted (ignore what I said previously).

However, this tool is meant as a demonstration of upgrading the version of kubernetes using automated (self-hosted-enabled) methods -- so this even could mean "kubectl get" to store locally, then sed, then "kubectl apply" to apply the update.

But as a step 1.. let's have a script that can upgrade the version and exit... how people manage their backups / version control doesn't need to be tackled yet.

aaronlevy:
Also, what about the version labels breaks the upgrade?

derekparker:
> Also, what about the version labels breaks the upgrade?

This error:

```
Updating controller-manager
The Deployment "kube-controller-manager" is invalid.
spec.template.metadata.labels: Invalid value: {"k8s-app":"kube-controller-manager","version":"v1.3.0-beta.2"}: `selector` does not match template `labels`
```

aaronlevy:
The deployment labels and the spec labels need to match ... just need to make sure they are updated in sync (which should be sane enough with sed)

derekparker:
IIRC I updated all labels. I think the mismatch comes from trying to "apply" this to the deployment object that already exists. Could be wrong, though.

derekparker:
@aaronlevy last commit adds update without needing local manifests. I've ran through an update a few times locally and it seems to work fine. If you run though it, just bring up a single node as normal, and then run the update script.

This should outline the basic procedure. As I mentioned, it works fine for me locally. If you run it and get issues let me know, I had some weird stuff happen on Friday when I was working on this where it would hang, but not seeing that happen now.

The script basically just rolls everything, starting with apiserver.

README.md
```diff
@@ -52,6 +52,25 @@ Example:
 bootkube start --asset-dir=my-cluster
 ```
 
+## Hack
```
aaronlevy: I just moved this section in another PR, can we drop this change here.

hack/single-node/bootkube-up
```diff
@@ -10,10 +10,11 @@ fi
 # Render assets
 if [ ! -d "cluster" ]; then
   ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.100:443
+
+  # Add rendered kubeconfig to the node user-data
+  cat user-data.sample > cluster/user-data && sed 's/^/      /' cluster/auth/kubeconfig >> cluster/user-data
```
aaronlevy: nit - this could be separate PR (seems wanted, but unrelated)

hack/update-cluster
```diff
@@ -0,0 +1,81 @@
+#!/bin/sh
+set -e
```
aaronlevy: set -euo pipefail

hack/update-cluster
```diff
@@ -0,0 +1,81 @@
...
+}
+
+# Update kube-apiserver
+# Object is a DaemonSet, so we must first delete, then create to update.
+echo
+echo "Updating apiserver"
+$kc get ds kube-apiserver -oyaml > "${temp}/kube-apiserver.yaml"
+update_image "${temp}/kube-apiserver.yaml"
+# Delete kube-apiserver and wait for temp-apiserver to take over.
+$kc delete --timeout=1s -f "${temp}/kube-apiserver.yaml"
```
aaronlevy: Hmm. This is a bit risky to just blow away the whole thing. At minimum we should set `--cascade=false` so we don't just kill all pods. But even then, the daemonset controller could nuke all of the pods at the same time.

I'd be okay with punting on more coordination for a "true" updater process with the `--cascade=false` -- as long as we clearly call this a POC (as it's really heavy-handed right now).

hack/update-cluster
```diff
@@ -0,0 +1,81 @@
...
+while true; do
+        echo "checking for apiserver..."
```
aaronlevy: I'm unclear on what this check is. Why delete then immediately check if an apiserver is running?

pkg/asset/templates/kube-apiserver.yaml
```diff
@@ -5,13 +5,11 @@ metadata:
   namespace: kube-system
   labels:
     k8s-app: kube-apiserver
-    version: v1.3.0-beta.2
```
aaronlevy: I don't see how this is an issue especially given you're deleting the entire daemonset anyway -- there shoulldn't be anything to conflict with

pkg/asset/templates/kube-controller-manager.yaml
```diff
@@ -5,13 +5,11 @@ metadata:
   namespace: kube-system
   labels:
     k8s-app: kube-controller-manager
-    version: v1.3.0-beta.2
```
aaronlevy: Can we dig into why this is necessary? Not being able to support label metadata seems odd. It's also used in upstream addons - so how are those being upgraded? https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/cluster-monitoring/standalone/heapster-controller.yaml

hack/update-cluster
derekparker: I don't understand what you mean here. Even if all pods get removed, our checkpointed API server will kick in and we can continue from there.

hack/update-cluster
derekparker: Yeah, probably not accomplishing much here, we might wanna wait or something.

The original idea is basically that since we're explicitly bringing down the API server, let's hang out for a bit until the checkpointed API server kicks in and we can continue rolling the rest of the cluster (API server will sort itself out).

So instead of a timeout or something when creating the new daemonset we just wait until we have an API server to talk to and then continue.

README.md
```diff
@@ -52,6 +52,25 @@ Example:
...
+
+The coreos-baremetal repository contains a reference implementation using bootkube with bootcfg and a baremetal cluster.
+
+* [hack/single-node](hack/single-node/README.md)
+* [hack/multi-node](hack/multi-node/README.md)
+* [hack/quickstart](hack/quickstart/README.md)
+* [coreos-baremetal](https://github.com/coreos/coreos-baremetal/blob/master/Documentation/bootkube.md)
+
+## Upgrading
+
```
aaronlevy: Add a note about this being a proof-of-concept tool.

hack/update-cluster
aaronlevy: Add comment about this being a proof-of-concept tool

pkg/asset/templates/kube-apiserver.yaml
derekparker: What do you mean? I removed all version labels for consistency, since we need to remove them for deployments. Also, I don't know what it buys us, having the version in the manifest 3 times per manifest file. 2 labels + image tag seems overkill?

aaronlevy:
General notes: the daemonset delete/create will sort of work -- but isn't very graceful and doesn't provide mechanisms for any rolling-update or even rollback (e.g. what happens when you create a bad api-server manifest?)

I'm okay with this as a proof-of-concept kubectl based upgrade, but I think we need to say that clearly in the readme & in the script itself - as this has the potential to leave you in a bad state.

I will also try and put together a design doc for the next iteration of this tool, so it's clear how we envision this working.

pkg/asset/templates/kube-controller-manager.yaml
derekparker: We just can't change label metadata apparently. Update all version labels and then try a `kubectl apply -f` on that updated file and it will refuse to apply the update because of label mismatch.

derekparker:
@aaronlevy yeah, this is meant to just be a rough POC to outline the process of upgrading. I imagine we will scrap this for a full on update controller that can properly manage updates/rollbacks/etc. I agree with calling this out as POC so more folks are aware. I'll update the docs.

hack/update-cluster
aaronlevy: What if you deploy an api-server that "runs" but is non-functional? This will blow away all api-servers at the same time, deploy the non-functional api-server which will then be checkpointed and you're in an unrecoverable state.

Ideally we provide a mechanism for rolling updates for daemonsets (until this is implemented upstream). One option would be to deploy daemonsets side-by-side and using node-labels to coordinate which version is running where.

E.g. node-label: kubelet=1.1.1 vs node-label: kubelet=1.1.2 -- the upgrade tool slowly changes each node to the new version and lets the daemonset controllers coordinate killing / creating the correct pods.

Or another option would be to delete with `--cascade=false`, create the new daemonset, then have this tool kill each pod one-by-one. Before killing each pod it could check that the last update was successful -- if at any point it is not it can be stopped and replaced with the original daemonset.

Less risky because we're only updating 1 pod at a time -- but still risky in that if a pod is deleted for another reason it will be re-created with new manifest (not us controlling the update).

See some upstream discussion about this:

https://github.com/kubernetes/kubernetes/issues/15310
https://github.com/kubernetes/kubernetes/issues/22543

pkg/asset/templates/kube-apiserver.yaml
aaronlevy: I'd prefer to have label metadata, as it makes selecting on this data much easier than having to filter from the image itself.

It's also possible that the version of this "application" (as pod or manifest ) becomes separated from the image version. Or what about situations of multiple containers (and images) per pod. Then we have to filter each by name then by image (rather than by an object-level versioning).

pkg/asset/templates/kube-controller-manager.yaml
aaronlevy: This might just be down to being explicit with selectors https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/cluster-monitoring/standalone/heapster-controller.yaml#L27

hack/update-cluster
derekparker: I like the second approach. I'll try and apply that to this script.

pkg/asset/templates/kube-controller-manager.yaml
derekparker: Yeah, that seems to be it. Nice!

derekparker:
@aaronlevy updated to add a slightly more nuanced DaemonSet upgrade. Also was able to add back in version labels and apply the update for deployments. Minor misc cleanup as well.

Note: This upgrade process is broken for `v1.3.0-beta.2` due to volume mounting issues: https://github.com/kubernetes/kubernetes/issues/27653. To test this you can upgrade from an earlier k8s version to `v1.3.0-beta.1`, e.g. `v1.3.0-alpha.5` to `v1.3.0-beta.1`.

derekparker:
Closing in favor of full implementation.

PR #73, opened 2016-06-25
pkg/asset: mount /var/lib/rkt in kubelet container

aaronlevy:
Ran into an error after kubelet reboot where it would not start the sync loop because cadvisor couldn't inspect path at /var/lib/rkt

```
Failed to start cAdvisor inotify_add_watch /var/lib/rkt/pods/run/<podid>/stage1/rootfs/opt/stage2/hyperkube/rootfs/sys/fs/cgroup/cpu,cpuacct/system.slice/locksmithd.service: no such file or directory
```


derekparker:
lgtm

PR #77, opened 2016-07-01
Render add cluster dns and service cluster ip range

dlamotte:
First time contributing to CoreOS.  Let me know if something is a miss.


aaronlevy:
Hi @dlamotte, thanks for the contribution!

A couple general comments:

One of the things we've always tried (slash had trouble with) balancing is choosing what configuration items we should plumb through the tooling & what we should leave up to the end user to modify themselves.

For example, in our coreos-kubernetes reference installations these options more or less come down to: https://github.com/coreos/coreos-kubernetes/blob/a80ad85235aafe9a143231416bcd3d8af6d477b9/single-node/user-data#L4-L35

However, for each configurable we decide to support (even in the above cases)-- that can quickly spiral into a much more complex system -- and is a motivation for allowing for a simpler `render--> edit--> deploy` workflow where customization is left up to the end-user in the `edit` stage (for most options beyond the default templates).

Not saying that all options are off the table, but I think it's worthwhile having some clear reasoning for why they should be configurable in the tooling from the start.

So I guess from the perspective of this PR, I'm curious what in your case necessitates moving from the default service CIDR? Conflicts with existing internal IPs?

And then, if this is something we decide to move forward with, there would need to be a few more changes to this PR:

- we should probably just allow the service-ip-range to be the configurable flag - then choose an IP out of the range for clusterDNS (one less flag + guaranteed to be in the correct range).

- Should also check that the chosen service-ip-range does not conflict with the pod-cidr.

- The kubernetes service will be assigned to the first IP in that range, and need to sign certificates accordingly: https://github.com/coreos/bootkube/blob/33a3250edaed1df9d56266da6e08a4fd4885dd15/pkg/asset/tls.go#L81

aaronlevy:
One more note. If we do add more configuration items, I'd probably like to move to a versioned config format, so we're not relying on flags but can instead parse directly from a versioned file.

dlamotte:
So sorry it took so long to get back to this.  It's been a bit crazy around here.  Thanks so much for your detailed reply!

When I first submitted the PR, the premise of render/edit/deploy escaped me a bit.  You calling it out directly definitely helped me understand that and this became less of an issue.  Initially, I had been trying to setup Ansible to auto-provision our internal infrastructure.  I was planning to persist all details of a specific cluster in ansible variables and propagate those through the render operation of bootkube.

After a few weeks attempting to do everything via ansible, I would agree, the render/edit/deploy cycle is significantly easier and more flexible.

As for the reasons moving away from default CIDR, yes.  Its not clear how our company uses the internal 10.0.0.0/8 address space however I'm told the 192.168.0.0/16 address space is completely free.

I do appreciate your suggestions and agree with them whole-heartedly.

At this time now, we've since moved to deploying to AWS (still ansible based off of kube-aws).  And as such, I don't want this story to become anymore stale than it already has.  I'll close this and reopen later if I still think its a concern.

Thanks!

PR #78, opened 2016-07-05
cmd/checkpoint: prefer 127.0.0.1 over localhost

derekparker:


chancez:
@derekparker I'm fairly certain this will fix my issue, but I don't really understand what the checkpoint installer is doing so it's been troublesome to test with my own checkpointer. 

For future context: I found this while trying to run bootkube against my own "baremetal" setup (virtualized) where my DNS server isn't configured to respond to dns queries for `localhost`. Since the docker image you're building has no `/etc/hosts`, the DNS query gets forwarded, but nothing is authoritative for `localhost` so it ends up failing. 

We should never use `localhost` if possible and always prefer `127.0.0.1` where we might be using `localhost`.

aaronlevy:
lgtm

PR #81, opened 2016-07-09
Bump version v1.3.0_coreos.1

aaronlevy:
Still need to run through conformance as sanity check, but putting up for review.

/cc @pbx0 @derekparker


pbx0:
lgtm

PR #82, opened 2016-07-14
Node agent

derekparker:
This set of patches introduces the node agent. The job of this process is to monitor the API server and look for certain annotations on the node object. Those annotations describe the desired config and kubelet version. Whenever those change, the process will update the corresponding systemd unit files and restart the kubelet service.

This allows for updates of the host kubelet. As of right now, the annotation process must be done manually. In the future, the node will be annotated via the update controller. The node-agent also ensures that the kubelet version is less than or equal to the current API server version.


build/build-node-agent-image
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -euo pipefail
+
+BOOTKUBE_ROOT=${BOOTKUBE_ROOT:-$(git rev-parse --show-toplevel)}
+
+DOCKER_REPO=${DOCKER_REPO:-quay.io/derek_parker/node-agent}
```
aaronlevy: If all these build-image scripts are essentially the same (minus the tool name) we might want to make a generic script that takes the name as a parameter.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+	kubeletVersionService          = "kubelet-version"
+	kubeletConfigServicePath       = "/etc/systemd/system/kubelet.service.d/50-kubelet-config.conf"
+	kubeletVersionServicePath      = "/etc/systemd/system/kubelet.service.d/70-kubelet-version.conf"
+)
+
+var (
+	sysdConn *dbus.Conn
+	client   clientset.Interface
+
+	kubeAPIAddr = fmt.Sprintf("https://%s:%s", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS"))
```
aaronlevy: This doesn't seem to be used anywhere

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+	semver "k8s.io/kubernetes/pkg/version"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+const (
+	myPodNameEnvName               = "MY_POD_NAME"
+	desiredVersionAnnotation       = "updates.coreos.com/desired-version"
+	currentVersionAnnotation       = "updates.coreos.com/current-version"
+	desiredKubeletConfigAnnotation = "updates.coreos.com/desired-kubelet-config"
+	currentKubeletConfigAnnotation = "updates.coreos.com/current-kubelet-config"
```
aaronlevy: Let's alpha namespaces these: `updates.alpha.coreos.com`

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+	"k8s.io/kubernetes/pkg/runtime"
+	"k8s.io/kubernetes/pkg/util/wait"
+	semver "k8s.io/kubernetes/pkg/version"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+const (
+	myPodNameEnvName               = "MY_POD_NAME"
+	desiredVersionAnnotation       = "updates.coreos.com/desired-version"
+	currentVersionAnnotation       = "updates.coreos.com/current-version"
```
aaronlevy: I feel like we should be retrieving this from the source, rather than double storing in an annotation.

In the case of the node-object,  `node.Status.NodeInfo.KubeletVersion` would give us the current running version, but that could be off with self-hosted kubelet. For example, on-host is v1.1.1, but self-hosted is v1.1.2 -- then the node object would report 1.1.2 (and we would fail to update the on-host copy).

So in this case, I think we want to determine the on-host version (from the source). This makes me thing storing the actual version in a systemd drop-in might not be ideal (and instead just have the kubelet.service file source a file that is more easily parsed with probably just `KEY=value` pairs.

derekparker: That makes sense. I'll just use an EnvFile in the systemd unit.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+	sysdConn, err = dbus.New()
+	if err != nil {
+		glog.Fatal(err)
+	}
+
+	for {
+		// We say to never stop, but this method will
+		// return if it fails to contact an API server.
+		// This service should be resilient to temporary
+		// loss of an API server.
```
aaronlevy: Not necessarily opposed to this being in a loop, but the kubelet should restart this on failure (based on restart policy).

This would hide the failures (unless logs were checked), whereas letting the kubelet handle restarts would at least surface that in kubectl output.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+func newAPIClient() clientset.Interface {
+	config, err := restclient.InClusterConfig()
+	if err != nil {
+		glog.Fatal(err)
+	}
+	return clientset.NewForConfigOrDie(config)
+}
+
+func nodeUpdateCallback(oldObj, newObj interface{}) {
+	newNode := newObj.(*v1.Node)
```
aaronlevy: Can we check that the assertion was successful? I could see possibilities of getting bit by this (in the clientset 1_2 to 1_3 change I had `api.Pod` changed to `v1.Pod` -- and just happened to be surfaced in another error)

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+	ans := newNode.Annotations
+	glog.Infof("begin node update callback. Annotations: %#v", ans)
```
aaronlevy: nit: increase log level -- this will output all annotations every 30 seconds (on resync). Also, why print just values?

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+	if ans[desiredVersionAnnotation] != ans[currentVersionAnnotation] {
+		glog.Infof("node version update: %s", ans[desiredVersionAnnotation])
```
aaronlevy: nit: changing node version from %s to %s

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+		updateNodeVersion(ans[desiredVersionAnnotation], newNode)
+	}
+	if ans[desiredKubeletConfigAnnotation] != ans[currentKubeletConfigAnnotation] {
+		glog.Infof("node config update: %s", ans[desiredKubeletConfigAnnotation])
```
aaronlevy: nit: changing node config from %s to %s

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+	)
+	replaceSystemdDropInAndRestartUnit(kubeletConfigServicePath, kubeletConfigService, []*unit.UnitOption{uo})
+	node.Annotations[currentKubeletConfigAnnotation] = configMapName
+	updateNode(node)
+}
+
+func updateNode(node *v1.Node) {
+	_, err := client.Core().Nodes().Update(node)
+	if err != nil {
+		glog.Fatal(err)
```
aaronlevy: There are a lot of `glog.Fatal` -- do we really want this process crashing if something like this fails -- or just log error and will eventually try and resync.

pkg/asset/templates/kubelet.yaml
```diff
@@ -24,20 +24,60 @@ spec:
...
+        - --cluster-domain=$(CLUSTER_DOMAIN)
+        - --kubeconfig=$(KUBECONFIG)
+        - --lock-file=$(LOCK_FILE)
+        - --minimum-container-ttl-duration=$(MINIMUM_CONTAINER_TTL_DURATION)
         env:
           - name: MY_POD_IP
             valueFrom:
               fieldRef:
                 fieldPath: status.podIP
+          - name: API_SERVERS
```
aaronlevy: I would really like to not have to handle each flag as an individual entity.

We could just have the configmap drop a file which contains all flags (see example at: http://kubernetes.io/docs/user-guide/configmap)

Then we could just do something like: `bash -c "/kubelet $(cat $CONFIG)"`

We will eventually be replacing this with the upstream component config work ... so even if a file full of flags is a little gross, it is pretty clear how to use it.

pkg/asset/templates/node-agent.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: node-agent
+  namespace: kube-system
+  labels:
+    k8s-app: node-agent
```
aaronlevy: We should add version labels before merging like other components

cmd/node-agent/main.go
aaronlevy: Maybe `node-agent.alpha.coreos.com` ? updates is pretty generic -- and this is configuration for this specific tool.

aaronlevy:
General discussion points:

Updating config & version separately is a little worrisome. Flags could definitely change between versions, so we likely want to make this an atomic operation (as much is reasonable). Right now we even update the drop-ins and restart the service separately (where maybe one succeeds and the other fails).

I almost feel like the annotation should be a serialization of kubelet-version and configMap so that these are defined as a single unit.

Along those lines, I also think we probably need to version the config map so more than one can exist simultaneously. For example,

`node-agent.alpha.coreos.com={"kubeletVersion":"v1.1.1", "kubeletConfig":"v1"}`

derekparker:
> Updating config & version separately is a little worrisome. 

Yeah, I agree, great point.

> Along those lines, I also think we probably need to version the config map so more than one can exist simultaneously.

That's how it already works, in essence. The name of the configMap changes from like `kubelet-config-v1` to `kubelet-config-v2`, and the agent picks up on that. Both can exist together.

cmd/node-agent/main.go
derekparker: Yeah, I'll keep this print line but remove the annotation printing, that was just for my own verification. FWIW `%#v` doesn't print just values, it prints a Go-syntax representation of the variable.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+		"Environment",
+		fmt.Sprintf("KUBELET_VERSION=%s", version),
+	)
+	verifyAPIServerVersionGreaterOrEqual(version)
+	replaceSystemdDropInAndRestartUnit(kubeletVersionServicePath, kubeletVersionService, []*unit.UnitOption{uo})
+	node.Annotations[currentVersionAnnotation] = version
+	updateNode(node)
+}
+
+func verifyAPIServerVersionGreaterOrEqual(version string) {
```
derekparker: Yeah I agree this should be enforced at a higher level (the update-controller).

cmd/node-agent/main.go
derekparker: I initially di that to make it obvious when something is wrong (you can see clearly with `get pods` or something that it's failing). I can drop this down a bit though.

pkg/asset/templates/kubelet.yaml
derekparker: What do you think about just a single entry in the config map that is just like `kubelet.config: [all flags]` instead of dropping a file on disk? That way they are not handled individually, and we can still add/remove flags without having to change the kubelet manifest.

cmd/node-agent/main.go
derekparker: Circling back to this, I'm not sure I fully understand. You mention getting the kubelet version from the source (e.g. the node object we have in memory), but as you mentioned that could be the value of the self-hosted kubelet not the on-host kubelet. I'm generally in favor of deriving information rather than denormalizing it, but in this case, to me at least, it expresses explicitly just the version of the on-host kubelet, and gives insight into the cluster operator. In other words, you know when the update is completed by verifying that `current-version == desired-version`.

aaronlevy: You make a good point about (easily) surfacing the on host information to the admin. Given that, I think the current-version annotation makes sense.

Along those lines, however, I would want the annotation set from the functional source on each sync loop. For example, if it is defined in an env file on disk, then on each sync loop we look at the value and make sure the annotation matches.

Right now it looks like the "current version" annotation is only changed based on the "desired version" during an update. This would potentially make it so the source of truth is no longer matching annotation. For example, if config on disk is manually modified - the currentVersion would never be updated.

derekparker: Got it, thanks.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+		glog.Fatal(err)
+	}
+	return clientset.NewForConfigOrDie(config)
+}
+
+func nodeUpdateCallback(oldObj, newObj interface{}) {
+	newNode := newObj.(*v1.Node)
+	ans := newNode.Annotations
+	glog.Infof("begin node update callback. Annotations: %#v", ans)
+	if ans[desiredVersionAnnotation] != ans[currentVersionAnnotation] {
```
aaronlevy: I think we should also allow there to be no desired version: `if desired && desired != current`

cmd/node-agent/main.go
```diff
@@ -0,0 +1,220 @@
...
+		glog.Infof("node version update: %s", ans[desiredVersionAnnotation])
+		updateNodeVersion(ans[desiredVersionAnnotation], newNode)
+	}
+	if ans[desiredKubeletConfigAnnotation] != ans[currentKubeletConfigAnnotation] {
```
aaronlevy: I also think we probably want to support no desired config

pkg/asset/templates/kubelet.yaml
aaronlevy: Sure that's fine. Later we may want to still use volumes for the config as it might give us more flexibility (e.g. you can modify a volume during runtime, but you can't modify env vars or the pod spec). Not a huge deal, and this will rely on the upstream component config soon - so we'll be changing it again either way.

cmd/node-agent/main.go
derekparker: Very true, but I figure that temporary loss of an API server is not an explicit "failure" condition that we need to surface. If it's more than temporary, the cluster admin is likely being paged anyways.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,290 @@
...
+				return client.Core().Nodes().Watch(nodeopts)
+			},
+		},
+		&v1.Node{},
+		30*time.Second,
+		framework.ResourceEventHandlerFuncs{
+			UpdateFunc: nodeUpdateCallback,
+		},
+	)
+	var configmapopts api.ListOptions
```
aaronlevy: nit: s/configmapopts/configMapOpts/

cmd/node-agent/main.go
```diff
@@ -0,0 +1,290 @@
...
+	var configMapController *framework.Controller
+	configMapStore, configMapController = framework.NewInformer(
+		&cache.ListWatch{
+			ListFunc: func(lo api.ListOptions) (runtime.Object, error) {
+				return client.Core().ConfigMaps(api.NamespaceSystem).List(configmapopts)
```
aaronlevy: Actually, you don't need the configmapopts at all if we're not changing them. Can just use `lo` as passed in.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,290 @@
...
+	if err != nil {
+		glog.Error(err)
+		return
+	}
+	if configHasChanged(onDiskConfig, desiredConf) {
+		if err := validateConfig(desiredConf); err != nil {
+			glog.Error(err)
+			return
+		}
+		if err := updateKubeletEnvFile(desiredConf, onDiskConfig); err != nil {
```
aaronlevy: Why pass onDiskConfig here? All the info we need is in desiredConf, right?

cmd/node-agent/main.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func restartKubeletService() error {
+	// systemd daemon-reload equivalent
+	err := sysdConn.Reload()
+	if err != nil {
+		return err
+	}
+	ch := make(chan string)
+	_, err = sysdConn.RestartUnit("kubelet.service", "replace", ch)
```
aaronlevy: nit: make service name a const

cmd/node-agent/main.go
```diff
@@ -0,0 +1,290 @@
...
+	if err != nil {
+		return err
+	}
+	// Wait for unit to restart.
+	result := <-ch
+	if result != "done" {
+		return err
```
aaronlevy: This will always be nil. Should we return the result string maybe?

derekparker: Good catch.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	return nil
+}
+
+func configMapToCommandLineFlags(configMapName string) (string, error) {
+	var cmkey string
+	for _, key := range configMapStore.ListKeys() {
```
aaronlevy: I think we can drop this loop and just use`Get` or `GetByKey()` directly:

`cmi, ok, err := configMapStore.GetByKey("%s/%s", api.NamespaceSystem, configMapName)`

or use `Get()` (something like):

```
cm := api.ConfigMap{Namespace: api.NamespaceSystem, Name: configMapName}
cmi, ok, err :=configMapStore.Get(cm)
```

cmd/node-agent/main.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	// Wait for unit to restart.
+	result := <-ch
+	if result != "done" {
+		return err
+	}
+	return nil
+}
+
+func configMapToCommandLineFlags(configMapName string) (string, error) {
```
aaronlevy: We currently call this twice, which round-trips to apiserver. Seems like we should be able to get the config once, then just call this conversion function on the already-retrieved data.

hack/multi-node/user-data.sample
```diff
@@ -25,30 +25,24 @@ coreos:
...
-          --lock-file=/var/run/lock/kubelet.lock \
-          --exit-on-lock-contention \
-          --config=/etc/kubernetes/manifests \
-          --allow-privileged \
-          --hostname-override=${COREOS_PUBLIC_IPV4} \
-          --node-labels=master=true \
-          --minimum-container-ttl-duration=3m0s \
-          --cluster_dns=10.3.0.10 \
-          --cluster_domain=cluster.local
+        ExecStart=/usr/lib/coreos/kubelet-wrapper $PERSISTENT_KUBELET_FLAGS $KUBELET_FLAGS
```
aaronlevy: nit: for consistency with other service files we ship: s/KUBELET_FLAGS/KUBELET_OPTS

hack/multi-node/user-data.sample
```diff
@@ -25,30 +25,24 @@ coreos:
...
         Restart=always
         RestartSec=5
 
         [Install]
         WantedBy=multi-user.target
 
 write_files:
+  - path: "/etc/kubernetes/kubelet.env"
+    content: |
+      KUBELET_FLAGS='--api-servers=https://172.17.4.101:443 --kubeconfig=/etc/kubernetes/kubeconfig --lock-file=/var/run/lock/kubelet.lock --allow-privileged --config=/etc/kubernetes/manifests --hostname-override=$public_ipv4 --minimum-container-ttl-duration=3m0s --cluster_dns=10.3.0.10 --cluster_domain=cluster.local'
```
aaronlevy: nit: is it possible to multi-line this?

pkg/asset/templates/node-agent.yaml
```diff
@@ -0,0 +1,40 @@
...
+spec:
+  template:
+    metadata:
+      labels:
+        k8s-app: node-agent
+    spec:
+      containers:
+      - name: node-agent
+        image: quay.io/derek_parker/node-agent:f2e3b7224e7a048971a33f1df36c2d86812dcc73
```
aaronlevy: Before merging we might want to break this into 2 steps. Merge the tool & build the image, then merge this manifest with an official image.

aaronlevy:
Done with this review pass.

What is the status of passing the config items as a single item (rather than requiring each added in a separate env var). I think requiring the separate values is a blocker to merging this -- it makes admin fragile / difficult (e.g. change a configmap, and its behavior could be different for on-host and self-hosted due to kubelet manifest not picking up particular fields.

If this truly isn't possible, we might have to consider removing the config option altogether. But I'd like us to be confident this really is not possible before going this route.

cmd/node-agent/main.go
aaronlevy: Sorry, this is just retrieving from local cache. But it still seems like an overloaded function, where we should be able to have something like `getConfigMap` and `configMapToCommandLine` as isolated functions.

cmd/node-agent/main.go
derekparker: We need to "merge" instead of overwrite the env, since there are env vars on disk that are not included in the update (e.g. `PERSISTENT_KUBELET_CONFIG`).

dghubble:
How significant are the changes the node agent can make to the on-host kubelet? Is the scope agreed upon or defined anywhere? For example, setting one or two variables in a dropin vs. updating the unit.

My concern is that if Node Agent rewrites unit customizations, it needs to take on full responsibility to the kubelet in all scenarios, its no longer in my control as the provisioner. In particular, we're adding rkt options or potentially other customizations which should be preserved https://github.com/coreos/coreos-baremetal/blob/master/examples/ignition/bootkube-master.yaml#L47

derekparker:
@dghubble it does not touch the kubelet unit file. All the node agent modifies on disk is a single, specific ENV file that the systemd unit refers to. And within that, it only modifies 2 variables, everything else it leaves alone.

I could be more specific on that in the README since you had questions about it.

derekparker:
@dghubble and by "does not touch" I move does not modify. It does restart the kubelet service, but not modify it directly.

dghubble:
perfect, that sounds like a good strategy imo

derekparker:
@aaronlevy alright, all feedback has been addressed.

I've tried to get the single line config working, but get stuck in yaml / bash land.

This is an example kubelet manifest w/ bash:

```yaml
      - name: kubelet
        image: quay.io/coreos/hyperkube:v1.3.0_coreos.1
        command:
        - /nsenter
        - --target=1
        - --mount
        - --wd=.
        - --
        - /bin/bash
        - -c
        - ./hyperkube
        - $FLAGS
```

If I put the `/bin/bash` _before_ the `nsenter`, I get:

```
stdin: is not a tty
```

If I put it after, I get:

```
shell-init: error retrieving current directory: getcwd: cannot access parent directories: Inappropriate ioctl for device
Error: No server specified
...
```

Any thoughts?

colhom:
@derekparker one unrelated thing to keep in mind when in bash land- all of `./hyperkube <args>` is a single string parameter to `/bin/bash -c`:

```sh
$ /bin/bash -c echo hello world

$ /bin/bash -c "echo hello world"
hello world
```

So you should have:

```yaml
- /bin/bash
- -c
- "./hyperkube $FLAGS"
```

As far as your nsenter business goes... no dice on that one.

cmd/node-agent/README.md
```diff
@@ -0,0 +1,59 @@
...
+  --overwrite \
+  node <your node> \
+  node-agent.alpha.coreos.com/desired-config='{"KUBELET_VERSION": "v1.3.2_coreos.0", "KUBELET_CONFIG": "kubelet-config-v2"}'
+```
+
+You can then inspect the status of the on-host update by checking the following
+annotation value:
+
+```
+"node-agent.alpha.coreos.com/current-config"
```
aaronlevy: nit: would be helpful to show how to get this info. Something like:
 
`kubectl get node <node-name> -o=custom-columns=ANNOTATIONS:.metadata.annotations`

cmd/node-agent/main.go
```diff
@@ -0,0 +1,278 @@
...
+		glog.Error(err)
+	}
+}
+
+func configHasChanged(onDiskConfig, desiredConfig map[string]string, cm *v1.ConfigMap) bool {
+	if onDiskConfig[kubeletVersionKey] != desiredConfig[kubeletVersionKey] {
+		return true
+	}
+	cl := cm.Data[configMapFlagsKey]
+	if onDiskConfig[kubeletFlagsKey] != cl {
```
aaronlevy: nit: cl doesn't seem to be used:

`if onDiskConfig[kubeletFlagsKey] != cm[configMapFlagsKey]` ?

cmd/node-agent/main.go
```diff
@@ -0,0 +1,278 @@
...
+			return
+		}
+	}
+	// Always update annotation at end of sync loop.
+	if err := updateNodeAnnotation(newNode, desiredConf); err != nil {
+		glog.Error(err)
+	}
+}
+
+func configHasChanged(onDiskConfig, desiredConfig map[string]string, cm *v1.ConfigMap) bool {
```
aaronlevy: is desiredConfig actually needed here? It would seem that we have the desired config in the form of the configMap

aaronlevy: Actually, nevermind. The configMap only contains the kubelet flags, not desired version.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,278 @@
...
+const (
+	myPodNameEnvName        = "MY_POD_NAME"
+	desiredConfigAnnotation = "node-agent.alpha.coreos.com/desired-config"
+	currentConfigAnnotation = "node-agent.alpha.coreos.com/current-config"
+	kubeletVersionKey       = "KUBELET_VERSION"
+	kubeletConfigKey        = "KUBELET_CONFIG"
+	kubeletFlagsKey         = "KUBELET_OPTS"
+	kubeletEnvPath          = "/etc/kubernetes/kubelet.env"
+	kubeletService          = "kubelet.service"
+	configMapFlagsKey       = "flags"
```
aaronlevy: nit: Be more specific (maybe "kubelet-flags") incase we need to overload this configMap later.

cmd/node-agent/README.md
derekparker: done

derekparker:
@aaronlevy updated.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,277 @@
...
+		glog.Error(err)
+		return
+	}
+	cm, err := getConfigMap(desiredConf[kubeletConfigKey])
+	if err != nil {
+		glog.Error(err)
+		return
+	}
+	if configHasChanged(onDiskConfig, desiredConf, cm) {
+		if err := validateConfig(desiredConf); err != nil {
```
aaronlevy: Probably should check this even prior to `configHasChanged`

cmd/node-agent/main.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	if _, ok := desiredConfig[kubeletConfigKey]; !ok {
+		return fmt.Errorf("configuration annotation does not contain required key: %s", kubeletConfigKey)
+	}
+	return nil
+}
+
+func updateKubeletEnvFile(conf, onDiskConfig map[string]string, cm *v1.ConfigMap) error {
+	var buf bytes.Buffer
+	onDiskConfig[kubeletVersionKey] = conf[kubeletVersionKey]
```
aaronlevy: nit: updating the onDiskConfig map as a side effect of this function feels a bit odd. Or if we do, it should be updated on success or return a new copy.

If the goal is to have the env file match our desired config exactly, we can probably just drop this merge and not pass in onDiskConfig at all (just construct the data you want on disk from conf/cm)

aaronlevy: Hmm. Looks like the PERSISTENT_KUBELET_OPTS would need to be merged from on-disk.

hack/multi-node/user-data.sample
```diff
@@ -25,30 +25,24 @@ coreos:
...
         RestartSec=5
 
         [Install]
         WantedBy=multi-user.target
 
 write_files:
+  - path: "/etc/kubernetes/kubelet.env"
+    content: |
+      KUBELET_OPTS='--api-servers=https://172.17.4.101:443 --kubeconfig=/etc/kubernetes/kubeconfig --lock-file=/var/run/lock/kubelet.lock --allow-privileged --config=/etc/kubernetes/manifests --minimum-container-ttl-duration=3m0s --cluster_dns=10.3.0.10 --cluster_domain=cluster.local'
+      PERSISTENT_KUBELET_OPTS='--exit-on-lock-contention=true --node-labels=master=true --hostname-override=$public_ipv4'
```
aaronlevy: Do we really need this? Could this not be solved by some nodes using a different configMap?

pkg/asset/templates/kubelet-config.yaml
```diff
@@ -0,0 +1,7 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: kubelet-config
+  namespace: kube-system
+data:
+  kubelet-flags: "--api-servers='{{ index .APIServers 0 }}' --config=/etc/kubernetes/manifests --allow-privileged --cluster-dns=10.3.0.10 --cluster-domain=cluster.local --kubeconfig=/etc/kubernetes/kubeconfig --lock-file=/var/run/lock/kubelet.lock --minimum-container-ttl-duration=3m0s"
```
aaronlevy: nit: if it's not too hard, multi line this (untested):

```
kubelet-flags: |
  --api-servers=asdfasdf \
  --config=/foo/bar \
```

cmd/node-agent/main.go
derekparker: Yeah, I agree. Probably makes sense to create a temporary "merge" map for this, instead of modifying `onDiskConfig`.

hack/multi-node/user-data.sample
derekparker: Area you referring to the `PERSISTENT_KUBELET_OPTS`? It's needed because the on-host kubelet by nature has certain distinct config items from the self-hosted kubelet, so these flags must persist to the on-host kubelet regardless of changes to the configMap.

derekparker: Ah, I guess maybe you were suggesting we have a configMap the self-hosted uses and a configMap the on-host uses? That's a possibility too, but then there's 2 things to update instead of 1.

pkg/asset/templates/kubelet.yaml
```diff
@@ -17,27 +17,19 @@ spec:
...
+        - "/nsenter --target=1 --mount --wd=. -- ./hyperkube kubelet --hostname-override=$(MY_POD_IP) $(FLAGS)"
         env:
           - name: MY_POD_IP
             valueFrom:
               fieldRef:
                 fieldPath: status.podIP
+          - name: FLAGS
+            valueFrom:
+              configMapKeyRef:
+                name: kubelet-config
```
aaronlevy: This might present a bit of a problem. This won't allow us to have different configuration for different nodes. For example, "register-unschedulable" or "--node-label=master=true".

The only option there is using the PERSISTENT_KUBELET_OPTS -- but then we're back to defining configuration on disk (which is not updatable via api)

derekparker: We can still have different configuration for different nodes, we would just need separate configMaps per node, which would be annoying. Nothing is stopping us from having a configMap with base configuration across nodes, and then another configMap per node w/ node specifc flags?

I guess it's assumed that `PERSISTENT_KUBELET_OPTS` is set during provisioning and not changed (e.g. consistent base flags for a functioning on-host kubelet).

aaronlevy: Well.. we already didn't have that option - so maybe this isn't a big deal.

So the upstream route which we can eventually use is likely going to allow "global config" vs "node-named config" .. which we can piggy back on.

In the interim, what do you think about using the on-disk checkpointed config? That way, whatever config is defined by the node-annotation is what we use (even in the self-hosted copy).

hack/multi-node/user-data.sample
aaronlevy: If we do keep this, let's keep a consistent prefix: `KUBELET_PERSISTENT_OPTS`

aaronlevy:
A few nits and a few questions, but this is looking close

hack/multi-node/user-data.sample
aaronlevy: Ah, right. The `--exit-on-lock-contention` is a good reason for this. Maybe named `KUBELET_SYSTEM_OPTS` or `KUBELET_HOST_OPTS` (alluding to the "system-hosted kubelet" we've been using to describe it.

pkg/asset/templates/kubelet.yaml
derekparker: So instead of the kubelet DS depending on the configMap directly it will essentially `cat /etc/kubernetes/kubelet.env` for it's flags? That seems reasonable.

What do you think, then, about having the node-agent kick the kubelet pod for it's node? That way, the node agent updates config, and then rolls both the on-host and self-hosted kubelets to get new config.

hack/multi-node/user-data.sample
derekparker: Yeah, that rename makes sense.

pkg/asset/templates/kubelet.yaml
aaronlevy: We wouldn't easily be able to define a different config-map per node because the daemonset describes all nodes (unless we started doing per-node manifests... which we don't want).

May need to think on this a bit more

derekparker: We could merge those flags if we go with the option of using the on-disk config for both on-host and self-hosted. For example, we could make it so that the node-agent will look for the config named in the annotation and then `$configName-$nodeName` as well, and merge those flags. That way we keep a single daemonset, but get the flexibility of node specific config. This is mimicing what the proposed upstream kubelet config would be doing too, correct?

derekparker:
Notes from offline discussion w/ @aaronlevy 

--

- On-disk config is always source of truth for both on-host and self-hosted kubelet
- Nodes are provisioned w/ initial config
- Document the annotation / configmap based config update strategy
- Ensure that self-hosted kubelet is sourcing from on-disk config

How to prevent storing inital config multiple places?
- store config on disk (systemd drop-in), both kubelets read from disk
- initial config stored in annotation
- new configuration is given via config map (can copy flags from node annotation to get started)
- new config replaces old, annotation is updated

How to prevent user-data overwriting config?
- use systemd drop-ins and prioritize node-agent config over initial config

How do we show what the current kubelet is configured with?
- Expose current flags via annotation (base64 encoded)

derekparker:
@aaronlevy updated.

Per our conversation I removed the initial config map and added documentation of how to create a config map to change config. Also I made it so that the current flags string is added to the annotations.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	if newNode.Annotations == nil {
+		newNode.Annotations = make(map[string]string)
+	}
+	rawconf, ok := newNode.Annotations[desiredConfigAnnotation]
+	if !ok {
+		if _, ok := newNode.Annotations[currentConfigAnnotation]; !ok {
+			setCurrentConfigAnnotation(newNode, onDiskConfig)
+		}
+		glog.Info("no configuration annotation found, ignoring")
```
aaronlevy: nit: `("no %s annotation found for node %s, ignoring", desiredConfigAnnotation, newNode.Name)`

cmd/node-agent/main.go
```diff
@@ -0,0 +1,302 @@
...
+	if err := updateCurrentConfigNodeAnnotation(newNode, updatedConfig); err != nil {
+		glog.Error(err)
+	}
+}
+
+func setCurrentConfigAnnotation(node *v1.Node, onDiskConfig map[string]string) {
+	currentConfig := make(map[string]string)
+	currentConfig[kubeletVersionKey] = onDiskConfig[kubeletVersionKey]
+	currentConfig[kubeletFlagsKey] = onDiskConfig[kubeletFlagsKey]
+	currentConfig[kubeletConfigKey] = "intiial on-disk configuration"
```
aaronlevy: nit: "on-disk configuration" -- not necessarily initial, as it could change over time. Or if left as is: s/intiial/initial/

cmd/node-agent/main.go
```diff
@@ -0,0 +1,302 @@
...
+	if !ok {
+		if _, ok := newNode.Annotations[currentConfigAnnotation]; !ok {
+			setCurrentConfigAnnotation(newNode, onDiskConfig)
+		}
+		glog.Info("no configuration annotation found, ignoring")
+		return
+	}
+	var desiredConfig map[string]string
+	if err = json.Unmarshal([]byte(rawconf), &desiredConfig); err != nil {
+		glog.Error(err)
```
aaronlevy: nit: add some context here. E.g. `("error unmarshaling config from %s: %v", desiredConfigAnnotaiton, err)`

cmd/node-agent/main.go
aaronlevy: nit: add a godoc description of what this function checks. Mentally parsing something like `onDiskConfig[kubeletFlagsKey] != cm.Data[configMapFlagsKey]` isn't immediately clear.

cmd/node-agent/main.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+		env[kv[0]] = kv[1]
+	}
+	return env, nil
+}
+
+func updateCurrentConfigNodeAnnotation(node *v1.Node, conf map[string]string) error {
+	newConf, err := json.Marshal(conf)
+	if err != nil {
+		return err
```
aaronlevy: nit: either provide more context here, or possibly in the caller.

hack/single-node/user-data.sample
```diff
@@ -20,37 +20,49 @@ coreos:
...
+        WantedBy=multi-user.target
     - name: kubelet.service
       enable: true
       command: start
       content: |
+        [Unit]
+        After=kubelet-initial-config.service
+
         [Service]
         EnvironmentFile=/etc/environment
```
aaronlevy: Can drop this line

hack/single-node/user-data.sample
```diff
@@ -20,37 +20,49 @@ coreos:
...
+    permissions: "0744"
+    owner: core
+    content: |
+      #!/bin/bash
+
+      if [[ ! -f /etc/kubernetes/kubelet.env ]]; then
+          /bin/echo \
+          "KUBELET_OPTS='--api-servers=https://172.17.4.100:443 --kubeconfig=/etc/kubernetes/kubeconfig --lock-file=/var/run/lock/kubelet.lock --allow-privileged --config=/etc/kubernetes/manifests --minimum-container-ttl-duration=3m0s --cluster_dns=10.3.0.10 --cluster_domain=cluster.local'
+      PERSISTENT_KUBELET_OPTS='--exit-on-lock-contention=true --node-labels=master=true --hostname-override=$public_ipv4'
+      KUBELET_VERSION=v1.3.0_coreos.1" > /etc/kubernetes/kubelet.env
```
aaronlevy: nit: indent the two above lines

aaronlevy:
Some minor comments / nits, but looking really close.

Overall comments:

- Won't block on this, but we should start adding tests especially for these longer-lived tools. But we can do this in follow up PRs if you like.

- Assuming next step is to build an image of the node-agent and add the manifest as follow up?

So I'll leave the nits up to you, but this should just need a rebase - then lgtm when you're ready.

derekparker:
@aaronlevy great, thanks. I agree wrt adding tests, I'll try and get them into this PR.

derekparker:
@aaronlevy updated this PR, it's squashed and rebased. I've decided to submit a follow up PR for the tests, as it'll be easier to review that way I think.

After this is merged I'll also follow up w/ the node-agent manifest.

build/build-image.sh
```diff
@@ -0,0 +1,27 @@
...
+BOOTKUBE_ROOT=${BOOTKUBE_ROOT:-$(git rev-parse --show-toplevel)}
+
+DOCKER_REPO=${DOCKER_REPO:-quay.io/coreos/$CMD}
+DOCKER_TAG=${DOCKER_TAG:-$(${BOOTKUBE_ROOT}/build/git-version.sh)}
+DOCKER_PUSH=${DOCKER_PUSH:-false}
+
+TEMP_DIR=$(mktemp -d -t bootkube.$CMD.XXXX)
+
+cp $BOOTKUBE_ROOT/image/$CMD/* ${TEMP_DIR}
+cp $BOOTKUBE_ROOT/_output/bin/linux/$CMD ${TEMP_DIR}/$CMD
```
aaronlevy: nit: use `${foo}` syntax for all variables

cmd/node-agent/README.md
```diff
@@ -0,0 +1,87 @@
+# Node Agent
+
+## Description
+
+The Node Agent is a process that runs on each node of a Kubernetes cluster. The
+purpose of this process is to monitor a specific annotation on the node object.
+This annotation holds a JSON object which describes the version and configuration
+to be used for the on-host kubelet.
+
+This is useful for upgrading your on-host kubelet without ever having to ssh
```
aaronlevy: nit: In the design doc we've changed "on-host kubelet" to "system-hosted kubelet". Maybe change here for consistency.

cmd/node-agent/README.md
```diff
@@ -0,0 +1,87 @@
...
+into a machine. Simply update the annotation with the desired version, and the
+name of the configMap that holds your configuration, and the Node Agent will
+ensure that the on-host kubelet matches those values, and will restart the on-host
+kubelet so those new changes take effect.
+
+## Usage
+
+If you are using a supported self-hosted cluster then the Node Agent will be
```
aaronlevy: nit: "if you are using bootkube to launch your cluster"

cmd/node-agent/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+2. Annotate the node you wish to update:
+
+```
+$ kubectl \
+  --kubeconfig=cluster/auth/kubeconfig \
+  annotate \
+  --overwrite \
+  node <your node> \
+  node-agent.alpha.coreos.com/desired-config='{"KUBELET_VERSION": "v1.3.2_coreos.0", "KUBELET_CONFIG": "kubelet-config-v2"}'
```
aaronlevy: nit: make the kubelet-config name match the example configMap from above

cmd/node-agent/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+$ kubectl get nodes -o=custom-columns=NAME:.metadata.name,ANNOTATIONS:.metadata.annotations
+```
+
+## More Detail
+
+The Node Agent is responsible for updating a config file on disk that the kubelet
+service file points to. Specifically, it will only modify the `$KUBELET_OPTS`
+and `KUBELET_VERSION` env vars. Any other env vars in that file will be left
```
aaronlevy: nit: s/env vars/environment variables/

cmd/node-agent/main.go
```diff
@@ -0,0 +1,107 @@
...
+	clientset "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_3"
+	"k8s.io/kubernetes/pkg/client/restclient"
+	"k8s.io/kubernetes/pkg/controller/framework"
+	"k8s.io/kubernetes/pkg/fields"
+	"k8s.io/kubernetes/pkg/runtime"
+	"k8s.io/kubernetes/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+const myPodNameEnvName = "MY_POD_NAME"
```
aaronlevy: nit: s/MY_POD_NAME/POD_NAME

pkg/node/agent.go
```diff
@@ -0,0 +1,248 @@
...
+		goto update
+	}
+
+	// Obtain and validate the desired configuration.
+	desiredConfig, err = a.getDesiredConfig(node)
+	if err != nil {
+		glog.Error(err)
+	}
+
+update:
```
aaronlevy: I've never actually seen a goto in the wild before. Not opposed I guess, but why not just function?

aaronlevy: Also it looks like we would still continue past the desired config even if we can't parse it anyway, so this essentially is just skipping one possible log entry.

pkg/node/agent.go
```diff
@@ -0,0 +1,248 @@
...
+	updatedNode, err := a.handleConfigUpdate(*node, onDiskConfig, desiredConfig, defaultKubeletEnvPath)
```
aaronlevy: Desired config could be a nil map here. Is that okay?

pkg/node/agent.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func (a *Agent) handleConfigUpdate(node v1.Node, onDiskConfig, desiredConfig map[string]string, kubeletEnvPath string) (*v1.Node, error) {
+	// `node` is already a copy, but just to make things clear, assign to new var
+	newNode := node
+
+	// Get the ConfigMap specified in the annotation from the local cache.
+	cm, err := a.getConfigMap(desiredConfig[kubeletConfigKey])
+	if err != nil {
+		glog.Error(err)
```
aaronlevy: okay to continue?

pkg/node/agent.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	err = validateConfig(desiredConfig)
+	if err != nil {
+		glog.Error(err)
```
aaronlevy: okay to continue?

pkg/node/agent.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	// Check desired config / ConfigMap against the on disk config.
+	if err == nil && configHasChanged(onDiskConfig, desiredConfig, cm) {
```
aaronlevy: Ignore the above comments, see that you're checking err here

aaronlevy:
So mostly just nits, and one question - but overall looks good!

derekparker:
This work has merged w/ full cluster-updater PR. Closing.

PR #87, opened 2016-07-25
hacks/: add corectl guide

AntonioMeireles:
Hi all, 

attached is, for review, an initial draft of a guide to use bootkube together with [corectl](github.com/TheNewNormal/corectl), on macOS

All the Best, 

António

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ vendor-v[0-9].[0-9].[0-9]/
 
 hack/*/.vagrant/
 hack/*/cluster/
+hack/*/*.qcow2
+.vscode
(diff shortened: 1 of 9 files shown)


aaronlevy:
Hi Antonio,

Thanks for this contribution - and super exciting people are starting to use bootkube for their own installations.

As far as including another reference implementation, bootkube is still moving really quickly, and the flags / api may change in breakable ways - and adding a 3rd party implementation means we have another piece to make sure works inline with any changes we make here.

Ideally external implementations would be built using versioned releases (binary or rkt/docker image). This is how the `hack/quickstart` guides work too. The `hack/single-node` and `hack/multi-node` were meant to be used during direct development of bootkube itself.

Maybe as an alternative we can add a readme section which can link out to projects built using bootkube? Would that seem reasonable?

Issue #89, opened 2016-07-26
Allow cluster members identified by DNS name


dghubble:
Background:

We're migrating provisioning of cluster members to use DNS names instead of IPs to ensure reference clusters are similar to production https://github.com/coreos/coreos-baremetal/issues/291. This allows network settings to be provided by network DHCP server since the gatway, DNS server, IP, etc. can change over the life of the machine. I'll note we're using static MAC address to IP mappings which are fine for now, this is really to move away from hardcoding networkd configurations and instead respect the network setup.

Problem:

There are a few places in `bootkube` which still assume IPs will be used (`--hostname-override=$(MY_POD_IP)` via downward api) and I think we'll need the ability to toggle that assumption. In the absence of an explicit `--hostname-override`, the kubelet [exec](https://github.com/kubernetes/kubernetes/blob/master/pkg/util/node/node.go#L33)'s `uname -n` to determine the hostname to use. During provisioning we can enforce that `/etc/hostname` is populated with the correct name to satisfy the kubelet, but we need `kubelet.yaml` to leave off the flag. Also, `kube-apiserver.yaml` also uses the downward api to set `- --advertise-address=$(MY_POD_IP)` which can be left off.

Desired Changes:

I'll share the tweaks I made which allowed bootkube to bootstrap a DNS-only cluster. The current behavior is that nodes are registered with their DNS name via on-host and then via their IP via the pod kubelet and `bootkube start` always times out.

dghubble:
See #90

aaronlevy:
Should be closed by: https://github.com/kubernetes-incubator/bootkube/pull/151

The advertise address change was reverted in: https://github.com/kubernetes-incubator/bootkube/pull/257 -- but lets open a separate issue for that if this was a concern.

PR #90, opened 2016-07-26
DO NOT MERGE: Allow cluster members identified by DNS

dghubble:
These are the raw changes I needed for make for #89. These changes probably need to be gated behind a flag and are not intended to merge.


pkg/asset/templates/kube-apiserver.yaml
```diff
@@ -32,7 +32,6 @@ spec:
         - --bind-address=0.0.0.0
         - --secure-port=443
         - --insecure-port=8080
-        - --advertise-address=$(MY_POD_IP)
```
aaronlevy: IIRC this was because with vagrant/virtualbox the default interface is not routable between vms (I would need to re-verify this). Given that the vagrant support is only for local clusters, i'd be okay resolving this in some other way to make our manifests more generic.

pkg/asset/templates/kubelet.yaml
```diff
@@ -27,17 +27,11 @@ spec:
         - --api-servers={{ index .APIServers 0 }}
         - --config=/etc/kubernetes/manifests
         - --allow-privileged
-        - --hostname-override=$(MY_POD_IP)
```
aaronlevy: We should probably use the node name here. Then however the on-host kubelet has registered with the apiserver (IP or hostname), the self-hosted kubelet will always use the same.

There isn't currently support for the node name in the downward api., but there is an open PR: https://github.com/kubernetes/kubernetes/pull/27880

dghubble: Could we carry that in the CoreOS Hyperkube and update bootkube to make use of it?

luxas: You can't drop it either for the time being... see https://github.com/kubernetes/kubernetes/pull/29907 that fixes the problem

In newer releases, you may omit `HostnameOverride` I think. kubelet will use `os.Hostname()` as name and the IP of the default gateway as the external ip

dghubble: No problem with dropping the hostname override since hosts to run the kubelet-wrapper with an `/etc/resolv.conf` mount, so its not always using 8.8.8.8.

dghubble:
Using a forked copy until Node name is available in the downward api and bootkube can update to using that without breaking Vagrant examples.

Issue #88, opened 2016-07-25
bootkube flag to tell which k8s version to install 


rimusz:
Would be very handy to have the bootkube flag to tell which k8s version to install, as right now
it looks like it is hardcoded.

As we have tested bootkube with [corectl](https://github.com/TheNewNormal/corectl) and there is a PR #87 how to use it.
I'm looking to swap go binaries to bootkube in [kube-solo](https://github.com/TheNewNormal/kube-solo-osx) and [kube-cluster](https://github.com/TheNewNormal/kube-cluster-osx) to take an advantage of an easy k8s setup, but having no way to tell which k8s version to use, makes that not very useful.

aaronlevy:
One concern is that the vendored kubernetes code would potentially be a different version than the cluster you're launching.

This would mean the temporary control-plane (bootkube) would then replace itself with a self-hosted control-plane of a different version.

This may not be much of an issue (particularly if the version skew is small), but it is something to should consider.

AntonioMeireles:
@aaronlevy, 

Hi!, for context- our tooling ([TheNewNormal/...](https://github.com/TheNewNormal) just wants to give users the flexibility to test/simulate in diferent contexts, where those could be just cocktails with diferent CoreOS versions / channels or, at an higher level, different k8s versions - which is where this matters.)

So, (thinking loud) would be any, by design, side effect if one just started building versioned bootkube blobs (suffixed with VENDOR_VERSION perhaps) each set built against a given k8s release (being that the subset of those that matters would be the ones that have a correpondence [here](https://quay.io/repository/coreos/hyperkube?tab=tags)) ? other than a makefile tweak and stopping hardcoding k8s versions on the templates nothing else would be needed right ?

thanks in advance!

aaronlevy:
You could build versioned bootkube blobs, but I think this might be a little out of scope for the current goals (essentially, launch a *new* cluster and get out of the way as quickly as possible).

I think as a start we should be very clear in our releases which version of kubernetes will be launched, so a user can go back through previous releases if they really need to launch an earlier version.

But as far as building multiple versions, it's not always as simple as changing a vendor version. For example, updating the api client code from the 1.2 to 1.3 required changes to bootkube itself: 

https://github.com/coreos/bootkube/commit/66e399c525dfa976d96ce3444077b6da3bce874e
https://github.com/coreos/bootkube/commit/2e0df9db493abdb5a282ec857aef35a61b1cc286

[...]

PR #92, opened 2016-07-28
pkg/asset/templates/kube-proxy.yaml: fix typo in image's URL

AntonioMeireles:
fwiw spotted it while playing with #88 interimn "solution"


aaronlevy:
lgtm

PR #94, opened 2016-08-03
hack/single-node: Update bootkube-up to add kubeconfig once

chancez:
This is to match multi-node. This was causing TLS errors because it
caused the user-data containing the kubeconfig to be updated on every
bootkube-up run, when it should only happen on the first run.

### DIFF
--- a/hack/single-node/bootkube-up
+++ b/hack/single-node/bootkube-up
@@ -10,10 +10,9 @@ fi
 # Render assets
 if [ ! -d "cluster" ]; then
   ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.100:443
+  # Add rendered kubeconfig to the node user-data
+  cat user-data.sample > cluster/user-data && sed 's/^/      /' cluster/auth/kubeconfig >> cluster/user-data
 fi
-
-# Add rendered kubeconfig to the node user-data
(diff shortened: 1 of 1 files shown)


chancez:
cc @aaronlevy @derekparker

hack/single-node/bootkube-up
```diff
@@ -9,12 +9,11 @@ fi
 
 # Render assets
 if [ ! -d "cluster" ]; then
-  ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.100:443
+  ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster $BOOTKUBE_RENDER_FLAGS
```
derekparker: Is this variable for being able to override settings? These scripts are set up such that you should not have to provide any config, they should "just work" with the vagrant setups. If you want/need the ability to configure we should at least provide a default.

derekparker: "this variable" being `$BOOTKUBE_RENDER_FLAGS`.

chancez: Ah crap, I accidently included that. Thats my customization because bootkube's default stuff isn't very flexible. Let me remove that.

chancez: @derekparker fixed

derekparker:
LGTM

PR #95, opened 2016-08-04
Bump version to v1.3.4_coreos.0

dghubble:
Probably needs to be bumped in the vagrant examples too


aaronlevy:
Yeah, would you mind updating this in the hack/ examples? I guess I'm gonna have to switch my self-hosted demos to down-grading the control-plane :)

dghubble:
Updated with find-replace. I have a Linux/libvirt setup so someone else will have to validate.

aaronlevy:
vagrant/virtualbox work on linux....

aaronlevy:
Also, this is missing the generated templates (ci will fail)

dghubble:
Last I checked, not without disabling the kvm module, they don't coexist. Added the generated templates.

aaronlevy:
lgtm

Issue #96, opened 2016-08-05
Self-hosted kubelet breaks port-forwarding


chancez:
I found that trying to use `kubectl port-forward` does not work with the self-hosted cluster. I experienced this at first on 1.3.0, and then tried again after the 1.3.4 PR got merged earlier today. As far as I can tell this has likely always been an issue.

In my gist below, I try to port-forward prometheus, and that doesn't work. I try to port-forward the api-server, and that doesn't work. So I delete the kubelet daemonset, and then I port-forward the api-server again, this time it works.

https://gist.github.com/chancez/bb993765348612f6ee47b59e0e01a8c2

derekparker:
Potentially related: https://github.com/kubernetes/kubernetes/issues/19765

luxas:
see https://github.com/kubernetes/kubernetes/issues/26093

AntonioMeireles:
this is a tricky one. as  kubernetes/kubernetes#26093 clearly shows over time `kubelet` gained a fair share of external dependencies, some of which may come with a non trivial tail. Given the focus on slimness and lightness not all packages that own those external dependencies ship with CoreOS (`socat` is the issue here).  

on production this is easily solved by going the `hyperkube` way (as those dependencies stop being an CoreOS concern as they just become containerized) **but** here we have an issue because we are running things directly on top of CoreOS. 

So, IMHO, one of two routes would have to be paved - either one assumes that all (that matter to us) dependencies of k8s should ship with stock CoreOS (fattens the OS a bit, and is a non backward compatible trick) and the issue is over or change the tooling so that a containerized `bootkube` is run so that when it runs there would be worries of any kind about what userland is or isn't available.

chancez:
I'm fairly certain `socat` is in the hyperkube image, and we're running both `kubelet`s via the hyperkube image. So I'm fairly certain it isn't a problem of the image, but rather how the self hosted-kubelet is configured/ran. I'm guessing it has to do with the `nsenter` stuff that's done as the entrypoint to hyperkube, but I'm not certain.

derekparker:
So, the `nsenter` prefix is messing with PATH resolution. Specifically, we lose the binaries installed into the container filesystem when we break out of the container namespace.

We could:

1) Attempt to use the `--containerized` flag, which will prefix `mount` commands with `nsenter` so they take effect on the host namespace (note: this is marked as experimental).
[...]

PR #98, opened 2016-08-09
Fix issue #96 by modifying PATH

luxas:
Current working directory (`.`)  is container root `/` while `/` is host root.
PATH is by default /foo,/bar => all binaries are searched for on host.
This modifies PATH to look for binaries inside the container.
Also removed some host mounts, as they are unecessary.


chancez:
Not a fan. This is probably going to break disastrously due to shared libraries. Binaries in the containerFS will be using sharedlibs from the host's file system == bad.

luxas:
But right now we aren't using executables in the container at all, right?

The problem you mentioned _already exists_. 
If the host doesn't have glibc, kubelet won't come up because it's linked to libraries on host fs

chancez:
@luxas We are using executables in the container, nsenter, socat, mount (plus helpers), kubelet itself, etc. Everything currently uses the container's file system for tools, libraries, and binaries.

luxas:
@chancez No, you aren't using the container fs at all right now (except for hyperkube itself)
PATH points to /, and the same for everything else.
Everything is using host (even libraries that hyperkube uses); and that may or may not be a problem (depending on how you look at it)
Play with nsenter in order to see how it works.

Repro:
```console
docker run -it --net host --pid host --privileged gcr.io/google_containers/hyperkube-amd64:v1.3.0 /bin/bash
$ ls .
bin  boot  copy-addons.sh  dev  etc  home  hyperkube  lib  lib64  make-ca-cert.sh  media  mnt  nsenter  opt  proc  root  run  sbin  setup-files.sh  srv  sys  tmp  usr  var
$ ls /
bin  boot  copy-addons.sh  dev  etc  home  hyperkube  lib  lib64  make-ca-cert.sh  media  mnt  nsenter  opt  proc  root  run  sbin  setup-files.sh  srv  sys  tmp  usr  var
$ which socat
/usr/bin/socat
$ shasum /lib/x86_64-linux-gnu/libc.so.6
c73bbf1027c5e386f19d340c72f0a7b985adcafc  /lib/x86_64-linux-gnu/libc.so.6
$ /nsenter --target=1 --mount --wd=. -- /bin/bash
$ ls .
bin  boot  copy-addons.sh  dev  etc  home  hyperkube  lib  lib64  make-ca-cert.sh  media  mnt  nsenter  opt  proc  root  run  sbin  setup-files.sh  srv  sys  tmp  usr  var
$ ls /
bin  boot  cdrom  dev  etc  etcd  home  initrd.img  initrd.img.old  lib  lib32  lib64  lost+found  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var  vmlinuz  vmlinuz.old  volumes
$ which socat
$ ldd hyperkube
	linux-vdso.so.1 =>  (0x00007ffd6eda7000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f273c088000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f273be84000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f273bab9000)
	/lib64/ld-linux-x86-64.so.2 (0x00005584919dd000)
$ shasum /lib/x86_64-linux-gnu/libc.so.6
95675a95f296148d9a98c09ffc7c1e4b21876422  /lib/x86_64-linux-gnu/libc.so.6
```

The problem here is; actually we don't need the `nsenter` hack in the beginning if we could set mount propagnation of `/var/lib/kubelet` to shared in the Kube API, but we can't https://github.com/kubernetes/kubernetes/pull/20698

chancez:
Yeah, you're right. ideally, there is no nsenter at all, except for the `nsenter` the `kubelet` executes when calling `mount`. But we definitely _don't_ want to use anything on the host at all if possible. The main crux here is that `--containerized` will do what we want mostly, except it _will_ use the host's `mount` binary, rather than the containers. Everything else would be correctly be used from inside the container. 

When talking to @euank yesterday, we pretty much agreed mount propagation could help solve this, but we don't think that it's the completely correct approach. Instead we were thinking if we could just have some form of `--containerized` where we could ensure it uses the mount binary/shared libs from within the containerFS (instead of using the hosts like today), but doing it within the host's mount namespace via nsenter (same as `--containerized` works today).

PR #100, opened 2016-08-11
quickstart: Mount host resolv.conf into kubelet container

aaronlevy:
This should fix -- https://github.com/coreos/bootkube/issues/99 (but don't want github to auto-close issue until confirmed)


dghubble:
LGTM

PR #105, opened 2016-08-19
render: add --cloud-provider

kalbasit:


aaronlevy:
@kalbasit I opened https://github.com/coreos/bootkube/issues/106 to track some thoughts about how we can make configuration options like these potentially not require internal code changes. Let me know if you have any thoughts

kalbasit:
@aaronlevy I like the idea of #106 I'll think more about it and give my feedback there. `--cloud-provider` is going to be a very common flag, shouldn't be part of the default templates?

pkg/asset/templates/kube-apiserver.yaml
```diff
@@ -41,6 +41,9 @@ spec:
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
         - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
         - --client-ca-file=/etc/kubernetes/secrets/ca.crt
+        {{ with .CloudProvider -}}
```
kalbasit: @aaronlevy the `-}}` is a Go 1.6+ feature, see golang/go#9969. Is that OK?

kalbasit:
@aaronlevy bumping this to see if we can get it merged while waiting for a better solution than having to create a fork.

aaronlevy:
Hi @kalbasit 

This is being partially implemented in: https://github.com/kubernetes-incubator/bootkube/pull/186 -- but not yet exposed as a cli flag - as we might still need a bit more work for this to be plumbed through as a cli option. See my comments here: https://github.com/kubernetes-incubator/bootkube/pull/186#issuecomment-261591571

kalbasit:
@aaronlevy why re-implement the same in another PR?

aaronlevy:
If you want to make the changes here, happy to merge this PR (yours was first). I'll make inline comments for changes that would be needed.

Then as far as exposing the flag -- we can move forward with it, but I do have a concern that the temporary control plane will not reflect the same options (and I'm unsure if this is a safe assumption at this time). I also don't think adding another flag to `start` is a safe alternative, because then there is just an implied assumption that same flags are used between commands (for it to work), but no actual enforcement of this.

pkg/asset/internal/templates.go
```diff
@@ -51,6 +51,9 @@ spec:
         - --kubeconfig=/etc/kubernetes/kubeconfig
         - --require-kubeconfig
         - --lock-file=/var/run/lock/kubelet.lock
+        {{ with .CloudProvider -}}
+        - --cloud-provider={{ . }}
+        {{ end -}}
```
aaronlevy: We should these changes from the self-hosted kubelet manifest. Otherwise the self-hosted kubelet could behave differently than the on-host kubelet (as documented). The kubelet should use "auto-detect" for both -- so no change should be necessary.

pkg/asset/internal/templates.go
```diff
@@ -194,20 +200,29 @@ spec:
         - --root-ca-file=/etc/kubernetes/secrets/ca.crt
         - --service-account-private-key-file=/etc/kubernetes/secrets/service-account.key
         - --leader-elect=true
+        {{ with .CloudProvider -}}
+        - --cloud-provider={{ . }}
+        {{ end -}}
```
aaronlevy: nit: Drop the with/end blocks.
The default cloud-provider is an empty string anyway. `--cloud-provider={{ .CloudProvider }}`

aaronlevy: Should also add `- --configure-cloud-routes=false` which is safe to always be set (we assume flannel will handle managing routes).

kalbasit:
@aaronlevy I've been running a production cluster with this on since August and I had no problems so far.

I was thinking the other day about the flag and about #106, what if instead of exposing `cloud-provider` or any of the other flags, maybe we could expose `--extra-api-flags, --extra-controller-flags`, etc.. Thoughts?

pkg/asset/internal/templates.go
```diff
@@ -194,20 +200,29 @@ spec:
...
         volumeMounts:
         - name: secrets
           mountPath: /etc/kubernetes/secrets
           readOnly: true
         - name: ssl-host
           mountPath: /etc/ssl/certs
           readOnly: true
+        - name: resolv-conf
+          mountPath: /etc/resolv.conf
+          readOnly: true
```
aaronlevy: Lets use `dnsPolicy: Default` instead of the explicit host mount.

pkg/asset/k8s.go
```diff
@@ -35,7 +35,9 @@ func newStaticAssets(selfHostKubelet bool) Assets {
 
 func newDynamicAssets(conf Config) Assets {
 	return Assets{
+		mustCreateAssetFromTemplate(AssetPathControllerManager, internal.ControllerManagerTemplate, conf),
 		mustCreateAssetFromTemplate(AssetPathAPIServer, internal.APIServerTemplate, conf),
+		mustCreateAssetFromTemplate(AssetPathKubelet, internal.KubeletTemplate, conf),
```
aaronlevy: kubelet shouldn't be templated if we're not going to pass in the cloud-provider flag to it.

kalbasit:
@aaronlevy PTAL

pkg/asset/k8s.go
```diff
@@ -38,6 +38,7 @@ func newStaticAssets(selfHostKubelet bool) Assets {
 
 func newDynamicAssets(conf Config) Assets {
 	return Assets{
+		mustCreateAssetFromTemplate(AssetPathControllerManager, internal.ControllerManagerTemplate, conf),
```
aaronlevy: The controller-manager entry should also be removed from `newStaticAssets`

kalbasit: oops, must have bee a merge mistake at some point.

kalbasit:
@aaronlevy PTAL

aaronlevy:
lgtm. Thanks @kalbasit ! Can you squash your commits and I'll merge.

kalbasit:
@aaronlevy thanks for the quick review, very excited to have this merged (hard to maintain a fork :)). I squashed the commits.

@aaronlevy when will this make be released?

aaronlevy:
@kalbasit I'll try and get a new release out today.

kalbasit:
@aaronlevy thank you!

aaronlevy:
@kalbasit sorry for the delay -- trying to work out some networking issues I'm seeing on vagrant/kube-proxy so want to fully understand issue before cutting a release

kalbasit:
@aaronlevy thank you for letting me know. I need the release at the beginning of December so please take your time and ping me once it has been released.

PR #107, opened 2016-08-24
pkg/asset: Update kube-dns to v17.1

### LINKED ISSUE #93: Use latest kube-dns
kube-dns got a rewrite in 1.3, and is required for features like federation to work properly. It's also lighter weight/simpler. Bootkube is deploying the old kube-dns, it should get upgraded (also while at it the file should be renamed to a -deployment suffix since its no longer an RC).

chancez:
Closes #93


aaronlevy:
Thank you for updating this!

Can we have it track the v1.3 release version (rather than head)? Looks like that's still on v17.1:
https://github.com/kubernetes/kubernetes/blob/v1.3.5/cluster/saltbase/salt/kube-dns/skydns-rc.yaml.base#L34

pkg/asset/templates/kube-dns-deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+            path: /readiness
+            port: 8081
+            scheme: HTTP
+          # we poll on pod startup for the Kubernetes master service and
+          # only setup the /readiness HTTP server once that's available.
+          initialDelaySeconds: 30
+          timeoutSeconds: 5
+        args:
+        # command = "/kube-dns"
+        - --domain=cluster.local
```
aaronlevy: Also we should keep the trailing "." here (easy to miss): https://github.com/kubernetes/kubernetes/blob/v1.3.5/cluster/saltbase/salt/kube-dns/skydns-rc.yaml.base#L76

chancez:
@aaronlevy switched it over to v17.1

aaronlevy:
@chancez mind updating this again? templating has changed...

dghubble:
He's on vacation. We're seeing flakiness in v11 and would like to update to v17.1. #126

aaronlevy:
fixed in: https://github.com/kubernetes-incubator/bootkube/pull/126

PR #108, opened 2016-08-26
[WIP] Self-host Flannel

kenan435:
Flannel is now self hosted when provisioning a single node vagrant setup - only. Before making it work for other scenarios some feedback would be appreciated if this PR is heading the right direction.


aaronlevy:
@kenan435 I'm actually working on the same thing!

The direction we plan on taking actually makes use of some new features both in kubernetes and flannel:

Periodically reload the CNI configuration : https://github.com/kubernetes/kubernetes/pull/28178
This makes it so we don't necessarily need to have CNI configuration exist prior to kubelet deployment (and can instead deploy pod network tools as a daemonset).

Add a "kubernetes manager" to flannel: https://github.com/coreos/flannel/pull/483
This will allow flannel to use the kubernetes assigned podCIDRs and store backend information as annotations on the pod object.

Another big motivation for the above is that it removes the need for all nodes to have direct access to the cluster etcd.

The WIP is here if you want to take a look: 
All changes: https://github.com/coreos/bootkube/pull/113
Self-hosting flannel: https://github.com/coreos/bootkube/pull/113/commits/9c275e89f8c982ff165faf522efc3fc9f5118018

PR #109, opened 2016-08-29
Cluster updater

derekparker:
This PR is based on the node-agent PR, so check out the last commit for the update controller specific patch.


Makefile
```diff
@@ -32,6 +34,14 @@ _output/bin/%/checkpoint: cmd/checkpoint/main.go
 	mkdir -p $(dir $@)
 	GOOS=$* go build -o _output/bin/$*/checkpoint github.com/coreos/bootkube/cmd/checkpoint
 
+_output/bin/%/node-agent: cmd/node-agent/main.go pkg/node/*.go
```
aaronlevy: nit (can be follow-up): using a specific dependency like `pkg/node*.go` seems a little fragile. It would mean if we happened to add another package to the node-agent we would also need to update this. It may make more sense just to use $(GOFILES) even if it is more heavy-handed.

Also, it probably would be beneficial to have these tools be versioned as well - which means we could hypothetically have the same prereqs for all of these tools and combine into single make target (untested):

```
_output/bin/%: LDFLAGS=-X github.com/coreos/bootkube/pkg/version.Version=$(shell $(CURDIR)/build/git-version.sh)
_output/bin/%: $(GOFILES) pkg/asset/internal/templates.go
	mkdir -p $(dir $@)
	GOOS=$* go build -ldflags "$(LDFLAGS)" -o $@ github.com/coreos/bootkube/cmd/$(notdir $@)
```

cmd/node-agent/README.md
```diff
@@ -0,0 +1,87 @@
+# Node Agent
+
+## Description
+
+The Node Agent is a process that runs on each node of a Kubernetes cluster. The
+purpose of this process is to monitor a specific annotation on the node object.
+This annotation holds a JSON object which describes the version and configuration
```
aaronlevy: drop wording about "configuration"

cmd/node-agent/README.md
```diff
@@ -0,0 +1,87 @@
...
+to be used for the on-host kubelet.
+
+This is useful for upgrading your on-host kubelet without ever having to ssh
+into a machine. Simply update the annotation with the desired version, and the
+name of the configMap that holds your configuration, and the Node Agent will
```
aaronlevy: drop wording about configuration/configMap

cmd/node-agent/README.md
```diff
@@ -0,0 +1,87 @@
+# Node Agent
```
aaronlevy: Rather than comment on all lines, just making note here: we should drop the stuff about configMaps now that we're going to rely on the upstream functionality

cmd/node-agent/main.go
```diff
@@ -0,0 +1,89 @@
...
+	clientset "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_3"
+	"k8s.io/kubernetes/pkg/client/restclient"
+	"k8s.io/kubernetes/pkg/controller/framework"
+	"k8s.io/kubernetes/pkg/fields"
+	"k8s.io/kubernetes/pkg/runtime"
+	"k8s.io/kubernetes/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+const myPodNameEnvName = "MY_POD_NAME"
```
aaronlevy: nit: s/MY_POD_NAME/POD_NAME/ -- not sure why the MY_ bugs me so

cmd/node-agent/main.go
```diff
@@ -0,0 +1,89 @@
...
+
+func main() {
+	myPodName := os.Getenv(myPodNameEnvName)
+	if myPodName == "" {
+		glog.Fatalf("%s env var not set.", myPodNameEnvName)
+	}
+
+	client := newAPIClient()
+	nodename, err := getNodeName(client, myPodName)
```
aaronlevy: Can you add a TODO note here to switch to using nodename directly once we're on v1.4

cmd/node-agent/main.go
```diff
@@ -0,0 +1,89 @@
...
+	glog.Infof("starting node agent, watching node: %s", nodename)
+
+	a := node.Agent{
+		NodeName: nodename,
+		Client:   client,
+		SysdConn: sysdConn,
+	}
+
+	nodeopts := api.ListOptions{
+		FieldSelector: fields.ParseSelectorOrDie(fmt.Sprintf("metadata.name=%s", nodename)),
```
aaronlevy: Newly learned from another PR: `fields.OneTermEqualSelector("metadata.name", nodename)` - but either works.

cmd/update-controller/main.go
```diff
@@ -0,0 +1,147 @@
...
+	"k8s.io/kubernetes/pkg/controller/framework"
+	"k8s.io/kubernetes/pkg/fields"
+	"k8s.io/kubernetes/pkg/runtime"
+	"k8s.io/kubernetes/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+type versionPayload struct {
+	newVersion *cluster.ContainerImage
+	configMap  *v1.ConfigMap
```
aaronlevy: We should drop this now that we're going to rely on the upstream functionality

cmd/update-controller/main.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	return &cluster.ContainerImage{
+		Repo: parts[0],
+		Tag:  parts[1],
+	}, nil
+}
+
+// TODO: This could potentially be better acheived by storing this in the configmap object that
+// we get the version from. If it does not exist, we can create it.
+func getCurrentVersion(client clientset.Interface) (*cluster.ContainerImage, error) {
```
aaronlevy: I don't think its safe to assume api-server version == cluster version. It's possible this was updated but the rest of the components have not. Need to think on this for a bit to see how we could track this status.

cmd/update-controller/main.go
```diff
@@ -0,0 +1,147 @@
...
+				if err := cu.Update(); err != nil {
+					glog.Error(err)
+				}
+			} else {
+				glog.Info("no new version, nothing to do")
+			}
+		}
+	}(client)
+	opts := api.ListOptions{
+		FieldSelector: fields.ParseSelectorOrDie(fmt.Sprintf("metadata.name=%s", cluster.ClusterConfigMapName)),
```
aaronlevy: `fields.OneTermEqualSelector("metadata.name", cluster.ClusterConfigMapName)`

hack/multi-node/user-data.sample
```diff
@@ -19,36 +19,49 @@ coreos:
           [Unit]
           Requires=flanneld.service
           After=flanneld.service
+    - name: kubelet-initial-config.service
```
aaronlevy: Can drop this now that we're not handling the config.

hack/single-node/user-data.sample
```diff
@@ -20,37 +20,49 @@ coreos:
           [Unit]
           Requires=flanneld.service
           After=flanneld.service
+    - name: kubelet-initial-config.service
```
aaronlevy: Can be dropped

pkg/asset/templates/kube-apiserver.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
     k8s-app: kube-apiserver
     version: v1.3.6_coreos.0
```
aaronlevy: I know you wanted to do this before: until we handle updating label (which I'd rather not bother with right now), we should probably remove these (otherwise image version won't match label anymore)

pkg/asset/templates/kube-scheduler.yaml
```diff
@@ -7,6 +7,10 @@ metadata:
     k8s-app: kube-scheduler
     version: v1.3.6_coreos.0
 spec:
+  selector:
+    matchLabels:
+      k8s-app: kube-scheduler
+  replicas: 2
```
aaronlevy: Any particular reason to bump this up initially?

pkg/asset/templates/kubelet.yaml
```diff
@@ -17,22 +20,9 @@ spec:
...
-        - --allow-privileged
-        - --hostname-override=$(MY_POD_IP)
-        - --cluster-dns=10.3.0.10
-        - --cluster-domain=cluster.local
-        - --kubeconfig=/etc/kubernetes/kubeconfig
-        - --lock-file=/var/run/lock/kubelet.lock
-        - --minimum-container-ttl-duration=3m0s
+        - /bin/bash
+        - -c
+        - "source /etc/kubernetes/kubelet.env; /nsenter --target=1 --mount --wd=. -- ./hyperkube kubelet --hostname-override=$(MY_POD_IP) $KUBELET_OPTS"
```
aaronlevy: Let's switch this back now that we're going to use upstream work for updating kubelet config

cmd/update-controller/main.go
```diff
@@ -0,0 +1,147 @@
...
+	"k8s.io/kubernetes/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+type versionPayload struct {
+	newVersion *cluster.ContainerImage
+	configMap  *v1.ConfigMap
+}
+
+func main() {
```
aaronlevy: This component should be leader-elected. Can pretty much just pull same functionality from scheduler or controller-manager mains.

pkg/asset/templates/kube-apiserver.yaml
aaronlevy: Nevermind, looks like you are updating the version labels

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+			glog.Infof("updating image for container: %s", c.Name)
+			ds.Spec.Template.Spec.Containers[i].Image = image.String()
+			break
+		}
+	}
+	ds, err = client.Extensions().DaemonSets(api.NamespaceSystem).Update(ds)
+	if err != nil {
+		return err
+	}
+	l, err := labels.Parse(fmt.Sprintf("k8s-app=%s", name))
```
aaronlevy: This is a bit fragile. We should be able to use the daemonset selector to get pods managed by this daemonset:

```
	selector, err := unversioned.LabelSelectorAsSelector(ds.Spec.Selector)
	if err != nil {
		return nil, err
	}
	daemonPods, err := client.Core().Pods(ds.Namespace).List(selector)
	if err != nil {
		return nil, err
	}
```

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+	pl, err := client.Core().Pods(api.NamespaceSystem).List(lo)
+	if err != nil {
+		return err
+	}
+	for _, p := range pl.Items {
+		// Delete old DS Pod.
+		glog.Infof("Deleting pod %s", p.Name)
+		err = client.Core().Pods(api.NamespaceSystem).Delete(p.Name, nil)
+		if err != nil && err != io.EOF {
```
aaronlevy: Why ignoring EOF error?

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+			glog.Info("Waiting for new pod to come up...")
+			upl, err := client.Core().Pods(api.NamespaceSystem).List(lo)
+			if err != nil {
+				// Do not return an error here, since we have just deleted a pod,
+				// it is quite possible we have deleted an API Server pod. We may fail
+				// to contact it for a bit until the checkpointer API Server takes over.
+				glog.Error(err)
+			} else {
+				if ds.Status.DesiredNumberScheduled == int32(len(upl.Items)) {
+					if allContainersRunning(upl) {
```
aaronlevy: You could use upstream logic here: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L629

Main differences are looking at the pod status transition time (I'm assuming in case the last status was never updated) and looking at podReady condition: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L651

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+						if name == "kube-apiserver" {
```
aaronlevy: This may be unnecessary if we change the above to check podReady condition (if api-server has proper health checks on the pod itself).

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+		glog.Infof("Deleting pod %s", p.Name)
+		err = client.Core().Pods(api.NamespaceSystem).Delete(p.Name, nil)
+		if err != nil && err != io.EOF {
+			return err
+		}
+		glog.Infof("Deleted pod %s", p.Name)
+		// Wait for new DS Pod.
+		for {
+			glog.Info("Waiting for new pod to come up...")
+			upl, err := client.Core().Pods(api.NamespaceSystem).List(lo)
```
aaronlevy: Similar to above, we should get pods based on the daemonset we're updating (via selector).

aaronlevy: Also this might be better implemented as a Watch() in a loop rather than list/loop. Probably minor assuming there aren't usually going to be that many pods in this namespace, but I guess in some scenarios it could be more expensive.

If we're pretty regularly asking for sets of pods, might make sense to just setup a nodeStore and perform the list operations against that.

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+			if err != nil {
+				// Do not return an error here, since we have just deleted a pod,
+				// it is quite possible we have deleted an API Server pod. We may fail
+				// to contact it for a bit until the checkpointer API Server takes over.
+				glog.Error(err)
+			} else {
+				if ds.Status.DesiredNumberScheduled == int32(len(upl.Items)) {
```
aaronlevy: Unless we're periodically updating the ds object, it's possible to get into a scenario where the desiredNumberScheduled will never match the number running. Consider:

3 nodes, all should run ds pod.
1 node permanently deleted while an upgrade is in process.
This will now run waiting for desired=3 which will never be the case.

So we should either periodically refresh, or let this time out and retry.

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+			break
+		}
+	}
+	// Update the deployment, which will trigger an update.
+	_, err = client.Extensions().Deployments(api.NamespaceSystem).Update(dep)
+	if err != nil {
+		return err
+	}
+	// Wait for update to finish.
+	l, err := labels.Parse(fmt.Sprintf("version=%s,k8s-app=%s", image.Tag, name))
```
aaronlevy: Can use some upstream code for this: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L621

But make sure to use an updated deployment object (I would base it off an api server copy, not your local copy - as it might not have been updated for some reason - could wait on the observed generation to make sure:

https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L750

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+			glog.Error(err)
+		} else {
+			if int32(len(upl.Items)) == replicas {
+				if allContainersRunning(upl) {
+					return nil
+				}
+			}
+			glog.Infof("not all new containers have started, expected %d got %d", dep.Status.Replicas, len(upl.Items))
+		}
+		time.Sleep(time.Second)
```
aaronlevy: We need some kind of release valve here - we don't want it to be possible just to sit in this state forever.

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+			return err
+		}
+		n.Annotations[node.DesiredConfigAnnotation] = string(b)
+		_, err = client.Core().Nodes().Update(&n)
+		if err != nil {
+			return err
+		}
+		lo := api.ListOptions{
+			Watch:         true,
+			FieldSelector: fields.ParseSelectorOrDie(fmt.Sprintf("metadata.name=%s", n.Name)),
```
aaronlevy: nit: `fields.OneTermEqualSelector("metadata.name", n.Name)`

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+		w, err := client.Core().Nodes().Watch(lo)
```
aaronlevy: You may need to establish this watch prior to the update operation, otherwise there is a window for you to miss the node update. But... nodes objects are frequently updated (kubelet will update nodeStatus so we know node is still alive) - where we would catch the change. So if we're okay for waiting for that next window, this may be a non-issue.

So choice is - establish watch prior to making the update
or, rely on the node status update window (should check how long this is) -- but that could artificially delay this process.

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+			case watch.Modified:
+				nn, ok := e.Object.(*v1.Node)
+				if !ok {
+					glog.Infof("recieved unexpected type from Node watch. Expected *v1.Node, got: %T", e.Object.GetObjectKind())
+					continue
+				}
+				if nn.Annotations[node.DesiredConfigAnnotation] == nn.Annotations[node.CurrentConfigAnnotation] {
+					break
+				}
+				glog.Infof("Node %s update completed", nn.Name)
```
aaronlevy: You won't actually hit this on node update complete (break above)

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+		w, err := client.Core().Nodes().Watch(lo)
+		if err != nil {
+			return err
+		}
+		for e := range w.ResultChan() {
+			switch e.Type {
+			case watch.Modified:
+				nn, ok := e.Object.(*v1.Node)
+				if !ok {
+					glog.Infof("recieved unexpected type from Node watch. Expected *v1.Node, got: %T", e.Object.GetObjectKind())
```
aaronlevy: errorf?

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+				}
+				if nn.Annotations[node.DesiredConfigAnnotation] == nn.Annotations[node.CurrentConfigAnnotation] {
+					break
+				}
+				glog.Infof("Node %s update completed", nn.Name)
+				break
+			case watch.Error:
+				aerr, ok := e.Object.(*unversioned.Status)
+				if !ok {
+					glog.Infof("recieved unexpected error type from Node watch. Expected *api.Status, got: %T", e.Object.GetObjectKind())
```
aaronlevy: errorf?

pkg/cluster/updater.go
```diff
@@ -0,0 +1,372 @@
...
+	for _, c := range cu.Components {
+		cu.Events.Eventf(cu.SelfObj, api.EventTypeNormal, "Begin update of component: %s", c.Name)
+		glog.Infof("Begin update of component: %s", c.Name)
+		if err := c.UpdateFn(cu.Client, c.Name, cu.NewVersion); err != nil {
+			cu.Events.Eventf(cu.SelfObj, api.EventTypeWarning,
+				"Failed update of component: %s, rolling back to version: %s from %s",
+				c.Name, cu.OldVersion, cu.NewVersion)
+			glog.Errorf("Failed update of component: %s, rolling back to version: %s from %s due to: %v",
+				c.Name, cu.OldVersion, cu.NewVersion, err)
+			return cu.Rollback()
```
aaronlevy: I think this should be an operator choice to rollback - not a rollback on any error. There are going to be transient network errors that the updateFn could hit, and we don't want to immediately roll back a cluster.

pkg/node/agent.go
```diff
@@ -0,0 +1,215 @@
...
+	DesiredConfigAnnotation = "node-agent.alpha.coreos.com/desired-config"
+	// CurrentConfigAnnotation is the annotation that describes the current
+	// configuration state for the on-host kubelet.
+	CurrentConfigAnnotation = "node-agent.alpha.coreos.com/current-config"
+	// KubeletVersionKey is the key in the JSON stored in the config annotations that
+	// describes the version of the kubelet to run on the host.
+	KubeletVersionKey = "KUBELET_VERSION"
+	// KubeletConfigKey is the key in the JSON stored in the config annotations that
+	// describes the configuration flags for the kubelet to run on the host.
+	KubeletConfigKey = "KUBELET_CONFIG"
```
aaronlevy: drop this (using upstream dynamic config instead)

pkg/node/agent.go
```diff
@@ -0,0 +1,215 @@
...
+		return nil, fmt.Errorf("error attempting to marshal current config: %s", err)
+	}
+	newNode.Annotations[CurrentConfigAnnotation] = string(newConf)
+	return &newNode, nil
+}
+
+func (a *Agent) getDesiredConfig(node *v1.Node) (map[string]string, error) {
+	rawconf, ok := node.Annotations[DesiredConfigAnnotation]
+	if !ok {
+		err := fmt.Errorf("no %s annotation found for node %s, ignoring", DesiredConfigAnnotation, node.Name)
```
aaronlevy: nit: If we're just ignoring, maybe log.info and exit with err=nil

pkg/node/agent.go
```diff
@@ -0,0 +1,215 @@
...
+	// Check annotations on Node.
+	if node.Annotations == nil {
+		node.Annotations = make(map[string]string)
+	}
+
+	var desiredConfig map[string]string
+	onDiskConfig, err := parseKubeletEnvFile(defaultKubeletEnvPath)
+	if err != nil {
+		glog.Error(err)
+		goto update
```
aaronlevy: Why is this necessary? Can't we just put the `getDesiredConfig` below in an else block?

pkg/node/agent.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	// Obtain and validate the desired configuration.
+	desiredConfig, err = a.getDesiredConfig(node)
+	if err != nil {
+		glog.Error(err)
+	}
+
+update:
+	updatedNode, err := a.handleConfigUpdate(*node, onDiskConfig, desiredConfig, defaultKubeletEnvPath)
```
aaronlevy: `desiredConfig` could be nil here, and looks like this could cause nil pointer error in `validateConfig`

pkg/node/agent.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+}
+
+func (a *Agent) handleConfigUpdate(node v1.Node, onDiskConfig, desiredConfig map[string]string, kubeletEnvPath string) (*v1.Node, error) {
+	// `node` is already a copy, but just to make things clear, assign to new var
+	newNode := node
+
+	err := validateConfig(desiredConfig)
+	if err != nil {
+		glog.Error(err)
```
aaronlevy: okay to continue past error?

pkg/node/agent.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	// Check desired config against the on disk config.
+	if err == nil && configHasChanged(onDiskConfig, desiredConfig) {
+		// Update on disk config.
+		onDiskConfig, err = updateKubeletEnvFile(kubeletEnvPath, desiredConfig, onDiskConfig)
+		if err != nil {
+			// Log error but continue, we still want to update config.
```
aaronlevy: Why do we want to restart kubelet if we get an error updating the config?

pkg/node/agent.go
```diff
@@ -0,0 +1,215 @@
...
+		// Restart kubelet via systemd to pick up new config.
+		if err = a.restartKubeletService(); err != nil {
+			// Log error but continue, we still want to update config.
+			glog.Error(err)
+		}
+	}
+
+	configName := desiredConfig[KubeletConfigKey]
+	if configName == "" {
+		configName = "on-disk configuration"
```
aaronlevy: nit: use const for static values

aaronlevy:
Some general thoughts:

#### Node agent config

In node-agent we're essentially now just tracking version (not flags) - and I'm wondering if we should drop the use of json objects for simpler annotations. For example:

`node-agent.alpha.coreos.com/desired-config = '{"KUBELET_VERSION": "v1.3.6_coreos.0"}'`
vs.
`node-agent.alpha.coreos.com/kubelet-version = v1.3.6_coreos.0`

We may want to add ability to specify repo too:

`node-agent.alpha.coreos.com/desired-config = '{"KUBELET_VERSION": "v1.3.6_coreos.0", "KUBELET_REPO="quay.io/coreos/hyperkube"}'`
vs.
`node-agent.alpha.coreos.com/kubelet-version = v1.3.6_coreos.0`
`node-agent.alpha.coreos.com/kubelet-repo = quay.io/coreos/hyperkube`

(fwiw I'm leaning toward the separated annotations)

#### Updater responsibility

Talked a bit about this offline, but want to capture some of the discussion here. Right now the updater determines what is is responsible for based on a fixed set of components. But this is rather fragile / inflexible (people may not self-host a kubelet, for example).

I  think we can make this more similar to how, for example, a deployment works: the deployment is able to query the replicasets and the pods which it is responsible for based on selectors. Similarly we should be able to query the cluster and ask what components the updater is responsible for.

One option would be to handle this with annotations on objects we wish to manage:

`update-controller.alpha.coreos.com/managed=true`
`update-controller.alpha.coreos.com/update-priority=99`

The update-controller can then query the cluster and ask for all daemonsets & deployments which should be managed, then order their updates based on priority (e.g. apiserver first, etc.)

As far as the node-agent, we could either just let it set desiredConfig annotations on node objects always (if node-agent isn't running, nothing will happen), or this could be a global configurable set on the update-agent itself.

#### Updater flow

We are currently triggering an update based on a change to a configMap (or soon thirdPartyResource) - but I'm starting to think we might want to instead be constantly asking the cluster "what am I responsible for, and is it the expected version" - rather than "has version changed, now I'll try and make it match then die". So the flow essentially becomes:

```
while true
    get desired version
    err = update(version)
    sleep 1

func update()
    get all daemonsets + deployments which are managed
    for each component (by priority) attempt to update version
        if err; return err 
```

There are likely going to be transient errors, so our update is just continually starting back over and trying to make sure we're always converging on the desired version.

Along these lines I also don't think a rollback should be automatic, and instead we leave that up to the admin. If the failure is not transient and something is going wrong, it should just error at the same point and essentially functions as a "pause until fixed". I also think we could add a "pause" option to the thirdparty resource. Doesn't have to be super fine-grained, but maybe check at start and between components - and this gives some "inspection" time to an admin if something seems off (rather than the only option being killing the updater or immediate rollback).

pkg/asset/templates/kube-scheduler.yaml
derekparker: It keeps the update code more generic. I could add some logic where if replicas == 1, first scale up 2 to, update, and then scale back down.

rochacon: I believe we may achieve this by setting `stategy.rollingUpdate.maxUnavailable = 0`, right?

derekparker: @rochacon good point, that may well work.

cmd/update-controller/main.go
derekparker: How do you feel about (which we discussed a bit earlier) but the highest version of a resource running in the cluster (manged by the update-controller) is considered the "cluster version"?

derekparker: Even that get's a bit tricky and is disingenuous, though.

derekparker: (sorry to stack comments) But it actually won't REALLY matter, especially now that we're going to check all components anyways I'm going to remove this simplified check, and then we just converge on the desired version, without trying to determine a current cluster version.

cmd/node-agent/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Description
```
mikedanese: cc @vishh, you talked about something similar

cmd/checkpoint/main.go
```diff
@@ -69,6 +71,7 @@ func run() {
 			if err := os.Remove(activeManifest); err != nil {
 				glog.Error(err)
 			}
+			sleepDuration = 5 * time.Minute
```
aaronlevy: Why the change in duration here?

cmd/node-agent/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+The Node Agent is a process that runs on each node of a Kubernetes cluster. The
+purpose of this process is to monitor a specific annotation on the node object.
+This annotation holds a JSON object which describes the version to be used for
```
aaronlevy: This annotation describes the version to be used...

cmd/node-agent/README.md
```diff
@@ -0,0 +1,58 @@
...
+the on-host kubelet.
+
+This is useful for upgrading your on-host kubelet without ever having to ssh
+into a machine. Simply update the annotation with the desired version, and the
+and the Node Agent will ensure that the on-host kubelet matches that value,
```
aaronlevy: s/and the and the/and the/

cmd/update-controller/main.go
```diff
@@ -0,0 +1,98 @@
...
+	clientset "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_3"
+	"k8s.io/kubernetes/pkg/client/restclient"
+	"k8s.io/kubernetes/pkg/controller/framework"
+	"k8s.io/kubernetes/pkg/fields"
+	"k8s.io/kubernetes/pkg/runtime"
+	"k8s.io/kubernetes/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+func main() {
```
aaronlevy: This component needs to be leader elected

cmd/update-controller/main.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		glog.Fatal(err)
+	}
+	run(clientset.NewForConfigOrDie(config))
+}
+
+func run(client clientset.Interface) {
+	glog.Info("update controller running")
+	versionChan := make(chan *components.Version)
+	handleConfigChange := func(newConfigMap *v1.ConfigMap) {
```
aaronlevy: Per in-person discussion, we should change the version configuration to a thirdPartyResource

cmd/update-controller/main.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	go func(client clientset.Interface, updater *cluster.ClusterUpdater) {
+		for v := range versionChan {
+			if err := updater.UpdateToVersion(v); err != nil {
+				glog.Error(err)
+			}
+		}
+	}(client, cu)
+	opts := api.ListOptions{
+		FieldSelector: fields.ParseSelectorOrDie(fmt.Sprintf("metadata.name=%s", cluster.ClusterConfigMapName)),
```
aaronlevy: nit: fields.OneTermEqualSelector("metadata.name", cluster.ClusterConfigMapName)

cmd/update-controller/main.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		handleConfigChange(newConfigMap)
+	}
+	cu, err := cluster.NewClusterUpdater(client)
+	if err != nil {
+		glog.Error(err)
+		return
+	}
+	go func(client clientset.Interface, updater *cluster.ClusterUpdater) {
+		for v := range versionChan {
```
aaronlevy: I feel like queuing up version changes could be problematic. Ideally we are only targeting the "currently set" version -- not what might have been set and happens to be queued.

Let's start with just a single process targeting a single version, and periodically re-running (and checking desired version each time). As follow up in functionality, we should allow an upgrade to be stopped mid-process rather than having to wait to completion to reset / change desired.

hack/multi-node/user-data.sample
```diff
@@ -19,36 +19,49 @@ coreos:
...
-          --lock-file=/var/run/lock/kubelet.lock \
-          --exit-on-lock-contention \
-          --config=/etc/kubernetes/manifests \
-          --allow-privileged \
-          --hostname-override=${COREOS_PUBLIC_IPV4} \
-          --node-labels=master=true \
-          --minimum-container-ttl-duration=3m0s \
-          --cluster_dns=10.3.0.10 \
-          --cluster_domain=cluster.local
+        ExecStart=/usr/lib/coreos/kubelet-wrapper $PERSISTENT_KUBELET_OPTS $KUBELET_OPTS
```
aaronlevy: Change this back

hack/multi-node/user-data.sample
```diff
@@ -19,36 +19,49 @@ coreos:
...
         Restart=always
         RestartSec=5
 
         [Install]
         WantedBy=multi-user.target
 
 write_files:
+  - path: "/home/core/install-initial-kubelet-config"
```
aaronlevy: Can drop this

hack/single-node/user-data.sample
```diff
@@ -51,6 +63,16 @@ coreos:
         WantedBy=multi-user.target
 
 write_files:
+  - path: "/home/core/install-initial-kubelet-config"
```
aaronlevy: Can drop

hack/single-node/version.yaml
```diff
@@ -0,0 +1,7 @@
+kind: ConfigMap
```
aaronlevy: Need to change to thirdPartyResource

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,178 @@
...
+				return client.Core().Pods(ds.Namespace).Watch(lo)
+			},
+		},
+		&v1.Pod{},
+		30*time.Minute,
+		framework.ResourceEventHandlerFuncs{},
+	)
+	go podController.Run(wait.NeverStop)
+	return &DaemonSetUpdater{
+		name:     ds.Name,
```
aaronlevy: We should also track namespace rather than assuming 'kube-system' (and we have that info available here)

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if v == nil {
+		return nil, fmt.Errorf("unable to get current version for DaemonSet %s", dsu.Name())
+	}
+	return v, nil
+}
+
+// UpdateToVersion will update the DaemonSet to the given version.
+func (dsu *DaemonSetUpdater) UpdateToVersion(client clientset.Interface, v *Version) error {
+	ds, err := client.Extensions().DaemonSets(api.NamespaceSystem).Get(dsu.Name())
```
aaronlevy: After tracking ds namespace, use it here.

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return err
+	}
+	// Create new DS.
+	ds.Labels["version"] = v.Image.Tag
+	ds.Spec.Template.Labels["version"] = v.Image.Tag
+	for i, c := range ds.Spec.Template.Spec.Containers {
+		if c.Name == dsu.Name() {
```
aaronlevy: This implicit assumption is a bit fragile - but I don't have any immediate ideas for how we should handle this. Can you add a TODO: figure out how to determine which container should be update? Or maybe it's just a matter of us documenting this convention

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,178 @@
...
+			glog.Infof("updating image for container: %s", c.Name)
+			ds.Spec.Template.Spec.Containers[i].Image = v.Image.String()
+			break
+		}
+	}
+	ds, err = dsu.client.Extensions().DaemonSets(api.NamespaceSystem).Update(ds)
```
aaronlevy: nit: we're not re-using ds after this, so:  `_, err = dsu.client...`

aaronlevy: s/api.NamespaceSystem/ds.Namespace

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return err
+	}
+	pods := dsu.podStore.List()
```
aaronlevy: The way this is structured we are setting up and tearing down all of these pod stores between each update. I'll add more thoughts as a general comment, but I think this needs to be restructured - or we don't bother with the podStore and just request directly from api

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,178 @@
...
+	for i, pi := range pods {
+		p := pi.(*v1.Pod)
+		// Delete old DS Pod.
+		glog.Infof("Deleting pod %s", p.Name)
+		err = dsu.client.Core().Pods(api.NamespaceSystem).Delete(p.Name, nil)
```
aaronlevy: How do we know this needs to be deleted? It could have already been updated.

aaronlevy: We should probably do something like compare the ds.podSpec.Container.Image to the pod.Container.Image

hack/multi-node/user-data.sample
derekparker: Still need it since we're setting the version via that file and not a drop-in.

hack/single-node/user-data.sample
derekparker: Still need this for version.

pkg/asset/templates/kube-scheduler.yaml
derekparker: done

pkg/asset/templates/kubelet.yaml
derekparker: done

cmd/checkpoint/main.go
derekparker: Mostly just to give the Kubelet a bit more time to kill the static pod and let he managed pod run before we try to inspect state again. Not strictly needed, added this investigating something else.

pkg/cluster/components/version.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/blang/semver"
+)
+
+// Version represents versioned cluster information,
+// including the semver information and container
+// image information.
+type Version struct {
+	// Semver is the semver parsed version for comparisons.
+	Semver semver.Version
```
aaronlevy: nit: Having both `Semver` and `Image` be public means they could become out of sync. Maybe we just expose a `Semver()` method which extracts from Image.

pkg/cluster/updater.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+	return nil
+}
+
+func getManagedComponentList(client clientset.Interface) ([]ComponentUpdater, error) {
+	var comps []ComponentUpdater
+	lo := api.ListOptions{FieldSelector: fields.OneTermEqualSelector(clusterManagedAnnotation, "true")}
+	// Get DaemonSets
+	dsl, err := client.Extensions().DaemonSets(api.NamespaceSystem).List(lo)
```
aaronlevy: I think we still need the priority annotation here. Otherwise it's possible, for example, to update kubelet or kube-proxy prior to the apiserver

aaronlevy:
Formatting was all messed up above, and I don't think I can change my review comment, so reposting here:

General thoughts:

The per-component updaters / stores I think add more complexity than is needed - and also removes a lot of the benefit of using the informer/stores anyway - as they are being thrown out on each update loop. Instead I'm envisioning an "updateController" which can have informers for the objects it cares about, which can then be passed to stateless updaters. We can start really general and optimize later. For example:

```
struct UpdateController {
    client *clientset.Interface
    podStore *cache.Store
    nodeStore *cache.Store
    deploymentStore *cache.Store
    daemonsetStore *cache.Store
}

func main() {
    uc := NewUpdateController()
    for {
        v := getDesiredVersion()
        uc.Update(v)
        sleep 30
    }
}

func NewUpdateController() *UpdateController {
    // setup all informers / stores
    return &uc
}

func (uc *UpdateController) Update(version) error {
    cs = getComponentsByPriority()
    for c := range cs {
        switch c.Kind
        case Daemonset:
            err = uc.UpdateDaemonset(c, version)
        case Deployment:
            err = uc.UpdateDeployment(c, version)
        case ... 
    }
}

func (uc *UpdateController) UpdateDaemonset(ds *v1beta1.Daemonset, version string) error {
    pods = uc.podStore.List(by ds.Selector)
    ...
}
```

The above would simplify a lot, and allow us to just query our global cache's. We could later try and optimize the caches by restricting what they're watching (could just start with 'kube-system' namespace too).

We can also get more complex with things like "wake up on version config change" or start/stop/resume functionality -- but lets start as simple as possible (essentially serially check desired version -- try once and sleep)

pkg/cluster/components/deployment.go
```diff
@@ -0,0 +1,82 @@
...
+	for _, c := range dp.Spec.Template.Spec.Containers {
+		if c.Name == du.Name() {
+			return ParseVersionFromImage(c.Image)
+		}
+	}
+	return nil, fmt.Errorf("unable to get current version for Deployment %s", du.Name())
+}
+
+// UpdateToVersion will update the Deployment to the given version.
+func (du *DeploymentUpdater) UpdateToVersion(client clientset.Interface, v *Version) error {
```
derekparker: Oversight, fixed.

pkg/cluster/components/deployment.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return deployment.WaitForObservedDeployment(func() (*extensions.Deployment, error) {
+		dp, err := du.client.Extensions().Deployments(api.NamespaceSystem).Get(du.Name())
+		if err != nil {
+			return nil, err
+		}
+		var out extensions.Deployment
+		v1beta1.Convert_v1beta1_Deployment_To_extensions_Deployment(dp, &out, nil)
+		return &out, nil
+	}, oldGeneration+1, time.Second, 10*time.Minute)
```
derekparker: Yeah, I can change the timeout to something shorter. 

What do you recommend instead of `oldGeneration+1`? It was my understanding the generation of the deployment would be incremented after the update. The docs of the function says it will poll until the `observedGeneration >= desiredGeneration` and we don't want the current generation, so I figured we increment it and poll until the deployment has updated.

pkg/cluster/components/deployment.go
```diff
@@ -0,0 +1,82 @@
...
+}
```
derekparker: I thought that was encoded in the deployment generation. E.g., a deployment does not get an incremented generation until the rollout has completed. I may be wrong on that, if so I'll add in this check.

pkg/cluster/updater.go
derekparker: Fixed, correctly sorting by priority.

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			return err
+		}
+		glog.Infof("Deleted pod %s", p.Name)
+
+		// Wait for all pods to be available before moving on.
+		err = wait.Poll(time.Second, 10*time.Minute, func() (bool, error) {
+			glog.Infof("checking new pod availability for DS: %s", dsu.Name)
+
+			updatedPodCount := i + 1
```
derekparker: Yeah, I agree the count is fragile. I think instead what we want here is to ensure that the Node that _was_ running the Pod we just deleted is now running a new Pod of the correct version *AND* that all pods are running. That should be the exit condition for this polling loop, I think.

aaronlevy: I don't think it should matter which node / which specific pod - all we really care about is if we are past our threshold of allowable failures.

pkg/cluster/components/node.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (nu *NodeUpdater) Name() string {
+	return nu.node.Name
+}
+
+// Priority for Nodes should be such that they
+// get updated last, and rolled back first.
+func (nu *NodeUpdater) Priority() int {
+	return 100
```
aaronlevy: nit: use const for static values

pkg/cluster/components/node.go
```diff
@@ -0,0 +1,103 @@
...
+	if n.Annotations[node.CurrentVersionAnnotation] == v.image.String() {
+		return false, nil
+	}
+	n.Annotations[node.DesiredVersionAnnotation] = v.image.String()
+	_, err = nu.client.Core().Nodes().Update(n)
+	if err != nil {
+		return false, err
+	}
+	// Second step: wait until the node-agent has updated the Node.
+	err = wait.Poll(time.Second, 10*time.Minute, func() (bool, error) {
```
aaronlevy: nit: do we really need this long of timeout? Also can we change these to be constants (e.g. nodeUpdateTimeout, deploymentUpdateTimeout, etc.)

pkg/cluster/update_controller.go
```diff
@@ -0,0 +1,290 @@
...
+		&cache.ListWatch{
+			ListFunc: func(_ api.ListOptions) (runtime.Object, error) {
+				return uc.Nodes().List(mlo)
+			},
+			WatchFunc: func(_ api.ListOptions) (watch.Interface, error) {
+				return uc.Nodes().Watch(mlo)
+			},
+		},
+		&api.Node{},
+		30*time.Minute,
```
aaronlevy: We probably want a much lower resync period for all of these informers. Maybe every few minutes? Also can we add it as a const.

pkg/cluster/update_controller.go
```diff
@@ -0,0 +1,290 @@
...
+		GetAllManagedComponentsFn: DefaultGetAllManagedComponentsFn,
+		nodes:       cache.StoreToNodeLister{nodeStore},
+		deployments: cache.StoreToDeploymentLister{deploymentStore},
+		daemonSets:  cache.StoreToDaemonSetLister{daemonSetStore},
+		pods:        components.StoreToPodLister{podStore},
+	}, nil
+}
+
+// UpdateToVersion will update the cluster to the given version.
+func (cu *UpdateController) UpdateToVersion(v *components.Version) error {
```
aaronlevy: It looks like we are only re-calling this on configMap re-sync (or change). However, because we are exiting this process after each component update - we should re-trigger this more often.

Might make sense to have 2 retry periods on the caller. If we're in the process of an update, we should re-enter this quickly (if not right away). If no update was made, or an error -- we should have a longer retry period.

vendor/k8s.io/kubernetes/pkg/client/leaderelection/leaderelection.go
```diff
@@ -214,7 +214,7 @@ func (le *LeaderElector) acquire() {
 		le.config.EventRecorder.Eventf(&api.Endpoints{ObjectMeta: le.config.EndpointsMeta}, api.EventTypeNormal, "%v became leader", le.config.Identity)
 		glog.Infof("sucessfully acquired lease %v/%v", le.config.EndpointsMeta.Namespace, le.config.EndpointsMeta.Name)
 		close(stop)
-	}, 0, stop)
+	}, time.Millisecond, stop)
```
aaronlevy: If we still need this change, it should be as part of the coreos/kubernetes fork, then we should be vendoring that.

pkg/asset/asset.go
```diff
@@ -33,6 +33,7 @@ const (
 	AssetPathKubeDNSDeployment       = "manifests/kube-dns-deployment.yaml"
 	AssetPathKubeDNSSvc              = "manifests/kube-dns-svc.yaml"
 	AssetPathSystemNamespace         = "manifests/kube-system-ns.yaml"
+	AssetPathUpdateController        = "manifests/update-controller.yaml"
```
aaronlevy: Where does this exist? I'm not seeing the manifest in this PR

Makefile
```diff
@@ -27,7 +29,15 @@ _output/bin/%: $(GOFILES)
 	mkdir -p $(dir $@)
 	GOOS=$(word 1, $(subst /, ,$*)) go build -ldflags "$(LDFLAGS)" -o $@ github.com/kubernetes-incubator/bootkube/cmd/$(notdir $@)
 
-_output/release/bootkube.tar.gz: _output/bin/linux/bootkube _output/bin/darwin/bootkube _output/bin/linux/checkpoint
+_output/bin/%/node-agent: $(GOFILES)
```
derekparker: done

build/build-image.sh
```diff
@@ -10,7 +10,11 @@ if [ -z "${BUILD_IMAGE}" ]; then
 fi
 
 BOOTKUBE_ROOT=$(git rev-parse --show-toplevel)
-source "${BOOTKUBE_ROOT}/build/build-release.sh"
+if [[ "${BUILD_TARGET}" == "" ]]; then
+    source "${BOOTKUBE_ROOT}/build/build-release.sh"
+else
+    make _output/bin/linux/$BUILD_TARGET
+fi
```
derekparker: done

pkg/cluster/update_controller.go
```diff
@@ -0,0 +1,284 @@
...
+const (
+	// ClusterConfigMapName is the name of the config map that holds cluster configuration,
+	// including the cluster version to run.
+	ClusterConfigMapName = "cluster-config"
+	// ClusterVersionKey is the key in the cluster ConfigMap that holds the version of the cluster
+	// that should be running.
+	ClusterVersionKey = "cluster.version"
+	// clusterMangedAnnotation is the annotation used to denote a managed component within
+	// the cluster.
+	clusterManagedLabel = "update-controller-managed=true"
```
derekparker: done

pkg/cluster/update_controller.go
derekparker: Makes sense. What do you think then of getting rid of the informer callbacks, and instead just looping / checking the configmap from the local store each iteration? That way it becomes a lot easier to control the iteration period.

aaronlevy: That sounds fine (s/configMap/thirdPartyResource/)

hack/multi-node/user-data.sample
```diff
@@ -19,36 +19,58 @@ coreos:
           [Unit]
           Requires=flanneld.service
           After=flanneld.service
+    - name: kubelet-initial-config.service
+      enable: true
+      command: start
+      content: |
+        [Service]
+        ExecStart=/home/core/install-initial-kubelet-config
```
derekparker: The ordering doesn't seem to work here, gave it a shot and it complained about the EnvironmentFile missing.

hack/multi-node/user-data.sample
```diff
@@ -19,36 +19,58 @@ coreos:
...
-          --hostname-override=${COREOS_PUBLIC_IPV4} \
-          --node-labels=master=true \
+          --config=/etc/kubernetes/manifests \
           --minimum-container-ttl-duration=3m0s \
           --cluster_dns=10.3.0.10 \
-          --cluster_domain=cluster.local
+          --cluster_domain=cluster.local \
+          --exit-on-lock-contention=true \
+          --node-labels=master=true,update-controller-managed=true \
+          --hostname-override=$public_ipv4
```
derekparker: Hmm, good catch, must've accidentally changed it switching the flags around from the earlier config changes (trying to use configmaps and such). But that env var is populated from cloud config.

hack/single-node/user-data.sample
```diff
@@ -20,29 +20,41 @@ coreos:
...
           --config=/etc/kubernetes/manifests \
-          --hostname-override=${COREOS_PUBLIC_IPV4} \
-          --node-labels=master=true \
           --minimum-container-ttl-duration=3m0s \
           --cluster_dns=10.3.0.10 \
-          --cluster_domain=cluster.local
+          --cluster_domain=cluster.local \
+          --exit-on-lock-contention=true \
+          --node-labels="master=true,update-controller-managed=true" \
+          --hostname-override=$public_ipv4
```
derekparker: Same answer.

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,238 @@
...
+	name string
+	// client is an API Server client.
+	client clientset.Interface
+	// pods is backed by an informer, contains list of Pods
+	// that belong to the DaemonSet.
+	pods StoreToPodLister
+	// daemonsets is a cache of DaemonSets backed by an informer.
+	daemonsets cache.StoreToDaemonSetLister
+	// selector is the DaemonSet selector.
+	selector labels.Selector
```
derekparker: `unversioned.LabelSelector` is a different type from `labels.Selector` which is what we need. Keeps us from having to convert every time.

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,238 @@
...
+	// priority is the update priority for this DaemonSet.
+	priority int
```
derekparker: Keeps us from having to do string -> int conversion more than once.

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,239 @@
...
+	// obj is the DaemonSet Object.
+	obj *extensions.DaemonSet
```
derekparker: done

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,239 @@
...
+
+// Name returns the name of the DaemonSet this updater
+// is responsible for.
+func (dsu *DaemonSetUpdater) Name() string {
+	return dsu.name
+}
+
+// Version returns the highest version of any Pod managed
+// by this DaemonSet.
+func (dsu *DaemonSetUpdater) Version() (*Version, error) {
```
derekparker: done

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	// Create new DS.
+	for i, c := range ds.Spec.Template.Spec.Containers {
+		if c.Name == dsu.Name() {
+			glog.Infof("updating image for container: %s", c.Name)
+			ds.Spec.Template.Spec.Containers[i].Image = v.image.String()
+			break
+		}
+	}
+	ds, err = dsu.client.Extensions().DaemonSets(api.NamespaceSystem).Update(ds)
```
derekparker: done

pkg/cluster/components/daemonset.go
```diff
@@ -0,0 +1,239 @@
...
+			if err != nil {
+				return false, nil
+			}
+
+			if dsu.numberOfDesiredPods() == len(pl) && dsu.allPodsAvailable() {
+				glog.Info("Pod ready, moving on")
+				return true, nil
+			}
+
+			glog.Info("Pod not ready, will check again")
```
derekparker: done

pkg/cluster/components/node.go
derekparker: done

pkg/cluster/components/node.go
derekparker: done

hack/multi-node/user-data.sample
aaronlevy: :(

So my goal is that the user-data we need to provide is only `kubelet.service` and a `kubeconfig`. So with that in mind, what about just setting two Environment variables, and an optional (`=-`) EnvironmentFile which the node-agent can use. For example:

```
Environment=KUBELET_ACI=quay.io/coreos/hyperkube
Environment=KUBELET_VERSION=v1.3.6_coreos.0
EnvironmentFile=-/etc/kubernetes/kubelet.env
```

We can revisit this later if it's really not the behavior we want - but I'd like to start clean/simple and work from there.

hack/multi-node/user-data.sample
```diff
@@ -19,36 +19,58 @@ coreos:
...
+        WantedBy=multi-user.target
     - name: kubelet.service
       enable: true
       command: start
       content: |
+        [Unit]
+        After=kubelet-initial-config.service
+
         [Service]
         EnvironmentFile=/etc/environment
```
aaronlevy: I think this should be deleted, as we have another EnvirontmentFile below

hack/multi-node/user-data.sample
aaronlevy: Unless there is a specific reason we want the change, can we switch it back - this is specific to certain cloud-providers. The COREOS_PUBLIC_IPV4 is just coming from /etc/environment file -- which is often pre-populated but would also allow a deployer to easily populate this info.

pkg/asset/internal/templates.go
```diff
@@ -243,14 +247,15 @@ metadata:
   name: kube-proxy
   namespace: kube-system
   labels:
-    k8s_app: kube-proxy
-    version: v1.3.6_coreos.0
+    k8s-app: kube-proxy
+    update-controller-managed: "true"
+  annotations:
+    update-controller.alpha.coreos.com/priority: "100"
```
aaronlevy: Just for (easier to figure out) determinism, can we skew this from the kubelet (proxy first, then kubelet)

pkg/asset/k8s.go
```diff
@@ -30,9 +30,9 @@ func newStaticAssets() Assets {
 
 func newDynamicAssets(conf Config) Assets {
 	return Assets{
-		mustCreateAssetFromTemplate(AssetPathKubelet, internal.KubeletTemplate, conf),
 		mustCreateAssetFromTemplate(AssetPathAPIServer, internal.APIServerTemplate, conf),
 		mustCreateAssetFromTemplate(AssetPathProxy, internal.ProxyTemplate, conf),
+		mustCreateAssetFromTemplate(AssetPathKubelet, internal.KubeletTemplate, conf),
```
aaronlevy: If re-ordering isn't necessary - may as well leave this out of the changeset

pkg/cluster/components/daemonset.go
aaronlevy: ack. fine with this since we dropped also storing ds object

derekparker:
Going to close this in favor of going a different direction internally at CoreOS. This closed patch will still be around and available, however we will no longer be contributing to efforts to merge this into the standard bootkube repo.

Issue #75, opened 2016-06-27
Self-host Flannel


aaronlevy:
We should see how realistic it is to self-host the flannel components. This may just be deploying flannel-server (deployment) and flannel-client (daemonset), then determining if we can get the kubelet to coordinate around "CNI is configured".

aaronlevy:
ref: https://github.com/kubernetes/kubernetes-anywhere/issues/151

aaronlevy:
added in https://github.com/kubernetes-incubator/bootkube/pull/247

PR #113, opened 2016-08-30
[wip] V1.4 updates & self-host flannel

aaronlevy:
Fixes: https://github.com/kubernetes-incubator/bootkube/issues/75


chancez:
Don't forget DNS if you haven't already, hard to tell without checking locally since the file diff kills my chrome. Just noticed last time it was forgotten.

kenan435:
I tried provisioning a cluster (vagrant, multi-node) with the changes you committed but keep having problems when the dns pod is scheduled on the worker node. kube2sky keeps restarting... Had the same problem in my PR as well. Any clue as to why?

PR #119, opened 2016-09-01
Move asset templates to Go source intead of generating them

ericchiang:
@dghubble and I are having issues with the templates "compiled" into the go source causing a ton of merge conflicts. To prevent this, generate string literals instead of byte arrays so git can reason about diffs more easily.

This PR has no functional changes.

cc @aaronlevy @derekparker


dghubble:
+1 for improving readability in this way.

aaronlevy:
Should we just drop the templates altogether and just use these directly?

ericchiang:
> Should we just drop the templates altogether and just use these directly?

Yeah we can totally do that and might be preferable. Just a trade off by causing editor confusion (syntax highlighting, indenting, etc). With that said I'd vote for just putting the templates in the Go source code.

ericchiang:
Update. No more generation to embed files in the Go source.

aaronlevy:
lgtm. @ericchiang were you able to run this through a sanity-check deploy?

ericchiang:
No. Will run one and update this PR with the results.

ericchiang:
Tested using coreos-baremetal. Everything works.

PR #120, opened 2016-09-07
Bootkube checks for active checkpointer pod.

### LINKED ISSUE #117: Have bootkube start check for active checkpointer pod
This will reduce the window on first boot where an apiserver could fail and there is no checkpoint.

Can just add the expected checkpoint pod to status.go

kenan435:
During the provisioning of the cluster bootkube will be checking for running checkpointing pod to reduce the window on first boot where an api-server could fail and there is no checkpoint.

Fixes #117

### DIFF
--- a/image/checkpoint/checkpoint-pod.yaml
+++ b/image/checkpoint/checkpoint-pod.yaml
@@ -3,6 +3,8 @@ kind: Pod
 metadata:
   name: kube-api-checkpoint
   namespace: kube-system
+  labels:
+    k8s-app: kube-api-checkpoint
 spec:
   containers:
   - name: checkpoint
--- a/pkg/bootkube/bootkube.go
+++ b/pkg/bootkube/bootkube.go
@@ -31,6 +31,7 @@ const (
 )
 
 var requiredPods = []string{
+	"kube-api-checkpoint",
 	"kubelet",
 	"kube-apiserver",
 	"kube-scheduler",


aaronlevy:
@kenan435 would you mind rebasing. And @derekparker would you be able to do a review?

kenan435:
@derekparker done with rebase

pkg/asset/internal/templates.go
```diff
@@ -129,7 +129,7 @@ spec:
       hostNetwork: true
       containers:
       - name: checkpoint-installer
-        image: quay.io/coreos/pod-checkpointer:91d5a311eee40d579a8f0549c10eeed57979d6c4
+        image: kenan435/pod-checkpointer:v1.0.0
```
derekparker: Let's continue using the official image, the tag will need to be updated in a follow-up PR, and another official image built.

pkg/bootkube/create.go
```diff
@@ -48,7 +48,7 @@ func CreateAssets(manifestDir string, timeout time.Duration) error {
 		return fmt.Errorf("Failed to create assets: %v", err)
 	}
 
-	UserOutput("Spawning self-hosted control plane...\n")
+	UserOutput("Spawning self-hosted control plane and checkpointing components...\n")
```
derekparker: nit: I don't think we need to make the distinction here.

pkg/bootkube/status.go
```diff
@@ -35,7 +36,7 @@ func WaitUntilPodsRunning(pods []string, timeout time.Duration) error {
 		return fmt.Errorf("error while checking pod status: %v", err)
 	}
 
-	UserOutput("All self-hosted control plane components successfully started\n")
+	UserOutput("All self-hosted control plane and checkpointing components successfully started\n")
```
derekparker: nit: For this and the change below, I don't think we really need to make a distinction between component types in the output during the installation.

pkg/bootkube/create.go
kenan435: I was thinking we did since the typical k8s control plane does not include checkpointing. Reverted nonetheless.

kenan435:
@derekparker changes have been applied.

pkg/bootkube/create.go
derekparker: Yeah, I definitely get that, and I appreciate it. My only real concern is that it's very possible we have bootkube install other services as well that are outside of the control plane, so once that starts happening we either continue to not distinguish, or we come up with a more generic way of labeling and displaying these components.

PR #121, opened 2016-09-08
[WIP] Kubelet tls bootstrap

derekparker:


derekparker:
Notes:

For some reason, with the changes in the `wip` patch, the API server cannot run.

Two behaviors are exhibited:

1) Self-hosted kubelet tries to run apiserver pod, but it fails on NodeLabel predicate because it does not have `--node-labels=master=true`.
2) (After updating the kubelet manifest to have the `--node-labels=master=true` flag [temp hack to get past issue 1]): When using the kubelet TLS bootstrap functionality, the downward API seems to break for host network pods when trying to get `status.podIP` (not sure the correlation there). This causes the apiserver to not start because it cannot get a valid `--advertise-address`.


/cc @yifan-gu @aaronlevy 

@yifan-gu if you wanna meet up at some point to talk through some of these issues it might help since you worked on the kubelet TLS bootstrap and may have some insight.

Putting up this WIP PR so I can switch gears for a bit.

derekparker:
FWIW These issues do not manifest without my `wip` commit, so something about enabling the kubelet TLS bootstrap, or some way that I've reconfigured the cluster is causing these issues.

aaronlevy:
Closing in favor of https://github.com/kubernetes-incubator/bootkube/pull/261

PR #123, opened 2016-09-09
pkg/bootkube: add simple stdout status info

pbx0:
Basic status information goes to stdout and diagnostics/logs go to
stderr.

Running `bootkube start 2> /dev/null`gives the following output:
```
Running temporary bootstrap control plane...
Sending self-hosted assets to temporary control plane...
Spawning self-hosted control plane...
       	Pod Status: kube-controller-manager    	Pending
       	Pod Status:          kube-scheduler    	Pending
       	Pod Status:                 kubelet    	Pending
       	Pod Status:          kube-apiserver    	Pending
       	Pod Status:          kube-scheduler    	Running
       	Pod Status:          kube-apiserver    	Running
       	Pod Status:                 kubelet    	Running
       	Pod Status: kube-controller-manager    	Running
All self-hosted control plane components successfully started

Bootstrap complete. Access your kubernetes cluster using:
kubectl --kubeconfig=cluster/auth/kubeconfig get nodes
```

First 3 lines of output happen pretty fast. I think its probably worth deleting the line about sending assets but left it in for first review.

### DIFF
--- a/pkg/bootkube/bootkube.go
+++ b/pkg/bootkube/bootkube.go
@@ -1,6 +1,7 @@
 package bootkube
 
 import (
+	"fmt"
 	"net/url"
 	"path/filepath"
 	"time"
@@ -94,6 +95,8 @@ func NewBootkube(config Config) (*bootkube, error) {
 }
 
 func (b *bootkube) Run() error {
+	UserOutput("Running temporary bootstrap control plane...\n")
+
 	errch := make(chan error)
 	go func() { errch <- apiapp.Run(b.apiServer) }()
 	go func() { errch <- cmapp.Run(b.controller) }()
@@ -107,4 +110,12 @@ func (b *bootkube) Run() error {
 
 	// If any of the bootkube services exit, it means it is unrecoverable and we should exit.
 	return <-errch
+}
+
+// All bootkube printing to stdout should go through this fmt.Printf wrapper.
+// The stdout of bootkube should convey information useful to a human sitting
+// at a terminal watching their cluster bootstrap itself. Otherwise the message
+// should go to stderr.
+func UserOutput(format string, a ...interface{}) {
+	fmt.Printf(format, a...)
 }
--- a/pkg/bootkube/create.go
+++ b/pkg/bootkube/create.go
@@ -2,6 +2,7 @@ package bootkube
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -18,6 +19,8 @@ import (
 )
 
 func CreateAssets(manifestDir string, timeout time.Duration) error {
+	UserOutput("Sending self-hosted assets to temporary control plane...\n")
+
 	upFn := func() (bool, error) {
 		if err := apiTest(); err != nil {
 			glog.Warningf("Unable to determine api-server version: %v", err)
@@ -44,6 +47,9 @@ func CreateAssets(manifestDir string, timeout time.Duration) error {
 	if err := wait.Poll(5*time.Second, timeout, createFn); err != nil {
 		return fmt.Errorf("Failed to create assets: %v", err)
 	}
+
+	UserOutput("Spawning self-hosted control plane...\n")
+
 	return nil
 }
 
@@ -111,6 +117,7 @@ func createAssets(manifestDir string) error {
 			f.PrintObjectSpecificMessage(info.Object, util.GlogWriter{})
 		}
 		cmdutil.PrintSuccess(mapper, shortOutput, util.GlogWriter{}, info.Mapping.Resource, info.Name, "created")
+		UserOutput("\tcreated %23s %s\n", info.Name, strings.TrimSuffix(info.Mapping.Resource, "s"))
 		return nil
 	})
 	if err != nil {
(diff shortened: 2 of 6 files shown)


pkg/bootkube/bootkube.go
```diff
@@ -94,6 +95,8 @@ func NewBootkube(config Config) (*bootkube, error) {
 }
 
 func (b *bootkube) Run() error {
+	fmt.Println("Running temporary bootstrap control plane...")
```
aaronlevy: nit: Can we have a wrapper func like `stdout()` to be explicit about our intention for these vs loglines?

pkg/bootkube/create.go
```diff
@@ -44,6 +46,9 @@ func CreateAssets(manifestDir string, timeout time.Duration) error {
 	if err := wait.Poll(5*time.Second, timeout, createFn); err != nil {
 		return fmt.Errorf("Failed to create assets: %v", err)
 	}
+
+	fmt.Println("Spawning self-hosted control plane...")
```
aaronlevy: We should move this to wherever this is occurring (not necessarily at the end of this function)

pbx0: It seems like the self-hosted control plane is in the process of being launched as soon as the assets are submitted to the bootstrap control plane which is when this function successfully returns.

The other place to put it would be inside the beginning of `WaitUntilPodsRunning` but then it will actually print earlier since these functions run concurrently. So the end of this function seemed the most accurate. Also putting it in `WaitUntilPodsRunning` will make it racey against the message `Sending self-hosted assets to temporary control plane...`

Its technically racey either way but in practice the way things are now, the messages should always be in the same order.

pkg/bootkube/bootkube.go
pbx0: Sounds good, I was trying to think of a way to make these sprinkled print functions clear that they are part of a specific output experience, but I didn't see an obvious way to group it all. 

A simple function wrapper to rename these sounds fine.

pkg/bootkube/status.go
```diff
@@ -34,13 +34,16 @@ func WaitUntilPodsRunning(pods []string, timeout time.Duration) error {
...
 }
 
 type statusController struct {
-	client    clientset.Interface
-	podStore  cache.Store
-	watchPods []string
+	client        clientset.Interface
+	podStore      cache.Store
+	watchPods     []string
+	lastPodPhases map[string]v1.PodPhase
```
aaronlevy: Why does this need to be part of the statusController? Seems like it's just used locally in a single function.

pbx0: It needs to persist between the polling of the `func (s *statusController) AllRunning() (bool, error)` function. 

I'm not sure if it belongs here myself, but, I think other options get more complicated too. Open to suggestions.

pkg/bootkube/create.go
aaronlevy: Maybe just a verbiage change then? "Self-hosted cluster assets submitted"

We could also log the exact components we are creating here:
https://github.com/pbx0/bootkube/blob/c12e788f71e778a85314ea4601a4a9d42912e399/pkg/bootkube/create.go#L118

Then we also might want to expose any reason we might have exited here (if non-nil):
https://github.com/pbx0/bootkube/blob/c12e788f71e778a85314ea4601a4a9d42912e399/pkg/bootkube/bootkube.go#L112

pkg/bootkube/status.go
aaronlevy: ah right. That's fine

pkg/bootkube/create.go
pbx0: Regardless of adding messages to indicate when cluster assets are submitted and messages indicating that this is all complete, I still think this message is useful because it indicates what is happening with the indented pod status messages that follow.

Playing around with exposing asset creation messages as well.

aaronlevy: I won't block on this, but this just means in the future if ordering of operations changes, or new functionality is introduced between these steps -- this log message no longer makes sense. It feels weird to describe functionality in a place where that functionality doesn't occur (except by chance of ordering at this moment in time).

pbx0: Hmm, I see your point. We could put this message in the AllRunning() func right where we print the actual pod status, but we need a persistent bool var to guard printing it only once. So it gets a little ugly either way since messaging about asset creation vs pod status will always be racey and it seems weird to synchonize them just for the user messages. I do want to leave the message in. Have a preference either way?

pbx0:
PTAL addressed all points except about the location in code that prints the "Spawning" message
```
Running temporary bootstrap control plane...
Sending self-hosted assets to temporary control plane...
       	created          kube-apiserver secret
       	created          kube-apiserver daemonset
       	created kube-controller-manager secret
       	created kube-controller-manager deployment
       	created            kube-dns-v11 deployment
       	created                kube-dns service
       	created              kube-proxy daemonset
       	created          kube-scheduler deployment
       	created                 kubelet daemonset
Spawning self-hosted control plane...
       	Pod Status:          kube-scheduler    	Pending
       	Pod Status: kube-controller-manager    	Pending
       	Pod Status:                 kubelet    	Pending
       	Pod Status:          kube-apiserver    	Pending
       	Pod Status:                 kubelet    	Running
       	Pod Status:          kube-apiserver    	Running
       	Pod Status:          kube-scheduler    	Running
       	Pod Status: kube-controller-manager    	Running
All self-hosted control plane components successfully started
```

pkg/bootkube/create.go
aaronlevy: I'd rather not add some kind of state around a log message. A common deployment is still going to just use `bootkube start` which means any effort towards layout of messages is going to be lost in the mix anyway. I'd prefer that the verbiage is changed to something along the lines of "this step is complete" or just removed.

aaronlevy:
lgtm

PR #126, opened 2016-09-13
pkg/asset: Update kube-dns to v17.1

dghubble:
Update kube-dns deployment from 11 to v17.1.

rel: https://github.com/coreos/coreos-kubernetes/commit/64e44562fd4da459d631d28f6269c22779372cca
replaces: #107
cc: @ericchiang

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -279,14 +279,14 @@ spec:
         hostPath:
           path: /etc/kubernetes
 `)
-	DNSRcTemplate = []byte(`apiVersion: extensions/v1beta1
+	DNSDeploymentTemplate = []byte(`apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
-  name: kube-dns-v11
+  name: kube-dns-v17.1
   namespace: kube-system
   labels:
     k8s-app: kube-dns
-    version: v11
+    version: v17.1
     kubernetes.io/cluster-service: "true"
 spec:
   replicas: 1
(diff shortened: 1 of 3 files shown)


aaronlevy:
lgtm

pkg/asset/internal/templates.go
```diff
@@ -369,23 +349,21 @@ spec:
...
             cpu: 10m
-            memory: 20Mi
+            memory: 50Mi
           requests:
             cpu: 10m
-            memory: 20Mi
+            memory: 50Mi
         args:
-        - -cmd=nslookup kubernetes.default.svc.cluster.local 127.0.0.1 >/dev/null
+        - -cmd=nslookup kubernetes.default.svc.cluster.local 127.0.0.1 >/dev/null && nslookup kubernetes.default.svc.cluster.local 127.0.0.1:10053 >/dev/null
```
aaronlevy: Why the additional check? This doesn't seem to be in the upstream version

dghubble: Mirrored the update done in coreos-kubernetes https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/generic/controller-install.sh#L557

dghubble:
I'll verify this through coreos-baremetal soon.

pkg/asset/internal/templates.go
aaronlevy: huh, odd. I'll have to go back and see where that came from.

For consistency with upstream, can you update this to more closely reflect: https://github.com/kubernetes/kubernetes/blob/release-1.3/cluster/saltbase/salt/kube-dns/skydns-rc.yaml.base

(but s/replicationController/Deployment)

dghubble: Are you wanting the YAML comments as well?

aaronlevy: I don't have strong feelings - but if they add useful context may as well leave them in (also makes the copy/paste job a little easier).

aaronlevy:
lgtm

dghubble:
Found a typo while validating.

PR #129, opened 2016-09-13
refactor image build process

aaronlevy:
This simplifies the build process a bit (to make it easier to add to CI)

The checkpoint build still requires a 2 step process (build image from commit, then update the desired image in api-server manifest) - but instead uses a template to enforce that the `checkpoint-install` script only ever installs the same version as itself.

/cc @derekparker @pbx0


Makefile
```diff
@@ -22,14 +22,10 @@ check:
 install: _output/bin/$(LOCAL_OS)/bootkube
 	cp $< $(GOPATH_BIN)
 
-_output/bin/%/bootkube: LDFLAGS=-X github.com/kubernetes-incubator/bootkube/pkg/version.Version=$(shell $(CURDIR)/build/git-version.sh)
-_output/bin/%/bootkube: $(GOFILES)
+_output/bin/%: LDFLAGS=-X github.com/kubernetes-incubator/bootkube/pkg/version.Version=$(shell $(CURDIR)/build/git-version.sh)
```
colhom: Probably should quote $(CURDIR) in case the absolute path has spaces in it.

build/build-image.sh
```diff
@@ -0,0 +1,13 @@
...
+
+BUILD_IMG=${BUILD_IMG:-}
+
+if [ -z "${BUILD_IMG}" ]; then
+    echo "BUILD_IMG env var must be set"
+    exit 1
+fi
+
+BOOTKUBE_ROOT=$(git rev-parse --show-toplevel)
+source ${BOOTKUBE_ROOT}/build/build-release.sh
```
colhom: Also quote any usages of `BOOTKUBE_ROOT` in case it has spaces. Yay bash.

derekparker:
Overall lgtm.

pbx0:
My one nit would be to add some documentation to the scripts about how they work similar to the PR message. Otherwise, LGTM.

aaronlevy:
@pbx0 added some docs

pbx0:
@aaronlevy Nice, docs LGTM.

PR #130, opened 2016-09-13
hack/tests: Fix conformance version separator

dghubble:
I'll post some results if/when they pass.


dghubble:
https://gist.github.com/dghubble/9875e76dde1369fd0330db07a7f1007c

aaronlevy:
Very likely due to resource exhaustion during the density test (which we should try and ignore). Going to merge this fix to at least get the conformance tests working again.

PR #133, opened 2016-09-15
Be more verbose with stdout logs

aaronlevy:
Will output any errors on exit and output all pod statuses if any have changed.

/cc @pbx0


pbx0:
LGTM

PR #134, opened 2016-09-16
Make bootkube fail when not successfully creating assets.

### LINKED ISSUE #131: bootkube should log an error when it cannot create a manifest
It seems that bootkube will silently continue if there is, for example, a typo in one of the manifests it tries to create. We should fail on anything but "already exists"

kenan435:
Bootkube will fail when its not able to successfully create assets for any other reason that "already exists error". 

Seems that bootkube is not failing when an error, such as incorrect yaml occurs. 

Fixes #131


aaronlevy:
LGTM, thanks!

PR #138, opened 2016-09-20
Log on asset creation error - do not exit

aaronlevy:
fixes https://github.com/kubernetes-incubator/bootkube/issues/131

/cc @kenan435 RE: https://github.com/kubernetes-incubator/bootkube/issues/131#issuecomment-248047201

### DIFF
--- a/pkg/bootkube/create.go
+++ b/pkg/bootkube/create.go
@@ -31,9 +31,17 @@ func CreateAssets(manifestDir string, timeout time.Duration) error {
 	createFn := func() (bool, error) {
 		err := createAssets(manifestDir)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pbx0:
I think we should make the error message when assets fail to create more actionable. If assets aren't created the user should be tracking them down and making sure to resubmit them to the new apiserver. So maybe we say something (after surfacing the asset error) like: "Some of the assets for the self-hosted cluster failed during creation. Its important that they are tracked and resubmitted to the apiserver before proceeding" I don't think that sentence is particularly eloquent, but something along those lines.

aaronlevy:
@pbx0 added some log output - PTAL

PR #139, opened 2016-09-21
Prioritizing the creation of kube-dns service

### LINKED ISSUE #136: Another service could win race for dns service IP allocation
We need a pre-assigned service IP for the kubernetes dns service - but it's possible when creating all assets (equivalent of `kubectl create -f cluster/manifests`) that another service is randomly assigned this ip (defaults to 10.3.0.10).

One option would just to force the kube-dns service to be created first - but this seems less than ideal.

kenan435:
Prioritizing the creations of kube-dns service asset by putting at the start of the queue so that we avoid the case where another service could win race for DNS service IP allocation.

fixes #136

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -199,7 +199,7 @@ func parseAPIPodSpec(podList v1.PodList) v1.PodSpec {
 		}
 	}
 	cleanVolumes(&apiPod)
-	modifyInsecurePort(&apiPod)
+	//modifyInsecurePort(&apiPod)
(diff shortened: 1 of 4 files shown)


aaronlevy:
I'd like to hold off on working toward this as the solution, as it does add a fair amount of complexity / fragility to solve an edge-case. I'll leave this PR open though in case this becomes a bigger issue - and we need to revisit

aaronlevy:
I'm going to close this as I don't think this is the long-term approach we'd end up taking. Please let me know if it should be re-opened.

PR #141, opened 2016-09-23
Run make target

ethernetdan:
Added make target that combines building, resetting the VM, and
provisioning it with bootkube to make iterating faster.

Installs Go archives to speed up builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,9 +43,11 @@ run-%: clean-vm-% _output/bin/linux/bootkube _output/bin/$(LOCAL_OS)/bootkube
 
 clean-vm-single:
 clean-vm-%:
-	@cd hack/$*-node && rm -rf cluster
 	@echo "Cleaning VM..."
-	@(cd hack/$*-node && vagrant destroy -f && vagrant up) &
+	@(cd hack/$*-node && \
+	    vagrant destroy -f && \
+	    vagrant up \
+	    rm -rf cluster) &
 
 #TODO(aaron): Prompt because this is destructive
 conformance-%: clean all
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -25,12 +25,27 @@ install: _output/bin/$(LOCAL_OS)/bootkube
...
+run-%: clean-vm-% _output/bin/linux/bootkube _output/bin/$(LOCAL_OS)/bootkube
+	@until (cd hack/$*-node && vagrant up 2>/dev/null);\
+	do \
+		echo "Vagrant not ready."; \
+		sleep 5; \
+	done;
+	@cd hack/$*-node && ./bootkube-up
+	@echo "Bootkube ready"
+
+clean-vm-single:
```
aaronlevy: Is this still necessary with the `clean-vm-%` pattern?

Makefile
```diff
@@ -25,12 +25,27 @@ install: _output/bin/$(LOCAL_OS)/bootkube
...
+clean-vm-%:
+	@echo "Cleaning VM..."
+	@(cd hack/single-node && vagrant destroy -f && vagrant up) &
```
aaronlevy: s/single-node/$*-node/

aaronlevy:
This is great! Could you add a short blurb to the README that describes the use?

Makefile
aaronlevy: Might also want to add `rm -rf /cluster` -- otherwise you could have compiled bootkube with new asset templates, but `bootkube-up` will not overwrite them if directory already exists.

Makefile
ethernetdan: It's hacky but the named targets were taking precedence over the wildcard ones. Any idea of a better way to do this?

aaronlevy: Not sure I understand the underlying issue. Doesn't seem like any other named targets would match `clean-vm-` -- and I would assume `clean-vm-%` would catch `clean-vm-single` -- but I might be misundestanding the issue. Not a big deal as-is - just making sure I understand.

PR #143, opened 2016-09-26
V1.4.0

aaronlevy:
Main changes:
- Update vendored k8s code to v1.4.0+coreos.0
- Update manifests to use v1.4.0_coreos.0 hyperkube image


hack/single-node/user-data.sample
```diff
@@ -32,12 +32,11 @@ coreos:
         ExecStartPre=/bin/mkdir -p /srv/kubernetes/manifests
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/checkpoint-secrets
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
-          --api-servers=https://172.17.4.100:443 \
           --kubeconfig=/etc/kubernetes/kubeconfig \
```
derekparker: Needs `--require-kubeconfig`

pkg/asset/internal/templates.go
```diff
@@ -349,20 +359,25 @@ spec:
       - name: healthz
         image: gcr.io/google_containers/exechealthz-amd64:1.1
         resources:
+          # keep request = limit to keep this container in guaranteed class
           limits:
             cpu: 10m
```
pbx0: nit: this is an slightly old version of kube-dns. A recent commit during beta removed a few of the cpu limits for kube-dns. See: https://github.com/kubernetes/kubernetes/commit/2b805e3d0f4bf9d377325e0433c799c3f4a56fb2

aaronlevy: Nice catch. Updated to reflect those changes.

hack/single-node/user-data.sample
aaronlevy: Added

PR #151, opened 2016-10-12
Use spec.nodeName as hostname-override, not podIP

dghubble:
* Respect names given to nodes, don't assume IP
* Allow DNS node names to be used
* kube-apiserver uses host's default iface for advertise-address

We're using this patch for bare-metal and AWS clusters to allow DNS names to be used. Kubernetes 1.4, introduced the `spec.nodeName` downward api field. Pods can make use of the node name which may be either an IP or a DNS name.

rel: #90

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -45,17 +45,17 @@ spec:
         - kubelet
         - --pod-manifest-path=/etc/kubernetes/manifests
         - --allow-privileged
-        - --hostname-override=$(MY_POD_IP)
+        - --hostname-override=$(NODE_NAME)
         - --cluster-dns=10.3.0.10
         - --cluster-domain=cluster.local
         - --kubeconfig=/etc/kubernetes/kubeconfig
         - --require-kubeconfig
         - --lock-file=/var/run/lock/kubelet.lock
         env:
-          - name: MY_POD_IP
+          - name: NODE_NAME
             valueFrom:
               fieldRef:
-                fieldPath: status.podIP
+                fieldPath: spec.nodeName
         securityContext:
           privileged: true
         volumeMounts:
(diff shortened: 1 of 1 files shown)


khrisrichardson:
@dghubble it appears that [spec.nodeName](https://github.com/kubernetes/kubernetes/pull/27880) has landed and been [vendored in](https://github.com/kubernetes-incubator/bootkube/blob/382fbcb18fa3d3c5ae4df8540550af5abf622fd2/vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go#L473-L475). Are you still intending to merge this? It could simplify the coreos-baremetal instructions for [bootkube](https://github.com/coreos/coreos-baremetal/blame/master/Documentation/bootkube.md#L15).

dghubble:
I recall there being some problem switching to nodeName and we haven't switched. We still patch this everywhere to not set the override, in which case the default behavior is to read the hosts hostname. Lemme revisit tmrw or early next week.

dghubble:
Here's the story, we use this patch on bare-metal and AWS clusters or anywhere else we'd like to name nodes by either DNS name or by IP. Using `spec.NodeName` is a now a great way to stop requiring nodes be given IP names.

The only setup I haven't verified is the vagrant/Virtualbox scripts used by this project. I suspect it may rely on:

```
- --advertise-address=$(MY_POD_IP)
```

but perhaps the default would suffice. I'll have to go get a setup to try that out.

coreos-baremetal has been updated to v1.4.7 and this required bootkube patch has been updated accordingly.

Quentin-M:
Tested against the multi-node deployment on _master_. An nginx Pod was able to reach another nginx Pod running on another worker using: service name (DNS), service IP, service NodePort, direct pod IP.

cc @dghubble

dghubble:
I spun up the `hack/multinode` example in a Vagrant / Virtualbox setup and the apiserver detects an address to use. Nothing seems amiss. So this patch seems to allow DNS names, while not harming vagrant / local examples. It would mean we can stop carrying this patch and also coreos-baremetal can use upstream again :)

```
genericapiserver.go:606] Will report 172.17.4.101 as public IP address.
```

@aaronlevy

dghubble:
Now that this has merged, https://github.com/coreos/coreos-baremetal/pull/407 updates coreos-baremetal docs to use `kubernetes-incubator/bootkube`. @khrisrichardson

khrisrichardson:
@dghubble @aaronlevy @Quentin-M thanks much for following through on this

PR #152, opened 2016-10-12
v1.4.1 + kube-dns v20

aaronlevy:


aaronlevy:
Yeah, have that locally didn't push it up.

PR #153, opened 2016-10-12
Install go-bindata as conformance test dependency

aaronlevy:


mischief:
too bad it isn't jessie-based, you could `apt-get install go-bindata`.

PR #159, opened 2016-10-20
Add self-hosted Kubernetes upgrade guide

dghubble:
Same as https://github.com/coreos/coreos-baremetal/blob/master/Documentation/bootkube-upgrades.md, but without bare-metal or QEMU/KVM mentions.

### DIFF
--- a/Documentation/upgrading.md
+++ b/Documentation/upgrading.md
@@ -1,0 +1,157 @@
+
+# Upgrading Self-hosted Kubernetes
+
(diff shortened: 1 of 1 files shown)


aaronlevy:
Actually, we should probably start a top level `docs` dir. Mind moving this into one?

dghubble:
Sure. `docs` or `Documentation` like our other projects?

aaronlevy:
Documentation sgtm

PR #162, opened 2016-10-24
[do-not-merge] Use the containerized flag rather than nsenter

aaronlevy:
/cc @dghubble @euank 

fixes https://github.com/kubernetes-incubator/bootkube/issues/96

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -38,11 +38,6 @@ spec:
       - name: kubelet
         image: quay.io/coreos/hyperkube:v1.4.3_coreos.0
         command:
-        - /nsenter
-        - --target=1
-        - --mount
-        - --wd=.
-        - --
         - ./hyperkube
         - kubelet
         - --pod-manifest-path=/etc/kubernetes/manifests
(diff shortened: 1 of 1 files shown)


euank:
Do we have a list of mount tools we ship on the OS vs what's supported in k8s so we can know what'll be broken by this?

aaronlevy:
Most of the common mount tools are included in the hyperkube image -- so should continue to work. I'm sure there are some areas that this breaks -- but the nsenter option also broke in other areas too. Part of the problem is we don't have good test coverage of this - so it's all a bit speculative.

I do know this is going to be a necessary change to support self-hosted flannel, which is going to be pretty high priority.

Figured I'd break this out -- but I'm also leaning heavily towards disabling the self-hosted kubelet by default until we can get better testing in place.

euank:
What I meant is that using `--containerized` means it won't even look at the hyperkube image for those mount utilities I thought (rather unconditionally breaking into the host and looking at utilities there).

aaronlevy:
You're right, it will use host path. I don't have a list, but could find out. We can just leave this PR as a placeholder for now.

aaronlevy:
This same change was implemented in https://github.com/kubernetes-incubator/bootkube/pull/247 - as it was necessary for the cni plugin binary paths to be correct with the self-hosted kubelet (using the hyperkube container fs).

PR #163, opened 2016-10-25
Remove version labels from manifests

derekparker:


aaronlevy:
Don't remove the version labels from the dns addon - that's part of the upstream manifest.

derekparker:
@aaronlevy updated.

PR #167, opened 2016-10-25
tests: add GCE based conformance test scripts

aaronlevy:


aaronlevy:
The ginkgo skip list still needs some work (randomly running some GCE only tests -- but might be upstream issue), but this should give us a good jenkins/CI starting point.

aaronlevy:
Just removed the skip list altogether. Test takes 45 min - but as of v1.4.5 all tests are passing.

PR #169, opened 2016-10-28
Deploy checkpointer independently from api-server

### LINKED ISSUE #161: checkpointer should be deployed independently from api-server
This can make upgrades of both checkpointer and api-server at the same time somewhat fragile. We should deploy this as a separate component so it can be upgraded independently.

kenan435:
Checkpointer deployed independently from api-server. Deploying it as a separate component so it can be upgraded independently.

fixes #161

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -127,13 +127,6 @@ spec:
         master: "true"
       hostNetwork: true
       containers:
-      - name: checkpoint-installer
-        image: quay.io/coreos/pod-checkpointer:f226b70d3a863a5dbcc5846ccd818296c30e703f
-        command:
-        - /checkpoint-installer.sh
-        volumeMounts:
-        - mountPath: /etc/kubernetes/manifests
-          name: etc-k8s-manifests
       - name: kube-apiserver
         image: quay.io/coreos/hyperkube:v1.4.3_coreos.0
         command:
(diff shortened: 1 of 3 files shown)


pkg/asset/internal/templates.go
```diff
@@ -168,13 +161,41 @@ spec:
...
           secretName: kube-apiserver
 `)
+	CheckpointerTemplate = []byte(`apiVersion: "extensions/v1beta1"
+kind: DaemonSet
+metadata:
+  name: kube-api-checkpointer
+  namespace: kube-system
+  labels:
+    k8s-app: kube-api-checkpointer
+    version: v1.0.0
```
aaronlevy: Can you drop this version label

pkg/asset/internal/templates.go
```diff
@@ -168,13 +161,41 @@ spec:
...
+spec:
+  template:
+    metadata:
+      labels:
+        k8s-app: kube-api-checkpointer
+        version: v1.0.0
```
aaronlevy: Drop version label

pkg/asset/internal/templates.go
```diff
@@ -168,13 +161,41 @@ spec:
...
-        hostPath:
-          path: /etc/kubernetes/manifests
       - name: secrets
         secret:
           secretName: kube-apiserver
 `)
+	CheckpointerTemplate = []byte(`apiVersion: "extensions/v1beta1"
+kind: DaemonSet
+metadata:
+  name: kube-api-checkpointer
```
aaronlevy: We may want to name this `checkpoint-installer` to match what it is actually doing. /cc @derekparker

aaronlevy:
Couple minor comments. Can you also be sure to sign the CLA so that check passes. And also will wait on a review from @derekparker

kenan435:
Code review findings fixed. Look like the cla check passed?

PR #176, opened 2016-11-05
etcd: add self hosted etcd mode

hongchaodeng:
- add a feature flag that will enable self hosted etcd
- on self hosted etcd mode, create a etcd static pod have bootkube wait
until it's up

### DIFF
--- a/cmd/bootkube/start.go
+++ b/cmd/bootkube/start.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/kubernetes-incubator/bootkube/pkg/bootkube"
 	"github.com/kubernetes-incubator/bootkube/pkg/util"
+	"github.com/kubernetes-incubator/bootkube/pkg/util/etcdutil"
 )
 
 var (
@@ -23,8 +24,9 @@ var (
 	}
 
 	startOpts struct {
-		assetDir   string
-		etcdServer string
+		assetDir       string
+		etcdServer     string
+		selfHostedEtcd bool
 	}
 )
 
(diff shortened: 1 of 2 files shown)


hongchaodeng:
@xiang90 @aaronlevy @philips 

If flag enabled, bootkube will create an etcd container using host network. Other than that, everything runs normal and bootkube will step to self hosted state at last.

This is also an initial step to do rest work. I would like to keep each PR small and self-contained.

pkg/util/etcdutil/start_etcd.go
```diff
@@ -0,0 +1,74 @@
...
+      --initial-cluster-state new
+      --data-dir=/var/etcd/data
+    env:
+      - name: MY_POD_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.podIP
+    image: quay.io/coreos/etcd:v3.1.0-alpha.1
+    name: etcd
+    ports:
```
aaronlevy: nit: these port entries are not required given you are using host-network. Fine to leave them in if you want to be explicit - but they are effectively non-functional.

hongchaodeng: Will remove them

cmd/bootkube/start.go
```diff
@@ -23,15 +24,17 @@ var (
...
+		etcdServer     string
+		selfHostedEtcd bool
 	}
 )
 
 func init() {
 	cmdRoot.AddCommand(cmdStart)
 	cmdStart.Flags().StringVar(&startOpts.etcdServer, "etcd-server", "http://127.0.0.1:2379", "Single etcd node to use during bootkube bootstrap process.")
 	cmdStart.Flags().StringVar(&startOpts.assetDir, "asset-dir", "", "Path to the cluster asset directory. Expected layout genereted by the `bootkube render` command.")
+	cmdStart.Flags().BoolVar(&startOpts.selfHostedEtcd, "self-hosted-etcd", false, "Self hosted etcd mode. Includes starting the initial etcd member by bootkube.")
```
aaronlevy: Can we change this to `--experimental-self-hosted-etcd` or at least add "experimental" to the help text.

cmd/bootkube/start.go
```diff
@@ -40,6 +43,12 @@ func runCmdStart(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Invalid etcd etcdServer %q: %v", startOpts.etcdServer, err)
 	}
 
+	if startOpts.selfHostedEtcd {
```
aaronlevy: I'm fine leaving this as-is for now to keep this self-contained/simple -- but this should likely move into `bootkube.Run()` eventually.

cmd/bootkube/start.go
```diff
@@ -40,6 +43,12 @@ func runCmdStart(cmd *cobra.Command, args []string) error {
...
+		if err := etcdutil.StartEtcd(); err != nil {
```
aaronlevy: Looks like our logging initialization comes after this (but glog is being used). Should either move this down, or move the logging init to be earlier.

pkg/util/etcdutil/start_etcd.go
```diff
@@ -0,0 +1,65 @@
...
+      --initial-cluster-state new
+      --data-dir=/var/etcd/data
+    env:
+      - name: MY_POD_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.podIP
+    image: quay.io/coreos/etcd:v3.1.0-alpha.1
+    name: etcd
+  dnsPolicy: ClusterFirst
```
aaronlevy: Will this etcd node need functional DNS (e.g. access any external resource on internet), because there will be no cluster dns until much later in bootstrap process (once kube-dns is running)

aaronlevy: If this isn't needed, I might just suggest dropping it.

cmd/bootkube/start.go
aaronlevy: actually, just looked at init code and this probably shouldn't matter much.

hongchaodeng:
@aaronlevy 
All fixed

PR #178, opened 2016-11-07
Stripping of nonessential info prior to checkpointing

### LINKED ISSUE #164: checkpointer should strip service accounts from pod spec
Mirror pods for the api-server are not being created due to the service account field still existing:

```
Failed creating a mirror pod for "temp-apiserver-foo_kube-system(x)": pods "temp-apiserver-ifoo" is forbidden: a mirror pod may not reference service accounts
```

We should strip the service account information before checkpointing.

Additionally there might be other fields that are not part of the base spec that we could safely strip (and all status fields should be as well.

/cc @derekparker @squat

kenan435:
Strips pod object of nonessential info. Its in [WIP] since these are few of the things that I could come up with to strip. If there are more let me know and I will add it to the function.

fixes #164

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -90,8 +90,14 @@ func run() {
 				}
 			}
 		}
-		time.Sleep(60 * time.Second)
+		time.Sleep(1 * time.Second)
 	}
+}
+
+func stripNonessentialInfo(p *v1.Pod) {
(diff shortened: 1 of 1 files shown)


aaronlevy:
LGTM. @derekparker do you have any objection to changing the delay from 60s to 1s?

derekparker:
@aaronlevy no, that should be fine.

PR #182, opened 2016-11-15
asset: add cloud provider as config option

aaronlevy:
This exposes a config option for setting the cloud-provider in the asset templates. The option is not plumbed through the cli options.

It is not part of the cli options yet because this could require a flag to be consistently used between `render` and `start`. I'd like to avoid this because a user could easily `render` with the flag, but then run `start` with out it -- which could end up in broken / difficult to debug situations. (this is also highly dependent on the cloud-provider implementation -- and whether we need this option to be part of the "internal/bootstrap" control-plane -- which I'm also leaving to a follow up PR if needed).

Instead I'd like both render and start to use a configuration file so rendered assets can also contain any options that might need to be passed to `start` -- but I will leave this to a follow-up issue.

/cc @dghubble @ericchiang @philips @sym3tri

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -141,6 +141,7 @@ spec:
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
         - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
         - --client-ca-file=/etc/kubernetes/secrets/ca.crt
+        - --cloud-provider={{ .CloudProvider }}
         env:
           - name: MY_POD_IP
             valueFrom:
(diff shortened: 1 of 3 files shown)


pkg/asset/internal/templates.go
```diff
@@ -141,6 +141,7 @@ spec:
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
         - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
         - --client-ca-file=/etc/kubernetes/secrets/ca.crt
+        - --cloud-provider={{ .CloudProvider }}
```
chancez: What if `.CloudProvider == ""`? Perhaps these need to be wrapped in a conditional check?

dghubble: Oh, I have nearly the same PR for our fork (mistakenly thought you weren't doing this) and I have the check

aaronlevy: Internally it defaults to an empty string, I believe this should be fine.

Issue #168, opened 2016-10-27
Possibly use static pods for temporary control-plane rather than compiled-in components


aaronlevy:
We currently compile the control-plane components (api-server, scheduler, controller-manager) as part of the bootkube binary. This works well enough as a means of running the temporary control plane, but it also adds some complications.

The original reason for doing this is that we wanted bootkube to be a "network only" tool that we might be able to use remotely via ssh-tunneling / reverse proxies. A hypothetical situation would be running bootkube locally (or on a "seed" box) then via an ssh reverse proxy, have a remote kubelet bootstrap a self-hosted cluster by connecting to this external control-plane on local interfaces (requiring bootkube to have no direct access to remote volumes).

Over time, it seemed like the above wasn't strictly necessary - and it was reasonable to expect an operator to ssh into a box, run bootkube once, then exit. This also had the benefit of being a relatively simple workflow.

However, compiling the components into bootkube makes installing varying versions of kubernetes difficult (without re-vendoring / recompiling the tool). Then also, if we want to start self-hosting other tools (e.g. etcd) we may also need to compile these into bootkube itself.

Alternatively, we could have bootkube manage the temporary control-plane using static pods. For example:
1. Bootkube starts, and drops api-server, scheduler, controller-manager, and etcd pod manifests into: `/etc/kubernetes/manifests`.
2. This static (and temporary) control-plane starts, then bootkube injects daemonsets, deployments, secrets, & configMaps which make up the self-hosted control-plane.
3. Bootkube waits for the self-hosted control-plane to start, and when all required pods are running, bootkube deletes the static manifests in /etc/kubernetes/manifests and exists (leaving us with a self-hosted cluster)

---

I'm not sure this is the route we want to take -- but it's relatively simple, and would work today without any upstream changes (e.g. adding a kubelet-pod-api might be an alternative sometime in the future).

This also might be more palatable to a tool like `kubeadm` to launch a self-hosted cluster by shuffling around static assets, rather than running a compiled-in control-plane.

ericchiang:
I'm for this. It helps users see and understand what bootkube is doing as well, which is an extremely important part of using the tool.

dhawal55:
I like this too. Are you going to pick this up?

aaronlevy:
I was hoping to sometime in the next few weeks -- but haven't currently picked this up. Is this something you were interested in working on?

[...]

PR #186, opened 2016-11-17
pkg/asset: Add CloudProvider field to Config

dghubble:
Similar to #182, it was determined that changes to `bootkube start` and coordination btw render and start weren't needed. This change does not expose CloudProvider as a flag, unless that's desired @aaronlevy?

### DIFF
--- a/pkg/asset/asset.go
+++ b/pkg/asset/asset.go
@@ -45,13 +45,14 @@ type Config struct {
 	CAPrivKey       *rsa.PrivateKey
 	AltNames        *tlsutil.AltNames
 	SelfHostKubelet bool
+	CloudProvider   string
 }
 
 // NewDefaultAssets returns a list of default assets, optionally
 // configured via a user provided AssetConfig. Default assets include
 // TLS assets (certs, keys and secrets), and k8s component manifests.
 func NewDefaultAssets(conf Config) (Assets, error) {
-	as := newStaticAssets(conf.SelfHostKubelet)
+	as := newStaticAssets()
 	as = append(as, newDynamicAssets(conf)...)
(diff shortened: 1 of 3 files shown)


aaronlevy:
Couple changes to request:

- Can you drop the `if` statements around the cloudprovider options. They're not really necessary as the default cloud provider is an empty string anyway.

- Remove the option from the self-hosted kubelet. My reasoning for this is that as-documented this would put a user in a state where the on-host kubelet (which we document, but do not render) could behave differently than the self-hosted kubelet. In the default state the kubelet should "auto-detect" in both cases -- so it should work without changes.

And just a side note: before we can have this as a reasonable flag/option -- we need some way of coordinating options from `render` and options from `start`. Otherwise you could render+commit assets, then depending on how you run `start`, those same assets may or may not work. 

Two ways I could see this handled:

- `render` saves any options that might affect `start` command in a configuration file in the asset-dir, which is then consumed by the `start` command.

- We move to using static manifests for the temporary control-plane, such that render also creates the same assets which will be used by bootkube during the initial bring-up (see: https://github.com/kubernetes-incubator/bootkube/issues/168) -- in general I think this is a preferable approach.

aaronlevy:
Or, I guess we could assume that the temporary control-plane does not need to reflect the same cloud-provider options (initially). I'm not entirely sure if this is a safe assumption, as the temporary control-plane would be acting with no cloud-provider, until the self-hosted control-plane takes over. This may be fine -- but not something I'm deeply familiar with.

aaronlevy:
closed in https://github.com/kubernetes-incubator/bootkube/pull/105

PR #189, opened 2016-11-18
use etcd server from config

hongchaodeng:
etcdServer is passed in in "bootkube start".
We should use it.


hongchaodeng:
@aaronlevy 
I have a couple of small fixes I want to make and finally support the new self hosted etcd change in etcd operator.

PR #190, opened 2016-11-18
create 1.4 clientset from config and use its RESTClient

hongchaodeng:
I'm getting error like "need to use GroupVersion to create RESTClient".

I don't really need to create rest client indeed. I can get that from clientset CoreClient.


hongchaodeng:
@aaronlevy 
I think the error might be due to dep update.

PR #191, opened 2016-11-18
use self hosted spec and change static etcd pod

hongchaodeng:
Self hosted spec is changed from etcd operator side.
Because we are now using host network, we can't conflict with self hosted etcd pods which use "2379" hard-coded right now, and we change the static etcd pod to use different ports.

## Changes in workflow:
"bootkube start":
```
$ _output/bin/linux/bootkube start --asset-dir=asset --experimental-self-hosted-etcd=true --etcd-server http://127.0.0.1:12379
```
We need to pass in etcd-server.


hongchaodeng:
@aaronlevy

xiang90:
lgtm. defer to @aaronlevy

Issue #194, opened 2016-11-22
service IP to remote host-networking pod does not work in vagrant


aaronlevy:
In the vagrant hack examples, you cannot reach a remote pod via service IP when that remote pod is using host-networking.

Repro

pod.yaml
```
apiVersion: v1
kind: Pod
metadata:
  name: n1
  labels:
    name: n1
spec:
  hostNetwork: true
  containers:
  - name: nginx
    image: nginx
```

```
kubectl create -f pod.yaml
kubectl expose pod n1 --port=80
kubectl get service n1
```

Then from a different machine than n1 is running on, try and curl the service IP from the host.

PR #195, opened 2016-11-22
Set proxy hostname and cluster cidr

aaronlevy:
fixes https://github.com/kubernetes-incubator/bootkube/issues/194

/cc @hongchaodeng @xiang90

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -277,6 +277,13 @@ spec:
         - proxy
         - --kubeconfig=/etc/kubernetes/kubeconfig
         - --proxy-mode=iptables
+        - --hostname-override=$(POD_IP)
+        - --cluster-cidr=10.2.0.0/16
+        env:
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
         securityContext:
           privileged: true
         volumeMounts:


xiang90:
Do you know why do we need to set this? Why does non host network service IP work?

hongchaodeng:
@aaronlevy 
I'm very new to kube proxy.
Mind explain how the two flags, "hostname-override" and "cluster-cidr", a bit more and how they fix the problem?

aaronlevy:
When the `--cluster-cidr` flag is used, the proxy will setup iptables MASQUERADE rules for all traffic which is *not* part of the pod-network. What I believe is happening with the vagrant/virtualbox installations is that the default interface (eth0) is not actually routable between nodes - so traffic leaves the node but is incorrectly routed on the return path.

Specifically virtualbox will assign all node's eth0 to 10.0.2.15.
Packet will leave one box with a source of 10.0.2.15, then when it gets to the destination, the return address is set to 10.0.2.15 -- which is also the remote node's eth0 -- so it essentially routes to itself on the remote node.

With the MASQUERADE rules, it will make sure the return route will use the interface the traffic is actually leaving from (eth1) (not the default route, eth0). Or that is my uninformed networking understanding.

The `--hostname-override` likely doesn't affect this actual issue (I tested it in isolation, and it did not resolve the issue, but I did not test only `--cluster-cidr`. However, I was seeing error loglines in the proxy which were stating that because it cannot resolve the hostname, it is assuming a nodename of `127.0.0.1`. I'm not sure how this would be used, execpt maybe to try and query its own node object -- in which case localhost is always incorrect. So this flag is actually setting this to an accurate address (the pod-ip of hostNetwork pods will be the hostIP as populated by the kubelet).

xiang90:
LGTM

hongchaodeng:
LGTM. Thanks!

PR #197, opened 2016-11-23
hack: multi-node use the ssh-config

philips:
As much as I love awk this is totally unnecessary. Just use `ssh -F`
instead.

### DIFF
--- a/hack/multi-node/bootkube-up
+++ b/hack/multi-node/bootkube-up
@@ -19,16 +19,14 @@ fi
 
 # Start the VM
 vagrant up
-
-ssh_ident="$(vagrant ssh-config c1 | awk '/IdentityFile/ {print $2}' | tr -d '"')"
-ssh_port="$(vagrant ssh-config c1 | awk '/Port [0-9]+/ {print $2}')"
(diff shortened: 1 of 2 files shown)


hack/multi-node/bootkube-up
```diff
@@ -9,7 +9,7 @@ fi
 
 # Render assets
 if [ ! -d "cluster" ]; then
-  ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.101:443 --etcd-servers=http://172.17.4.51:2379
+  ../../_output/bin/${local_os}/bootkube render --self-host-kubelet=false --asset-dir=cluster --api-servers=https://172.17.4.101:443 --etcd-servers=http://172.17.4.51:2379
```
aaronlevy: self-host-kubelet=false is the default. Do you want it here to be more easily changeable?

aaronlevy:
Mind also making the same change to `hack/single-node/bootkube-up`?

hack/multi-node/bootkube-up
philips: copy pasta from aaronlevy branch, fixed.

PR #199, opened 2016-11-23
checkpoint: make run func generic

xiang90:
Make run func generic. Preparing for adding etcd checkpoint support.

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -24,17 +24,15 @@ const (
 	kubeletAPIPodsURL = "http://127.0.0.1:10255/pods"
 	ignorePath        = "/srv/kubernetes/manifests"
 	activePath        = "/etc/kubernetes/manifests"
-	manifestFilename  = "apiserver.json"
 	kubeconfigPath    = "/etc/kubernetes/kubeconfig"
 	secretsPath       = "/etc/kubernetes/checkpoint-secrets"
+
+	tempAPIServer = "temp-apiserver"
+	kubeAPIServer = "kube-apiserver"
 )
 
 var (
-	tempAPIServer      = []byte("temp-apiserver")
-	kubeAPIServer      = []byte("kube-apiserver")
-	activeManifest     = filepath.Join(activePath, manifestFilename)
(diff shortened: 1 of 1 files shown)


xiang90:
If this approach looks good to you, I will continue to clean up the manifest file part.

cmd/checkpoint/main.go
```diff
@@ -30,8 +30,9 @@ const (
 )
 
 var (
-	tempAPIServer      = []byte("temp-apiserver")
-	kubeAPIServer      = []byte("kube-apiserver")
+	tempAPIServer = "temp-apiserver"
```
derekparker: Since these are just strings now instead of byte slices, make then `const`.

cmd/checkpoint/main.go
```diff
@@ -115,36 +116,35 @@ func getPodsFromKubeletAPI() []byte {
...
+		tempPodSeen = tempPodSeen || isPod(p, tn, ns)
+		if actualPodSeen && tempPodSeen {
 			return true
 		}
 	}
 	return false
 }
 
-func kubeSystemAPIServerRunning(pods v1.PodList, client clientset.Interface) bool {
+func actualPodRunning(pods v1.PodList, client clientset.Interface, n, ns string) bool {
```
derekparker: nit: Maybe would be better named generically `isPodRunning` or just `podRunning`?

xiang90: will change to is podRunning.

xiang90:
@derekparker I will continue to make the manifest part more generic.

xiang90:
@derekparker @aaronlevy 

All fixed. And I changed manifest part to be generic too. PTAL. I do not have setup to test checkpointer yet. It would be great, if you can help me to do a quick test. Thank you in advance.

aaronlevy:
@pbx0 this would be a good candidate for your destructive tests. Testing this will be a little wonky though - we would likely need to build a checkpoint image from this commit, then use that image in the checkpoint manifest in this PR.

pbx0:
@aaronlevy: LGTM I tested it against the basic destructive test here: https://github.com/kubernetes-incubator/bootkube/pull/201

PR #200, opened 2016-11-25
Enable bootkube setup behind proxy

takac:
Add http proxy settings to user-data.sample which is then expanded or
deleted when running bootkube-up. This  allows setup behind corporate proxies.


takac:
I signed it.

Issue #183, opened 2016-11-15
checkpointer needs to GC pods (and itself) when are no longer scheduled to node


aaronlevy:
An example:

If you were to label another node as `master=true` then an api-server & checkpointer will be scheduled to that node. If later this label changes, the checkpointer should know to remove any checkpointed pods (rather than keeping them running). It should also likely GC itself.

One option:

checkpointer tries to contact the api-server and if it is able, and the pod is no longer supposed to be scheduled to the node, delete active/inactive/secrets/configmaps for that pod.

Once all checkpointed pods/assets are cleaned up, checkpointer should also delete itself if the "installer" is no longer scheduled to the node.

aaronlevy:
checkpointer will now GC checkpointed pods which are no longer scheduled to the node: https://github.com/kubernetes-incubator/bootkube/pull/238

Will open a new issue to track GCing itself

PR #204, opened 2016-11-29
checkpoint: support etcd checkpoint

xiang90:
/cc @aaronlevy @derekparker 

Any idea about how to test this?

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"flag"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -29,6 +30,9 @@ const (
 
 	tempAPIServer = "temp-apiserver"
 	kubeAPIServer = "kube-apiserver"
+
+	tempEtcd   = "temp-self-hosted-etcd"
+	etcdPrefix = "self-hosted-etcd"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


aaronlevy:
One option for testing:

1. Build a checkpoint image from this PR (there is a private jenkins job you can use to do this)
2. Launch a cluster using https://github.com/kubernetes-incubator/bootkube/tree/multi-node-self-host-etcd -- but change the image repo:tag to reference the image built in step 1 (https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L184)
3. restart the node running etcd

In terms of general functionality -- I'd like to move to using pod annotations rather than searching for fixed pod names. Also, just to note, we also will need to implement checkpointed pod GC for this to work well with etcd.

https://github.com/kubernetes-incubator/bootkube/issues/205
https://github.com/kubernetes-incubator/bootkube/issues/183

xiang90:
> I'd like to move to using pod annotations rather than searching for fixed pod names. 

Do you want me to change this in this pr or a separate one? I feel adding etcd checkpointer and using annotation is orthogonal. But I can make the matching to use annotation first.

cmd/checkpoint/main.go
```diff
@@ -29,6 +30,9 @@ const (
 
 	tempAPIServer = "temp-apiserver"
 	kubeAPIServer = "kube-apiserver"
+
+	tempEtcd   = "temp-self-hosted-etcd"
+	etcdPrefix = "self-hosted-etcd"
```
xiang90: Sure.

cmd/checkpoint/main.go
```diff
@@ -46,12 +50,32 @@ var tempAPIServerManifest = v1.Pod{
...
+}
+
 var tempPodSpecMap = map[string]v1.Pod{
 	tempAPIServer: tempAPIServerManifest,
+	tempEtcd:      tempEtcdManifest,
 }
 
 func main() {
 	glog.Info("begin pods checkpointing...")
+	etcd := flag.Bool("experimental-self-hosted-etcd", false, "Checkpoint self hosted etcd.")
```
xiang90: If that is the case, I think it makes sense to finish the annotation thing first? Then the checkpointer knows what to pick up to checkpoint.

aaronlevy: That would be great if you wanted to make that change.

aaronlevy: Or we could get by with it as-is, just without the flag (checkpoint etcd pod by name if it exists)

xiang90: Let's me try it first tomorrow.

aaronlevy:
Closing, as generic pod checkpointing was implemented in https://github.com/kubernetes-incubator/bootkube/pull/238

PR #209, opened 2016-12-02
hack: openstack: initial commit

philips:
This adds OpenStack instructions tested against vexxhost. Overall
everything seems to work except the kubernetes cloud provider integration for LBaaS but I think
this is a vexxhost issue.

### DIFF
--- a/hack/quickstart/openstack/config.tf
+++ b/hack/quickstart/openstack/config.tf
@@ -1,0 +1,24 @@
+variable "key_pair" {
+  type    = "string"
+  default = "philips"
+}
+
+variable "flavor_id" {
+  type    = "string"
+  default = "bbcb7eb5-5c8d-498f-9d7e-307c575d3566"
+}
+
+variable "image_id" {
+  type    = "string"
+  default = "3a0c0bac-fa91-4c96-bfcb-ee215ba1cd4d"
+}
+
+variable "controller_count" {
+  type = "string"
+  default = "1"
+}
+
+variable "worker_count" {
+  type = "string"
+  default = "3"
+}
--- a/hack/quickstart/openstack/openstack.tf
+++ b/hack/quickstart/openstack/openstack.tf
@@ -1,0 +1,102 @@
+resource "openstack_compute_secgroup_v2" "kubernetes_control_public" {
+  name = "80AC0F3A-ECF8-424C-88FA-D8571F0C0EC9"
+  description = "external controller security group"
+  rule {
+    from_port = 22
+    to_port = 22
+    ip_protocol = "tcp"
+    cidr = "0.0.0.0/0"
+  }
+  rule {
+    from_port = 443
+    to_port = 443
+    ip_protocol = "tcp"
+    cidr = "0.0.0.0/0"
+  }
+}
+
+resource "openstack_compute_secgroup_v2" "kubernetes_control_private" {
+  name = "0C28F510-0569-4D47-BC90-710EB0111158"
+  description = "internal controller security group"
+  rule {
+    from_port = 1
+    to_port = 65535
+    ip_protocol = "tcp"
+    self = true
+  }
+
+}
+
+
+resource "openstack_compute_secgroup_v2" "kubernetes_control_worker_private" {
+  name = "1396C68D-8014-45AC-979B-A7EB4E6695F3"
+  description = "internal controller worker security group"
+  rule {
+    from_port = 2379
+    to_port = 2379
+    ip_protocol = "tcp"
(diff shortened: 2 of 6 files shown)


philips:
@aaronlevy can I just merge this? It has been tested so I would rather just have it than not.

hack/quickstart/openstack/config.tf
```diff
@@ -0,0 +1,24 @@
+variable "key_pair" {
+  type    = "string"
+  default = "philips"
```
aaronlevy: Use a more generic placeholder?

hack/quickstart/openstack/config.tf
```diff
@@ -0,0 +1,24 @@
...
+}
+
+variable "flavor_id" {
+  type    = "string"
+  default = "bbcb7eb5-5c8d-498f-9d7e-307c575d3566"
```
aaronlevy: What are these values mapping to? Is this specific to a particular provider or openstack in general?

hack/quickstart/openstack/openstack.tf
```diff
@@ -0,0 +1,102 @@
+resource "openstack_compute_secgroup_v2" "kubernetes_control_public" {
+  name = "80AC0F3A-ECF8-424C-88FA-D8571F0C0EC9"
```
aaronlevy: Is this significant or just placeholder?

hack/quickstart/quickstart-openstack.md
```diff
@@ -0,0 +1,110 @@
...
+Similarly on the controller manager:
+
+```
+kubectl edit deployment -n kube-system kube-controller-manager
+```
+
+```
+      volumes:
+      - hostPath:
+          path: /etc/kubernetes
```
aaronlevy: We shouldn't be relying on host paths - this should ideally be in a configMap

aaronlevy:
A couple minor comments.

One of my concerns is that I likely will not be able to adequately keep this updated as bootkube changes fairly rapidly right now (and I personally have no openstack experience).

Couple options:
- somehow mark it as "best effort maintained"
- owners files for each implementation
- move to a "contrib" directory (and note that it is "best effort" maintained)

philips:
@aaronlevy just assign issues to me.

philips:
@aaronlevy I will fix up the nits.

hack/quickstart/openstack/config.tf
s-urbaniak: This is just the default value which can be override in terraform using `terraform -var 'key_pair=mykey' apply/plan`, I'll update the readme accordingly.

hack/quickstart/openstack/config.tf
s-urbaniak: This seems to be provider specific, openstack comes with defaults, but vexxhost has its own. I'll add a small chapter on how to list flavors.

hack/quickstart/openstack/openstack.tf
s-urbaniak: I believe this is arbitrary and optionally used for back-references, did you auto-generate the names here @philips?

s-urbaniak: The name is not even used for back-references, I'll humanize them.

hack/quickstart/quickstart-openstack.md
```diff
@@ -0,0 +1,110 @@
...
+$ terraform apply openstack/
+```
+
+
+### Bootstrap Master
+
+*Replace* `<node-ip>` with the EXTERNAL_IP from output of `openstack server list --format=json --name 'control_node_.' | jq .[].Networks`.
+
+```
+export IDENT=~/.ssh/id_rsa.pub; ./init-master.sh <node-ip>
```
s-urbaniak: Note to myself: replace and document this with the generated openstack keypair.

hack/quickstart/quickstart-openstack.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+### Add Workers
+
+*Replace* `<node-ip>` with the EXTERNAL_IP from output of `openstack server list --format=json --name 'worker_node_.' | jq .[].Networks`.
+
+Initialize each worker node by replacing `<node-ip>` with the EXTERNAL_IP from the commands above.
+
+```
+$ IDENT=~/.ssh/google_compute_engine ./init-worker.sh <node-ip> cluster/auth/kubeconfig
```
s-urbaniak: Note to myself: replace and document this with the generated openstack keypair.

s-urbaniak:
I suggest to close this and proceed the discussion in #299

PR #210, opened 2016-12-02
checkpoint: use Annotations for finding checkpoint pod

xiang90:
/cc @aaronlevy 

Use checkpoint=name instead of just a boolean value, since we need to know its name for constructing the temp manifest anyway.

Can you please point me the place where we create the api server spec? I need to add the checkpoint annotation to it.

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -27,35 +27,40 @@ const (
 	kubeconfigPath    = "/etc/kubernetes/kubeconfig"
 	secretsPath       = "/etc/kubernetes/checkpoint-secrets"
 
-	tempAPIServer = "temp-apiserver"
-	kubeAPIServer = "kube-apiserver"
+	checkpointAPIServer = "checkpoint-apiserver"
+	kubeAPIServer       = "kube-apiserver"
+
+	checkpointAnnotation = "alpha.coreos.com/checkpoint"
 )
 
 var (
 	secureAPIAddr = fmt.Sprintf("https://%s:%s", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS"))
 )
 
-var tempAPIServerManifest = v1.Pod{
+var checkpointAPIServerManifest = v1.Pod{
 	TypeMeta: unversioned.TypeMeta{
 		APIVersion: "v1",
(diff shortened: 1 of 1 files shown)


xiang90:
Also I remove the temp in the name. I use `checkpoint` as a prefix for the checkpointed pod annotation.

aaronlevy:
@xiang90 the api-server manifest is defined in: https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go

I'm not sure that we do need the pod name. Doesn't seem like we're doing anything specific to the api-server. Also not sure that we need to even pre-define the checkpointPodSpec (likely can just copy the pod, then reset status / clean volumes / etc).

Pseudo code:

```
pods := getAllPods()
for _, p := range pods {
    if !p.Annotations[checkpointAnnotation] {
        continue
    }
    obj := api.Scheme.Copy(p)
    checkpointPod := obj.(v1.Pod)
    cleanVolumes(checkpointPod)
    stripNonessentialInfo(checkpointPod)
    ...
}
```

aaronlevy:
If we do need the name, we should be able to get it from the object itself (not as an annotation). e.g. `pod.ObjectMeta.Name`

xiang90:
@aaronlevy 

The current value in checkpoint annotation is for matching purpose. I do not want to match on name to determine if we want to checkpoint a pod.

So, I think the question is that does our checkpointer want to checkpoint ALL pods with checkpoint enabled or it should only checkpoint the pod defined here: https://github.com/kubernetes-incubator/bootkube/blob/master/cmd/checkpoint/main.go#L49.

If we want to checkpoint ALL, we need to be able to create go routines on demand and generate the checkpoint pod manifest on demand, which is a larger change.

aaronlevy:
I think we should be able to (or try) to checkpoint any pods which have the annotation (seems easiest). If we leave it structured as-is right now, we would need to launch multiple go routines, but seems like we could just change our main loop to get all pods, then filter based on annotation, then checkpoint. But I also may be missing something.

xiang90:
Discussed with @aaronlevy offline. We will try to make checkpointer generic, and be able to checkpoint all pods with checkpoint annotation when possible. This requires a larger change, I will try to finish it early next week.

aaronlevy:
Closing, as this was implemented in https://github.com/kubernetes-incubator/bootkube/pull/238

PR #212, opened 2016-12-03
Verbose glog level for bootstrap components

### LINKED ISSUE #198: Allow --v to be set
Currently there is no way to set the verbosity level of glog for the bootkube control plane components. Adding `--v` to the flag doesn't work because the flag is not being registered with cobra.

@aaronlevy

ethernetdan:
Increases the glog verbosity of the bootstrap components by using an
environment variable.

Closes #198 by making verbose logging default.

### DIFF
--- a/hack/multi-node/bootkube-up
+++ b/hack/multi-node/bootkube-up
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail
 
+GLOG_v=${GLOG_v:-1}
+
 un="$(uname)"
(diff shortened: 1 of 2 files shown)


ethernetdan:
I signed it!

aaronlevy:
Can we make this a configurable and default to the normal verbosity:

`GLOG_v=${GLOG_v:-<whatever is default>}`

ethernetdan:
@aaronlevy updated

PR #213, opened 2016-12-05
checkpoint: remove template checkpoint pod spec

xiang90:
/cc @aaronlevy @derekparker 

This is based on my offline discussion with @aaronlevy. Kubelet will add -xxx as a suffix to the checkpoint pod it starts, so we do not need to have the special name.

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -31,24 +31,14 @@ const (
 	kubeAPIServer = "kube-apiserver"
 )
 
+var podAPIServerMeta = unversioned.TypeMeta{
+	APIVersion: "v1",
+	Kind:       "Pod",
+}
+
 var (
 	secureAPIAddr = fmt.Sprintf("https://%s:%s", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS"))
 )
-
-var tempAPIServerManifest = v1.Pod{
-	TypeMeta: unversioned.TypeMeta{
-		APIVersion: "v1",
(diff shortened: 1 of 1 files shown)


xiang90:
@pbx0 How should I trigger CI to test this?

pbx0:
@xiang90 Its a little awkward, but one way to do it is to first build a checkpoint image from this branch via the internal jenkins job. Then add a commit to this branch that bumps the checkpoint image to use the newly built image in the `pkg/asset/internal/templates.go` file. See https://github.com/kubernetes-incubator/bootkube/pull/201 for an example. 

The test PR will trigger a test that performs a smoke test and another check that does a basic "destruction" test which restarts the master and runs a basic workload.

At somepoint the checkpoint code will be moved to its own repo or we will improve the PR triggers to work around this.

cmd/checkpoint/main.go
```diff
@@ -31,25 +31,15 @@ const (
 	kubeAPIServer = "kube-apiserver"
 )
 
+var podAPIServerMeta = unversioned.TypeMeta{
```
aaronlevy: s/podAPIServerMeta/podMeta/ -- not specific to apiServer, and we'll need to use it for other checkpointed pods.

aaronlevy:
One minor comment if you don't mind updating.

cmd/checkpoint/main.go
xiang90: I want to make it clear that the pod meta is from API sever (which we simulate here) not kubelet. But if you think this is confusing, I can remove the API server part.

aaronlevy: sgtm

PR #217, opened 2016-12-06
Jenkins 2

pbx0:
This time we break the checkpoint code on purpose -- the destructive test should fail only in the checkpoint-dev job


pbx0:
test checkpoint code please

PR #220, opened 2016-12-06
checkpoint: understand annotation

xiang90:
use annotation to indicate that the pod needs to be checkpointed.

/cc @aaronlevy

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/unversioned"
 	"k8s.io/kubernetes/pkg/api/v1"
@@ -28,13 +27,20 @@ const (
 	kubeconfigPath    = "/etc/kubernetes/kubeconfig"
 	secretsPath       = "/etc/kubernetes/checkpoint-secrets"
 
-	tempAPIServer = "temp-apiserver"
-	kubeAPIServer = "kube-apiserver"
+	kubeAPIServer = api.NamespaceSystem + "/" + "kube-apiserver"
+
+	checkpointAnnotation   = "checkpoint.alpha.coreos.com/checkpoint"
+	checkpointOfAnnotation = "checkpoint.alpha.coreos.com/checkpoint-of"
 )
 
 var podAPIServerMeta = unversioned.TypeMeta{
 	APIVersion: "v1",
 	Kind:       "Pod",
+}
+
+type checkpointPodPair struct {
+	parent *v1.Pod
+	child  *v1.Pod
 }
 
 var (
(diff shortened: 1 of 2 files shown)


xiang90:

cmd/checkpoint/main.go
```diff
@@ -47,49 +48,73 @@ func main() {
...
-			// kubelet will stop it.
-			if err := os.Remove(activeManifest(tempPodName)); err != nil {
-				glog.Error(err)
+
+		pwc := getPodsWithCheckpointAnnotation(podList)
+		// special case for handling API server readiness
+		if _, ok := pwc[kubeAPIServer]; ok {
+			// Make sure it's actually running. Sometimes we get that
+			// pod manifest back, but the server is not actually running.
+			if _, err := client.Discovery().ServerVersion(); err != nil {
```
aaronlevy: This could be incorrect when running multiple api-servers. The local pod we're trying to checkpoint could in fact be broken, but we happen to hit a different api-server using this client request.

This logic was already in place, so we don't need to fix it here. But maybe can you add a "TODO: use pod healthcheck to determine if apiserver pod is successfully running."

xiang90: agree. will add a todo.

cmd/checkpoint/main.go
```diff
@@ -47,49 +48,73 @@ func main() {
...
+				// The actual is running. Let's snapshot the pod,
+				// clean it up a bit, and then save it to the ignore path for
+				// later use.
+				checkpointPod := createCheckpointPod(pods[0], name, namespace)
+				convertSecretsToVolumeMounts(client, &checkpointPod)
+				writeManifest(checkpointPod, name)
+				checkpoints[checkpointManifest(name)] = struct{}{}
+				glog.Infof("finished creating checkpoint pod %v manifest at %s\n", name, checkpointManifest(name))
+
+			case len(pods) == 2:
```
xiang90: I totally agree. One way to solve this is to add annotation for the checkpoint pod. I looked through the k8s doc, it seems there is no reliable way to tell if a pod is a static one except from it does not have api server tyep meta...

cmd/checkpoint/main.go
xiang90: sure. open an issue please.

cmd/checkpoint/main.go
aaronlevy: Also RE: the api type meta -- that is just because we're talking to the kubelet api endpoint. Even static pods have that info if you get their status from the apiserver.

cmd/checkpoint/main.go
```diff
@@ -47,49 +48,73 @@ func main() {
...
+
+			if _, ok := pwc[pname]; !ok {
+				glog.Info("no actual pod running, installing checkpoint pod static manifest")
+				b, err := ioutil.ReadFile(checkpoint)
+				if err != nil {
+					glog.Error(err)
+					continue
+				}
+
+				// Think: touch activeManifest first? do not write if already exists?
```
aaronlevy: Looks like we have just been overwriting each time. We may need to test the kubelet behavior (e.g. is it just inotify, or does it check for changes). If an overwrite that contains no changes actually triggers a pod restart, we should be more careful about this. But also  as-is wouldn't be a behavior change, so we could leave that to follow up too.

cmd/checkpoint/main.go
```diff
@@ -47,49 +48,73 @@ func main() {
...
+				// kubelet will stop it.
+				if err := os.Remove(activeManifest(name)); err != nil {
 					glog.Error(err)
 				}
 			}
 		}
+
+		// start the checkpoint pod if it has not started yet
+		for checkpoint := range checkpoints {
+			pname := podNameFromManifestPath(checkpoint)
```
aaronlevy: If we were to track the "parent" as part of the annotations, we could use that to determine this as well.

cmd/checkpoint/main.go
aaronlevy: https://github.com/kubernetes-incubator/bootkube/issues/223

pkg/asset/internal/templates.go
```diff
@@ -113,6 +113,8 @@ metadata:
   namespace: kube-system
   labels:
     k8s-app: kube-apiserver
+  annotations:
+    alpha.coreos.com/checkpoint: kube-apiserver
```
aaronlevy: This annotation needs to be moved under `spec.template.metadata` -- we want the annotation on the pod (and this is on the daemonset)

xiang90:

aaronlevy:
@xiang90 one of the issues is that the annotation is in the wrong place (commented inline).

Then another issue seems to be some fields we are no longer clearing (now that we are keeping the pod metadata header) that the api-server doesn't like. 

An example of a checkpointed pod that won't start:

```
  "kind": "Pod",
  "apiVersion": "v1",
  "metadata": {
    "name": "kube-apiserver-m4nyo",
    "generateName": "kube-apiserver-",
    "namespace": "kube-system",
    "selfLink": "/api/v1/namespaces/kube-system/pods/kube-apiserver-m4nyo",
    "uid": "900cbe11-bc25-11e6-aab0-080027db44c2",
    "resourceVersion": "176",
    "creationTimestamp": "2016-12-07T02:33:39Z",
    "labels": {
      "k8s-app": "kube-apiserver"
    },
    "annotations": {
      "alpha.coreos.com/checkpoint": "kube-apiserver",
      "kubernetes.io/config.seen": "2016-12-07T02:33:39.84849503Z",
      "kubernetes.io/config.source": "api",
      "kubernetes.io/created-by": "{\"kind\":\"SerializedReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"DaemonSet\",\"namespace\":\"kube-system\",\"name\":\"kube-apiserver\",\"uid\":\"6e0d5a80-bc25-11e6-aab0-080027db44c2\",\"apiVersion\":\"extensions\",\"resourceVersion\":\"30\"}}\n"
    }
  },
```

The api server complains that `resourceVersion may not be set on objects to be created`.

So we do need to clean up some of that info. From the above list we would want to keep;

name
namespace
labels
And only our annotations - maybe this requirement changes, but we don't want those `kubernetes.io` ones.

xiang90:
@aaronlevy Thanks for looking into this. I will clean up the pod metadata soon.

xiang90:

xiang90:
I am going to try with annotations:

```
on pod

checkpoint.alpha.coreos.com/save=true

----
on checkpoint pod
checkpoint.alpha.coreos.com/checkpoint-of=kube-system/kube-apiserver
```

Opinions?

aaronlevy:
sgtm.

One thing comes to mind, (but maybe we can work around it for now).

If we're using the full pod name, we'll have something like:

```
checkpoint.alpha.coreos.com/checkpoint-of=kube-system/kube-apiserver-xxxxx
```

Where the suffix is randomly added (when using things like daemonsets/deployments).

Our resulting checkpointed static pod should end up automatically named something like:

```
kube-apiserver-xxxxx-1.1.1.1
```

Where the suffix is the nodename (because it is a static pod).

This so far is fine (and correct) -- our mapping of checkpoint to parent is 1-to-1.

But now let's say the parent pod is removed, and rescheduled. It will now have a different suffix.

```
kube-apiserver-yyyyy
```

Our logic would then keep running `kube-apiserver-xxxx-1.1.1.1` (because the parent is no longer running), but then also start checkpointing `kube-apiserver-yyyy`.

The simple workaround, I think, would be to still use a fixed name for the file on disk (some heuristic .. maybe just name prefix) - such that we overwrite the existing file, with the new checkpoint copy (then the logic will see that checkpoint now matches parent and should stop running). That feels a bit hackish -- but would require less changes that we could follow up on.

A less simple approach would be to also try and contact the apiserver and see if the pod should still in fact be scheduled. For example,

- kube-apiserver-xxxx-1.1.1.1 checkpoint is running (because its parent has died)
- kube-apiserver-yyyy pod gets scheduled to node (kube-apiserver-xxxx is no longer scheduled)
- kube-apiserver-yyyy starts creating a checkpoint with parent=kube-apiserver-yyyy
- checkpointer process is able to contact an api-server, sees that kube-apiserver-xxxx is no longer scheduled to the node, and GCs kube-apiserver-xxxx checkpoint.

I bring this up, because this might be more concerning for etcd pods. For example, we don't want etcd1 checkpoint to forever run on a node if it has potentially been rescheduled to another node.

xiang90:
@aaronlevy I understand your concern. That is also part of the reason I wanted to decouple the checkpoint annotation naming and the actual pod naming. Would you like to reconsider that approach?

```
on pod

checkpoint.alpha.coreos.com/checkpoint = kube-system/kube-apiserver

----
on checkpoint pod
checkpoint.alpha.coreos.com/checkpoint-of = kube-system/kube-apiserver

```

To solve the duplication problem, we can let the checkpointer to contact the actual api server to check the existence of original pod. But that requires to change the annotation to label I guess.

aaronlevy:
I'm still thinking we use the real pod name, as it truly is what we are tracking the checkpoint of. If we try and do this by arbitrary names - it could become problematic later (pods from deployments, multiple pods on a node, de-duping, etc).

As a simple workaround - we could still just use a fixed filename for now. Then we can follow up later with support of per-pod GC (but the annotations won't need to change). Could be as simple as `if startswith(pod.Name, "kube-apiserver") { filename="kube-apiserver.json" }`

xiang90:
> Could be as simple as if startswith(pod.Name, "kube-apiserver") { filename="kube-apiserver.json" }

I think if we do this, we go back forward. We should try not to have pod specific thing.

A simpler solutions might be kill all checkpoint pods if node has connected to the api server and has synced pods requirement with api server.

aaronlevy:
I'm not sure I fully understand.

On the one hand, I think the "most correct" is that we track per-pod checkpoints, rather than per-application. E.g. `kube-apiserver-abcde` vs `kube-apiserver`. This would be the most flexible, and allow us to support things like multiple pods being scheduled to the same node, and easy / accurate GC by namespace/podname, and when deploying an application - you don't need to know anything about *how* it is checkpointed (e.g. by particular name), just that it should be (save=true).

But I also don't want to pile all that work on you (as it encompasses more than just filtering on annotations, like GC on a per-pod basis - and per-pod secret/configmaps). So I'm mostly wondering about safe middle ground where we can still move toward that goal (why I'm leaning toward workarounds in the checkpointer rather than workarounds via annotations -- easier to change later).

xiang90:

cmd/checkpoint/main.go
xiang90: They are different things I think. checkpoints here actually track the ondisk manifests. so we need to translate the ondisk name (/xx/xx/api-server-a.json) -> api-server-a.

xiang90:

xiang90:

xiang90:

pbx0:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:

xiang90:
@aaronlevy OK. Test passed remotely and locally now. Manually checked the restart, checkpoint, everything seems to be fine.

No idea why bootkube-dev failed though. 

PTAL.

aaronlevy:
Closing, as this was implemented in https://github.com/kubernetes-incubator/bootkube/pull/238

PR #224, opened 2016-12-08
1.5.1 updates

pbx0:
rebased on master and updated to 1.5.1


pkg/bootkube/create.go
```diff
@@ -80,15 +80,19 @@ func createAssets(manifestDir string) error {
 		return err
 	}
 
-	includeExtendedAPIs := false
-	mapper, typer := f.Object(includeExtendedAPIs)
+	// aaron: is this ok that we can't specify the extend api option anymore?
```
aaronlevy: This is almost a verbatim copy of the kubectl implementation of `create -f manifestDir`. I'd re-check those changes if you haven't. Seems like as long as we support the same as that command we should be okay.

pkg/bootkube/create.go
```diff
@@ -116,7 +120,7 @@ func createAssets(manifestDir string) error {
 		if !shortOutput {
 			f.PrintObjectSpecificMessage(info.Object, util.GlogWriter{})
 		}
-		cmdutil.PrintSuccess(mapper, shortOutput, util.GlogWriter{}, info.Mapping.Resource, info.Name, "created")
+		cmdutil.PrintSuccess(mapper, shortOutput, util.GlogWriter{}, info.Mapping.Resource, info.Name, false, "created")
```
aaronlevy: What is the false flag here?

pkg/util/etcdutil/migrate.go
```diff
@@ -41,34 +41,36 @@ func Migrate() error {
...
-			return false, err
+		res := listETCDCluster(ns, restClient)
+		if res.Error() != nil {
+			return false, res.Error()
 		}
-		defer resp.Body.Close()
 
-		switch resp.StatusCode {
+		var status int
+		res.StatusCode(&status) // this func sig is very odd...
```
aaronlevy: nit: not sure the comment here is super helpful - unless we need to track that it is odd :) ...

pkg/util/etcdutil/migrate.go
```diff
@@ -80,8 +82,8 @@ func waitEtcdTPRReady(httpClient *http.Client, interval, timeout time.Duration,
 func getBootEtcdPodIP(kubecli clientset.Interface) (string, error) {
 	var ip string
 	err := wait.Poll(5*time.Second, 60*time.Second, func() (bool, error) {
-		podList, err := kubecli.Core().Pods(api.NamespaceSystem).List(api.ListOptions{
-			LabelSelector: labels.SelectorFromSet(labels.Set{"k8s-app": "boot-etcd"}),
+		podList, err := kubecli.CoreV1().Pods(api.NamespaceSystem).List(v1.ListOptions{
+			LabelSelector: labels.SelectorFromSet(labels.Set{"k8s-app": "boot-etcd"}).String(),
```
aaronlevy: If it just expects the selector string I believe we should be able to just use `"k8s-app=boot-etcd"`

pkg/util/etcdutil/migrate.go
```diff
@@ -119,16 +121,20 @@ func createMigratedEtcdCluster(httpcli *http.Client, host, podIP string) error {
...
 
-	resp, err := httpcli.Post(
-		fmt.Sprintf("%s/apis/coreos.com/v1/namespaces/kube-system/etcdclusters", host),
-		"application/json", bytes.NewReader(b))
-	if err != nil {
-		return err
+	uri := "/apis/coreos.com/v1/namespaces/kube-system/etcdclusters"
+	res := restclient.Post().RequestURI(uri).Body(b).Do()
+	// Where do i put "application/json" ?
+	//fmt.Sprintf("%s/apis/coreos.com/v1/namespaces/kube-system/etcdclusters", host),"application/json", bytes.NewReader(b))
```
aaronlevy: Yaml may not work at all at the moment: https://github.com/kubernetes/kubernetes/issues/37455 -- but maybe this instance was? Or does this convert the body for us to json?

Otherwise, you should be able to add `SetHeader` to the chain:

`restclient.Post().RequestURI(uri).SetHeader("content-type", "application/yaml").Body(b).Do()`

If it's still causing problems, may need to temporarily change it to json until the issue is resolved.

pbx0: Well we already have the json embedded here so I don't think we need to convert it to yaml. I'll add SetHeader to the function chain.

aaronlevy: Oh sorry, I mis-read this for some reason. We should be fine without explicitly setting the content type to json

pkg/bootkube/create.go
pbx0: I took a look at https://github.com/kubernetes/kubernetes/blob/v1.5.0-beta.3/pkg/kubectl/cmd/create.go#L100

As far as the section I was changing it looks pretty similar but notably the kubectl code calls `f.UnstructuredObject()`. Also the upstream visit function looks quite a bit different, but i'm not very familiar with this code.

pkg/util/etcdutil/migrate.go
pbx0: done

aaronlevy:
This overall looks good to me. Can you squash this maybe into a vendor+makefile commit and bump version commit?

pbx0:
@aaronlevy squashed all commits but the vendor commit.

PR #229, opened 2016-12-14
Add missing --cloud-provider parameter to kubelet

Quentin-M:
https://github.com/kubernetes-incubator/bootkube/pull/105 introduces usage of --cloud-provider in bootkube, however the parameter isn't passed to the kubelet.

@aaronlevy


aaronlevy:
The kubelet auto-detects the cloud-provider by default.

If we wanted to be explicit instead, the kubelet manifest doesn't currently take any template values, then also the on-host kubelet systemd units would also need to reflect this change. Rather than plumbing all that through - it's easier to just use the auto-detect.

Quentin-M:
Thanks. I didn't pay attention to these. Closing for now (at least).

PR #231, opened 2016-12-14
Checkpoint revert

aaronlevy:
We only got part way to where we wanted to be - and likely implementing this differently anyway. I was going to revert this for now so that we can layer some simple fixes on top (like logging and the name change). If we need to we can come back and revert the revert.

/cc @xiang90


aaronlevy:

xiang90:
lgtm

PR #232, opened 2016-12-14
Add support for template overrides

Quentin-M:
This PR enables developers who use bootkube as a library to override certain existing templates (with extra rendering data). For now, only dynamic assets may be overridden (API Server and Controller Manager).

When extra ConfigMaps are to be mounted (e.g. when `--authorization-policy-file=/etc/kubernetes/authz/policy.json` is used on the API server), I believe that they could be provided as part of the overridden template, as another resource.

This mainly is a draft and I'm indeed open to discussion.

### DIFF
--- a/pkg/asset/asset.go
+++ b/pkg/asset/asset.go
@@ -41,15 +41,23 @@ const (
 // AssetConfig holds all configuration needed when generating
 // the default set of assets.
 type Config struct {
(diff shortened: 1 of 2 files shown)


Quentin-M:
Ping ^

Quentin-M:
Closing, we discussed offline a better solution for this.

PR #233, opened 2016-12-14
rename api-checkpointer to pod-checkpointer

aaronlevy:
Will follow up with a rename in docs & bootkube -- need to cut a checkpoint image release (with this change) before those can be udpated.


xiang90:
lgtm

Issue #244, opened 2016-12-21
Add options to pass etcd certs to bootkube


dhawal55:
We are running etcd with mutual authentication enabled and bootkube doesn't provide any option to pass etcd certs to bootkube

dhawal55:
Any update on supporting secure etcd? I love all the cool stuff happening here, but I feel in order to really use bootkube in a production environment (like we are doing), adding support for secure etcd is a must. I feel this should be a higher priority than self-hosted etcd. Let me know if you guys think otherwise.

I would love to stop maintaining my fork and start using the latest bootkube for my production cluster.

Issue #31, opened 2016-05-02
Support deploying self-hosted etcd


philips:
From the README: 

> When you start bootkube, you must also give it the addresses of your etcd servers, and enough information for bootkube to create an ssh tunnel to the node that will become a member of the master control plane. Upon startup, bootkube will create a reverse proxy using an ssh connection, which will allow a bootstrap kubelet to contact the apiserver running as part of bootkube.

In the original prototype we had a built in etcd. Why is that no longer part of this?

aaronlevy:
We need the data that ends up in etcd to persist with the cluster that is launched. If that data lives in the bootkube, then bootkube must continue to run for the lifecycle of the cluster.

Alternatively, we need a way to pivot the etcd data injected during the bootstrap process to the "long-lived" etcd cluster. The long-lived cluster would essentially be a self-hosted etcd cluster launched as k8s components just like the rest of the control plane.

What I'd probably like to see is something along the lines of:

1. Bootkube runs etcd in-process
1. k8s objects injected to the api-server end up in the local/in-process etcd
1. One of those objects is an etcd pod definition, which is started as a self-hosted pod on a node.
1. The self hosted etcd "joins" the existing bootkube etcd, making a cluster of 2 nodes.
1. etcd replication copies all state to new joined etcd node
1. bootkube dies after a self-hosted control-plane is started, removing itself from etcd cluster membership
1. self-hosted etcd cluster is managed from that point forward as a k8s component.

Another option might be trying to copy the etcd keys from the in-process/local node to the self-hosted node, but this can get a little messy because we would be trying to manually copy (and mirror) data of a live cluster.

Some concerns with this approach:

* Managing etcd membership in K8s is not currently a very good story. It's either waiting on petsets or trying to handle this with lifecycle hooks, or relying on external mechanics for membership management.

* Pretty unproven and a bit risky from a production perspective to try and run etcd for the cluster, also "in" the cluster. But I can see the value in this from a "get started easily" while we exercise this as a viable option.

aaronlevy:
@philips what do you think about changing this issue to be "support self-host etcd", and dropping from 0.1.0 milestone ?

aaronlevy:
Adding notes from a side-discussion:

Another option that was mentioned is just copying keys from bootkube-etcd to cluster-etcd. This would require some coordination points in the bootkube process:

1. bootkube-apiserver configured to use bootkube-etcd
[...]

PR #245, opened 2016-12-21
Allow talking to secure etcd (authentication through client certs)

dhawal55:
Fixes issue #244
We will need more work to support secure etcd for self-hosted etcd


pbx0:
OK to test

pbx0:
ok to test

dhawal55:
Can anyone look at my pull requests? Really simple ones

aaronlevy:
Hi @dhawal55, sorry for the delay.

We are planning on migrating to self-hosted etcd cluster as the default very soon (https://github.com/kubernetes-incubator/bootkube/issues/31) -- so the way each etcd node is launched (and where it sources TLS assets) will change almost completely. This is definitely something we need, but if we try to implement this functionality now, it will pretty much have to be thrown out / re-done differently very soon.

dhawal55:
@aaronlevy The `experimental-self-hosted-etcd` currently doesn't support TLS, nor does the etcd-operator. I'm okay closing this PR if issue #31 includes TLS support for etcd.

aaronlevy:
https://github.com/kubernetes-incubator/bootkube/issues/31 doesn't explicitly cover handling TLS support (and the initial implementation likely won't). I think we should keep https://github.com/kubernetes-incubator/bootkube/issues/244 open to track this need.

aaronlevy:
This has been added in https://github.com/kubernetes-incubator/bootkube/pull/433

PR #247, opened 2016-12-29
Configure pod-network via flannel CNI daemonset

aaronlevy:
This change involves a switch to using the Container Network Interface (CNI) plugins for configuring the pod network. See the http://kubernetes.io/docs/admin/network-plugins/ for more information  on kubernetes + CNI plugin integration.

There is a known issue with using CNI plugins that setting a `hostPort` on a pod will no longer work. This is because the `hostPort` relies on docker to configure the host -> pod bridge, but in the case of CNI network plugins, docker does not know about the pod, network and the CNI plugin is used instead. 

See upstream issues for more information:
https://github.com/kubernetes/kubernetes/issues/23920
https://github.com/kubernetes/kubernetes/issues/31307

The most common place I've seen this being surfaced as an issue is with ingress controllers, which make use of pod hostPorts. Workarounds to this include:

- Using `hostNetwork=true` on the pod so that all containers in the pod will use the host network namespace, but then also setting the `hostPort` so that the scheduler can take this constraint into account when scheduling pods (one pod per host, as it will already be taken). The downside to this is that all containers in the pod will use the host network, not just a specific container.

- Another option is to use a [nodePort service](http://kubernetes.io/docs/user-guide/services/#type-nodeport) in front of the ingress controller so that a port will be available on all nodes running kube-proxy, which can then route to the ingress controller. The downside to this option is that the port will be in the (default) range of 30000-32000, and if you want a service exposed on `:443`, for example, the port mapping must be done externally.

### DIFF
--- a/hack/multi-node/user-data.sample
+++ b/hack/multi-node/user-data.sample
@@ -1,24 +1,7 @@
 #cloud-config
 
 coreos:
-  flannel:
-    interface: $public_ipv4
-    etcd_endpoints: http://172.17.4.51:2379
   units:
-    - name: flanneld.service
-      command: start
-      drop-ins:
-      - name: 50-network-config.conf
-        content: |
-          [Service]
-          ExecStartPre=/usr/bin/etcdctl --endpoint=http://172.17.4.51:2379 set /coreos.com/network/config '{ "Network": "10.2.0.0/16" }'
-    - name: docker.service
-      drop-ins:
-      - name: 50-flannel.conf
-        content: |
-          [Unit]
-          Requires=flanneld.service
-          After=flanneld.service
     - name: kubelet.service
       enable: true
       command: start
@@ -27,12 +10,17 @@ coreos:
         EnvironmentFile=/etc/environment
         Environment=KUBELET_ACI=quay.io/coreos/hyperkube
         Environment=KUBELET_VERSION=v1.5.1_coreos.0
+        Environment="RKT_OPTS=--volume var-lib-cni,kind=host,source=/var/lib/cni --mount volume=var-lib-cni,target=/var/lib/cni"
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/manifests
-        ExecStartPre=/bin/mkdir -p /srv/kubernetes/manifests
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes/cni/net.d
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/checkpoint-secrets
+        ExecStartPre=/bin/mkdir -p /srv/kubernetes/manifests
+        ExecStartPre=/bin/mkdir -p /var/lib/cni
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --kubeconfig=/etc/kubernetes/kubeconfig \
           --require-kubeconfig \
+          --cni-conf-dir=/etc/kubernetes/cni/net.d \
+          --network-plugin=cni \
           --lock-file=/var/run/lock/kubelet.lock \
           --exit-on-lock-contention \
           --pod-manifest-path=/etc/kubernetes/manifests \
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -19,6 +19,7 @@ contexts:
     cluster: local
     user: kubelet
 `)
+
 	KubeletTemplate = []byte(`apiVersion: extensions/v1beta1
 kind: DaemonSet
 metadata:
@@ -36,13 +37,11 @@ spec:
       - name: kubelet
         image: quay.io/coreos/hyperkube:v1.5.1_coreos.0
         command:
-        - /nsenter
-        - --target=1
-        - --mount
-        - --wd=.
-        - --
         - ./hyperkube
         - kubelet
+        - --network-plugin=cni
+        - --cni-conf-dir=/etc/kubernetes/cni/net.d
+        - --cni-bin-dir=/opt/cni/bin
         - --pod-manifest-path=/etc/kubernetes/manifests
         - --allow-privileged
         - --hostname-override=$(NODE_NAME)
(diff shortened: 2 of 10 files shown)


hack/multi-node/user-data.sample
```diff
@@ -27,12 +10,17 @@ coreos:
...
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/manifests
-        ExecStartPre=/bin/mkdir -p /srv/kubernetes/manifests
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes/cni/net.d
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/checkpoint-secrets
+        ExecStartPre=/bin/mkdir -p /srv/kubernetes/manifests
+        ExecStartPre=/bin/mkdir -p /var/lib/cni
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --kubeconfig=/etc/kubernetes/kubeconfig \
           --require-kubeconfig \
+          --cni-conf-dir=/etc/kubernetes/cni/net.d \
```
yifan-gu: question: where does the flannel binary live?

pkg/asset/internal/templates.go
```diff
@@ -513,4 +523,92 @@ spec:
...
+  template:
+    metadata:
+      labels:
+        tier: node
+        app: flannel
+    spec:
+      hostNetwork: true
+      containers:
+      - name: kube-flannel
+        image: quay.io/coreos/flannel-git:v0.6.1-60-g4b752d1-amd64
```
dghubble: Why does this differ from the `quay.io/coreos/flannel` image?

pkg/asset/internal/templates.go
```diff
@@ -513,4 +523,92 @@ spec:
...
+        env:
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: POD_IP
```
dghubble: Why does this need to be set? Its not used in the flanneld service shipped on CoreOS for example, the default is used.

pkg/asset/internal/templates.go
```diff
@@ -51,6 +49,7 @@ spec:
         - --kubeconfig=/etc/kubernetes/kubeconfig
         - --require-kubeconfig
         - --lock-file=/var/run/lock/kubelet.lock
+        - --containerized
```
dghubble: Is this new? How was kubelet pod running without this?

hack/multi-node/user-data.sample
aaronlevy: The flannel daemon is now run as a pod, but the flannel-cni plugin exists in the hyperkube container fs at [/opt/bin/cni](https://github.com/kubernetes/kubernetes/blob/v1.5.1/cluster/images/hyperkube/Dockerfile#L62)

pkg/asset/internal/templates.go
aaronlevy: It is new. It was running because we were using the nsenter workaround which was removed above. The problem with the existing option is that we are using the host mount namespace - which does not contain things like CNI plugins from the hyperkube image.

I had intended to make this change anyway in https://github.com/kubernetes-incubator/bootkube/pull/162 -- as it fixes the issue with port forwarding, but this ends up being required to support the CNI plugins.

pkg/asset/internal/templates.go
aaronlevy: Because there has not yet been a release of flannel which contains required changes. /cc @philips @tomdee -- the changes included in this release are passing k8s conformance tests, is there some additional criteria we should meet to get the changes into a release?

pkg/asset/internal/templates.go
aaronlevy: We normally would have set this via an environment file on the host (e.g. https://github.com/kubernetes-incubator/bootkube/blob/v0.3.0/hack/quickstart/init-master.sh#L38)

Flannel will guess which interface based on the default route -- unfortunately, the default route for virtualbox is an interface which is not routable between nodes.

Not sure on AWS/GCE -- but I'm also assuming you would want to use the private network (when using hostNetwork + podIP -- this will be the IP of the interface which is used as the node name).

hack/multi-node/user-data.sample
yifan-gu: @aaronlevy where is that cni-bin from, do we build from the latest cni repo every time?
Maybe adding some comment here would be helpful since it feels very implicit.

yifan-gu:
LGTM from my side as long as the CI is green.

hack/multi-node/user-data.sample
aaronlevy: It's part of the hyperkube image build process (so version is selected upstream: https://github.com/kubernetes/kubernetes/blob/v1.5.1/cluster/images/hyperkube/Makefile#L23)

I can add the `--cni-bin-dir=/opt/bin/cni` flag to be more explicit.

aaronlevy: added the flag

pkg/asset/internal/templates.go
```diff
@@ -41,6 +41,7 @@ spec:
         - kubelet
         - --network-plugin=cni
         - --cni-conf-dir=/etc/kubernetes/cni/net.d
+		- --cni-bin-dir=/opt/cni/bin
```
yifan-gu: looks like it's 2 tabs here vs 8 spaces elsewhere?

aaronlevy: woops

aaronlevy:
coreosbot test this please

aaronlevy:

aaronlevy:
coreosbot test this please

dghubble:
I'll try this with coreos-baremetal self-hosted clusters.

aaronlevy:
Ran this through conformance tests, and all passed. Going to go ahead and merge

pkg/asset/internal/templates.go
tomdee: I know this PR is merged now, but FTR I'm planning to do a release this week

pkg/asset/internal/templates.go
euank: Doesn't this break `mount` binaries that are present in hyperkube but not on container linux?

Issue #106, opened 2016-08-20
Provide a generic mechanism for flag customizations


aaronlevy:
My initial concern is that I don’t want to plumb through every possible configurable through bootkube -- because it just becomes unnecessarily complex. However, people have differing needs so we need a sane way of allowing customization.

I was thinking of breaking the rendering steps into multiple pieces
## init

command: `bootkube init`

This would output configMap objects (kubernetes objects themselves) with all of the default values we want for each component. So api-server for example would have a configMap with key=value for each of the flags we default.

At this stage a deployer can just go in and modify each component configMap to the values they would like. This also will help us in the future because eventually all core components will be able to retrieve their config from an api object (componentConfig).
## validate

command: `bootkube validate`

This step could be run standalone against the configMap objects, and/or as part of the next `bootkube render` step.

Here we will validate that options provided in the configMaps are compatible / recommended. For example:

Setting controller-manager `--service-cluster-ip-range=10.0.0.0/16` and kubelet `--cluster-dns=172.14.0.1` is likely an error and we can warn.

Or

Setting `--cloud-provider=aws` on the controller-manager, but not in kubelet & api-server are likely a misconfiguration.

This way, flags  and their inter-related dependencies can be modeled as a suite of tests -- rather than us trying to plumb this logic through bootkube / templating. And an end user has the same flexibility to modify any flags they want with no code changes necessary (and if validation issues arise - it is just a matter of adding additional tests).
## render

command: `bootkube render`

This step will be modified to get all configuration from the configMaps created above (and run the same validation from above.  After validation, we do the same thing we were before: render all of the necessary component manifests -- but have them reference the configMaps for their flags.

To use the configMaps as flags directly (until it is supported upstream), we would need to figure out the best way to convert the key=values to `--key=value` for use in the command line of the manifest (e.g. `command: /hyperkube apis-server $flags`)

Or as another option, we just let the render step do the actual conversion from the configMap object to the manifest -- and we don't directly use the configMap objects via the api until supported upstream

aaronlevy:
Another item brought up is that some flags are going to be common (but cannot be assumed as defaults). Such as `--cloud-provider=foo` (https://github.com/coreos/bootkube/pull/105#issuecomment-241238375)

I'm wondering about the option of layering multiple configMaps, such that we could say something like:

`bootkube init --profiles="default,cloudprovider-aws"`

[...]

PR #250, opened 2016-12-29
Added service-cluster-ip-range as flag to bootkube start

### LINKED ISSUE #236: Add flag to pass cluster CIDR to bootkube render
When I run bootkube to render assets, it assumes service-cluster-ip-range to be 10.3.0.0/24 and allocates etcd clusterIP and kube-dns clusterIP from that range. Make this as a parameter that can be passed to `bootkube render`.

Why does kube-proxy daemonset need to pass `--cluster-cidr` to `hyperkube proxy` image? Documentation says "It is used to bridge traffic coming from outside of the cluster". How does it work when using flanneld (its value is same as that set in etcd for /coreos.com/network/config Network value) and can this be made a parameter too so if not passed, it will not be set?

dhawal55:
Fixes #236

### DIFF
--- a/cmd/bootkube/start.go
+++ b/cmd/bootkube/start.go
@@ -24,9 +24,10 @@ var (
 	}
 
 	startOpts struct {
-		assetDir       string
-		etcdServer     string
-		selfHostedEtcd bool
+		assetDir              string
+		etcdServer            string
+		selfHostedEtcd        bool
+		serviceClusterIPRange string
 	}
 )
 
@@ -35,6 +36,7 @@ func init() {
 	cmdStart.Flags().StringVar(&startOpts.etcdServer, "etcd-server", "http://127.0.0.1:2379", "Single etcd node to use during bootkube bootstrap process.")
 	cmdStart.Flags().StringVar(&startOpts.assetDir, "asset-dir", "", "Path to the cluster asset directory. Expected layout genereted by the `bootkube render` command.")
 	cmdStart.Flags().BoolVar(&startOpts.selfHostedEtcd, "experimental-self-hosted-etcd", false, "Self hosted etcd mode. Includes starting the initial etcd member by bootkube.")
+	cmdStart.Flags().StringVar(&startOpts.serviceClusterIPRange, "service-cluster-ip-range", "10.3.0.0/24", "A CIDR notation IP range from which to assign service cluster IPs.")
 }
 
 func runCmdStart(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


aaronlevy:
Changing of the service IP ends up touching many different components. And while it really does need to be a configurable -- I'm hesitant to try and plumb this through as a configuration option right now.

As an example (and I'm likely forgetting things) - For this change to work properly, it also requires changes to the on-host kubelet, self-hosted kubelet, self-hosted controller-manager,  dns service manifest, validation of podCidrs, and templating tls asset generation.

It ends up being a fairly intrusive change. And while I'm not outright opposed to making this a configurable, it does open a pretty big door for complexity. Any change which would require coordination of flags between `bootkube render` and `bootkube start` means that we can leave users in a place where generated assets no longer work unless you use the correct flags (which I really want to avoid).

Longer term, I believe minimizing this risk will come in a couple stages:

* Changing `bootkube start` to using static manifests for the temporary control-plane, rather than compiled-in components (https://github.com/kubernetes-incubator/bootkube/issues/168). This will mean that `bootkube render` will create all the information needed for the temporary control-plane, and `bootkube start` can source all of that information from static manifests (not runtime flags).

* Changing `bootkube render` to source from versioned configuration, rather than runtime flags. See some of the brainstorming discussion here: https://github.com/kubernetes-incubator/bootkube/issues/106

dhawal55:
@aaronlevy I like the idea of `bootkube start` using static manifests for temporary control-plane. Currently, I'm not using `bootkube render` to generate assets becuase there are lots of configuration missing. I generate my own assets dir and pass it to `bootkube start`. It makes sense to do the same for temporary control-plane too. Will the init_master.sh script be responsible for copying over the static manifests to `/etc/kubernetes/manifests` directory?

aaronlevy:
@dhawal55 init-master.sh just copies from the remote location to your local box (so that you have a copy of things like the CA cert, etc.). But it merely executes bootkube remotely.

I think in the new mode of operation we would still do the same (execute bootkube remotely) -- but with `bootkube start --cluster-dir=foo`, the cluster-dir would also contain manifests for the "temporary" control plane. 

This would allow you to similarly not use `render` (or use, then modify) -- and self provide those assets however you see fit (and just reference a `--cluster-dir`)

aaronlevy:
I'm going to close this PR in favor of a (yet to be implemented) alternative solution discussed in: https://github.com/kubernetes-incubator/bootkube/issues/236#issuecomment-279797787

PR #257, opened 2017-01-04
asset: set the advertise address based on pod ip

aaronlevy:
This flag was removed in https://github.com/kubernetes-incubator/bootkube/pull/151, but looks like we do still need this in some cases.

The bootkube-apiserver was setting the correct endpoint record for the default kubernetes service (`kubectl get endpoint/kubernetes`). However, after bootkube exits, the self-hosted api-server will eventually overwrite the endpoint again -- and it was using the host's default interface, which is not routable in some cases (e.g. virtualbox).

There may be some other options here (e.g. why does bootkube apiserver use correct interface, but hostNetwork api-server pod does not) -- but going to revert this part of the change so we avoid the regression until we have a better option.

/cc @dghubble @quentin-m @bassam

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -139,6 +139,7 @@ spec:
         - --bind-address=0.0.0.0
         - --secure-port=443
         - --insecure-port=8080
+        - --advertise-address=$(POD_IP)
         - --etcd-servers={{ range $i, $e := .EtcdServers }}{{ if $i }},{{end}}{{ $e }}{{end}}
         - --storage-backend={{.StorageBackend}}
         - --allow-privileged=true
@@ -151,6 +152,11 @@ spec:
         - --client-ca-file=/etc/kubernetes/secrets/ca.crt
         - --cloud-provider={{ .CloudProvider  }}
         - --anonymous-auth=false
+        env:
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
         volumeMounts:
         - mountPath: /etc/ssl/certs
           name: ssl-certs-host


bassam:
I'm still seeing the same behavior with this change. bootkube times out and the pods are in the following state:

```
> kubectl --kubeconfig=cluster/auth/kubeconfig get pods --all-namespaces
NAMESPACE     NAME                                       READY     STATUS              RESTARTS   AGE
kube-system   checkpoint-installer-rj673                 1/1       Running             0          8m
kube-system   kube-apiserver-153w3                       0/1       CrashLoopBackOff    6          8m
kube-system   kube-controller-manager-3438979800-sltl6   0/1       ContainerCreating   0          9m
kube-system   kube-controller-manager-3438979800-zrhg6   0/1       ContainerCreating   0          9m
kube-system   kube-dns-4101612645-j08jp                  0/4       ContainerCreating   0          9m
kube-system   kube-flannel-7m2fl                         2/2       Running             0          8m
kube-system   kube-flannel-t050l                         1/2       CrashLoopBackOff    6          8m
kube-system   kube-proxy-bwcpb                           1/1       Running             0          8m
kube-system   kube-proxy-gjc7x                           1/1       Running             0          8m
kube-system   kube-scheduler-694795526-0ktfv             0/1       ContainerCreating   0          9m
kube-system   kube-scheduler-694795526-g5z6f             0/1       ContainerCreating   0          9m
kube-system   pod-checkpointer-172.17.4.101              1/1       Running             0          7m
```

bassam:
@aaronlevy things are a lot better if I revert #247:

```
> kubectl --kubeconfig=cluster/auth/kubeconfig get pods --all-namespaces
kube-system   checkpoint-installer-tmf2c                 1/1       Running   0          3m
kube-system   kube-apiserver-9d5zg                       1/1       Running   3          3m
kube-system   kube-controller-manager-2751499733-0s565   1/1       Running   0          4m
kube-system   kube-controller-manager-2751499733-8z5h3   1/1       Running   0          4m
kube-system   kube-dns-4101612645-th3nx                  4/4       Running   0          4m
kube-system   kube-proxy-8vsc4                           1/1       Running   0          3m
kube-system   kube-proxy-q83rs                           1/1       Running   0          1m
kube-system   kube-scheduler-694795526-blt33             1/1       Running   0          4m
kube-system   kube-scheduler-694795526-l641q             1/1       Running   0          4m
kube-system   pod-checkpointer-172.17.4.101              1/1       Running   0          2m
```

dghubble:
This may mean coreos-baremetal needs to use a fork again :(

aaronlevy:
@dghubble there may be other options here - but it is a regression, so need to fix in the short term.

PR #259, opened 2017-01-04
fix: self hosted etcd

hongchaodeng:
See https://github.com/coreos/etcd-operator/issues/594#issuecomment-270475704

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ conformance-%: clean all
 # This will naively try and create a vendor dir from a k8s release
 # USE: make vendor VENDOR_VERSION=vX.Y.Z
 VENDOR_VERSION = v1.5.2+coreos.1
+ETCD_OPERATOR_VERSION = 7315cda48a511c8bcab1e575ee0626eae7d07d2d
 vendor:
 	@echo "Creating k8s vendor for: $(VENDOR_VERSION)"
 	@rm -rf vendor
@@ -66,6 +67,12 @@ vendor:
 	@mv $@/k8s.io/kubernetes/vendor/* $(abspath $@)
 	@cd $@/k8s.io/ && ln -sf kubernetes/staging/src/k8s.io/client-go client-go
 	@rm -rf $@/k8s.io/kubernetes/vendor $@/k8s.io/kubernetes/.git
+	@echo "vendoring etcd-operator spec"
+	@git clone https://github.com/coreos/etcd-operator /tmp/etcd-operator > /dev/null 2>&1
+	@cd /tmp/etcd-operator && git checkout $(ETCD_OPERATOR_VERSION) > /dev/null 2>&1
+	@mkdir -p $@/github.com/coreos/etcd-operator/pkg/
+	@cp -r /tmp/etcd-operator/pkg/spec $@/github.com/coreos/etcd-operator/pkg/
+	@rm -rf /tmp/etcd-operator
 
 clean:
 	rm -rf _output
--- a/pkg/util/etcdutil/migrate.go
+++ b/pkg/util/etcdutil/migrate.go
@@ -1,14 +1,17 @@
 package etcdutil
 
 import (
+	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
-	"net/http"
 	"time"
 
+	"github.com/coreos/etcd-operator/pkg/spec"
 	"github.com/coreos/etcd/clientv3"
 	"github.com/golang/glog"
-	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/api"
+	apierrors "k8s.io/kubernetes/pkg/api/errors"
 	"k8s.io/kubernetes/pkg/api/v1"
 	clientset "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_5"
 	"k8s.io/kubernetes/pkg/client/restclient"
@@ -18,6 +21,13 @@ import (
 const (
 	apiserverAddr = "http://127.0.0.1:8080"
 	etcdServiceIP = "10.3.0.15"
+
+	etcdClusterName = "kube-etcd"
+)
+
+var (
+	waitEtcdClusterRunningTime = 300 * time.Second
+	waitBootEtcdRemovedTime    = 300 * time.Second
 )
 
 func Migrate() error {
@@ -33,6 +43,7 @@ func Migrate() error {
 	if err != nil {
 		return err
 	}
+	glog.Infof("etcd cluster TPR is setup")
 
 	ip, err := getBootEtcdPodIP(kubecli)
 	if err != nil {
@@ -41,11 +52,20 @@ func Migrate() error {
 	glog.Infof("boot-etcd pod IP is: %s", ip)
 
 	if err := createMigratedEtcdCluster(restClient, apiserverAddr, ip); err != nil {
-		glog.Errorf("fail to create migrated etcd cluster: %v", err)
-		return err
+		return fmt.Errorf("fail to create migrated etcd cluster: %v", err)
 	}
+	glog.Infof("etcd cluster for migration is created")
 
-	return checkEtcdClusterUp()
+	if err := waitEtcdClusterRunning(restClient); err != nil {
+		return fmt.Errorf("wait etcd cluster running failed: %v", err)
+	}
+	glog.Info("etcd cluster for migration is now running")
+
+	if err := waitBootEtcdRemoved(); err != nil {
+		return fmt.Errorf("wait boot etcd deleted failed: %v", err)
+	}
+	glog.Info("the boot etcd is removed from the migration cluster")
+	return nil
 }
 
 func listETCDCluster(ns string, restClient restclient.Interface) restclient.Result {
@@ -56,21 +76,13 @@ func listETCDCluster(ns string, restClient restclient.Interface) restclient.Resu
 func waitEtcdTPRReady(restClient restclient.Interface, interval, timeout time.Duration, ns string) error {
 	err := wait.Poll(interval, timeout, func() (bool, error) {
 		res := listETCDCluster(ns, restClient)
-		if res.Error() != nil {
-			return false, res.Error()
+		if err := res.Error(); err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
 		}
-
-		var status int
-		res.StatusCode(&status)
-
-		switch status {
-		case http.StatusOK:
-			return true, nil
-		case http.StatusNotFound: // not set up yet. wait.
-			return false, nil
-		default:
-			return false, fmt.Errorf("invalid status code: %v", status)
-		}
+		return true, nil
 	})
 	if err != nil {
 		return fmt.Errorf("fail to wait etcd TPR to be ready: %v", err)
@@ -108,7 +120,7 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
   "apiVersion": "coreos.com/v1",
   "kind": "EtcdCluster",
   "metadata": {
-    "name": "etcd-cluster",
+    "name": "%s",
     "namespace": "kube-system"
   },
   "spec": {
@@ -118,59 +130,74 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
 		"bootMemberClientEndpoint": "http://%s:12379"
     }
   }
-}`, podIP))
+}`, etcdClusterName, podIP))
 
 	uri := "/apis/coreos.com/v1/namespaces/kube-system/etcdclusters"
 	res := restclient.Post().RequestURI(uri).SetHeader("Content-Type", "application/json").Body(b).Do()
 
-	if res.Error() != nil {
-		return res.Error()
-	}
-
-	var status int
-	res.StatusCode(&status)
-
-	if status != http.StatusCreated {
-		return fmt.Errorf("fail to create etcd cluster object, status (%v), object (%s)", status, string(b))
-	}
-	return nil
+	return res.Error()
 }
 
-func checkEtcdClusterUp() error {
+func waitEtcdClusterRunning(restclient restclient.Interface) error {
 	glog.Infof("initial delaying (30s)...")
 	time.Sleep(30 * time.Second)
 
-	// The checking does:
-	// - Trying to talk to etcd cluster via etcd service. The assumption here is that
-	//   the etcd service only selects the etcd pods newly created, excluding the boot one.
-	//   Once we can talk to it, we are sure that etcd cluster is created.
-	// - Then we list members and see if it's been reduced to 1 member. Because when we
-	//   can talk to the etcd cluster, we are certain there are 2 members at the beginning,
-	//   and will reduce to 1 eventually. That's the timeline of expected events.
-	//   As long as 1 member cluster is reached, we are certain cluster has been migrated successfully.
-	err := wait.PollImmediate(10*time.Second, 60*time.Second, func() (bool, error) {
+	err := wait.Poll(10*time.Second, waitEtcdClusterRunningTime, func() (bool, error) {
+		b, err := restclient.Get().RequestURI(makeEtcdClusterURI(etcdClusterName)).DoRaw()
+		if err != nil {
+			return false, fmt.Errorf("fail to get etcdcluster: %v", err)
+		}
+
+		e := &spec.EtcdCluster{}
+		if err := json.Unmarshal(b, e); err != nil {
+			return false, err
+		}
+		if e.Status == nil {
+			return false, nil
+		}
+		switch e.Status.Phase {
+		case spec.ClusterPhaseRunning:
+			return true, nil
+		case spec.ClusterPhaseFailed:
+			return false, errors.New("failed to create etcd cluster")
+		default:
+			// All the other phases are not ready
+			return false, nil
+		}
+	})
+	return err
+}
+
+func waitBootEtcdRemoved() error {
+	err := wait.Poll(10*time.Second, waitBootEtcdRemovedTime, func() (bool, error) {
 		cfg := clientv3.Config{
 			Endpoints:   []string{fmt.Sprintf("http://%s:2379", etcdServiceIP)},
 			DialTimeout: 5 * time.Second,
 		}
 		etcdcli, err := clientv3.New(cfg)
 		if err != nil {
-			glog.Errorf("fail to create etcd client, retrying...: %v", err)
+			glog.Errorf("fail to create etcd client, will retry: %v", err)
 			return false, nil
 		}
+
 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-		defer cancel()
 		m, err := etcdcli.MemberList(ctx)
+		cancel()
+		etcdcli.Close()
 		if err != nil {
-			glog.Errorf("fail to list etcd members, retrying...: %v", err)
+			glog.Errorf("fail to list member, will retry: %v", err)
 			return false, nil
 		}
+
 		if len(m.Members) != 1 {
-			glog.Infof("Still migrating boot etcd member, retrying...")
+			glog.Info("Still waiting boot etcd to be deletd...")
 			return false, nil
 		}
-		glog.Infof("etcd cluster is up. Member: %v", m.Members[0].Name)
 		return true, nil
 	})
 	return err
+}
+
+func makeEtcdClusterURI(name string) string {
+	return fmt.Sprintf("/apis/coreos.com/v1/namespaces/kube-system/etcdclusters/%s", name)
 }
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -529,7 +529,7 @@ metadata:
 spec:
   selector:
     app: etcd
-    etcd_cluster: etcd-cluster
+    etcd_cluster: kube-etcd
   clusterIP: 10.3.0.15
   ports:
   - name: client
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -513,7 +513,7 @@ spec:
     spec:
       containers:
       - name: etcd-operator
-        image: quay.io/coreos/etcd-operator
+        image: quay.io/coreos/etcd-operator:c391d8b7638deb81aa877773a0acce389f602415
         env:
         - name: MY_POD_NAMESPACE
           valueFrom:
--- a/vendor/github.com/coreos/etcd-operator/pkg/spec/backup.go
+++ b/vendor/github.com/coreos/etcd-operator/pkg/spec/backup.go
@@ -1,0 +1,50 @@
+// Copyright 2016 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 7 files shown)


pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +139,34 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
+		if err := d.Decode(e); err != nil {
+			return false, err
 		}
-		glog.Infof("etcd cluster is up. Member: %v", m.Members[0].Name)
-		return true, nil
+		switch e.Status.Phase {
+		case spec.ClusterPhaseRunning:
+			return true, nil
+		case spec.ClusterPhaseFailed:
+			return false, errors.New("etcd cluster failed")
```
xiang90: failed to create etcd cluster

pkg/util/etcdutil/migrate.go
```diff
@@ -18,6 +19,8 @@ import (
 const (
 	apiserverAddr = "http://127.0.0.1:8080"
 	etcdServiceIP = "10.3.0.15"
+
+	etcdClusterName = "etcd-cluster"
```
xiang90: @aaronlevy we probably need a better name for it. self-hosted-etcd? system-etcd? kube-etcd?

aaronlevy: system-etcd or kube-etcd both seem good.

Then you would interact with it as:

```
kubectl get etcdclusters/kube-etcd
or
kubectl get etcdclusters/system-etcd
```

@yifan-gu any bikeshed thoughts here?

yifan-gu: Give just my 2 cents to kube-etcd, which gives me more information by the first glance.

hongchaodeng:
@aaronlevy fixed. PTAL

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +134,36 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
-	//   and will reduce to 1 eventually. That's the timeline of expected events.
-	//   As long as 1 member cluster is reached, we are certain cluster has been migrated successfully.
-	err := wait.PollImmediate(10*time.Second, 60*time.Second, func() (bool, error) {
-		cfg := clientv3.Config{
-			Endpoints:   []string{fmt.Sprintf("http://%s:2379", etcdServiceIP)},
-			DialTimeout: 5 * time.Second,
+	err := wait.Poll(10*time.Second, 60*time.Second, func() (bool, error) {
+		uri := fmt.Sprintf("%s/apis/coreos.com/v1/namespaces/kube-system/etcdclusters/%s", host, etcdClusterName)
+		res := restclient.Get().RequestURI(uri).Do()
+		if res.Error() != nil {
```
aaronlevy: I had to add: https://github.com/kubernetes-incubator/bootkube/compare/master...aaronlevy:sh-etcd#diff-2d3c8cde0a58be225d2f6d6b5d64059fR68 -- otherwise it would exit early if the TPR hadn't been created yet.

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +134,36 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
-	//   Once we can talk to it, we are sure that etcd cluster is created.
-	// - Then we list members and see if it's been reduced to 1 member. Because when we
-	//   can talk to the etcd cluster, we are certain there are 2 members at the beginning,
-	//   and will reduce to 1 eventually. That's the timeline of expected events.
-	//   As long as 1 member cluster is reached, we are certain cluster has been migrated successfully.
-	err := wait.PollImmediate(10*time.Second, 60*time.Second, func() (bool, error) {
-		cfg := clientv3.Config{
-			Endpoints:   []string{fmt.Sprintf("http://%s:2379", etcdServiceIP)},
-			DialTimeout: 5 * time.Second,
+	err := wait.Poll(10*time.Second, 60*time.Second, func() (bool, error) {
```
aaronlevy: Can you increase the total timeout (5 min?), and maybe make a const.

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +134,36 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
+		case spec.ClusterPhaseRunning:
+			return true, nil
+		case spec.ClusterPhaseFailed:
+			return false, errors.New("failed to create etcd cluster")
 		}
-		glog.Infof("etcd cluster is up. Member: %v", m.Members[0].Name)
-		return true, nil
+		return false, nil
 	})
 	return err
```
aaronlevy: Add some error context to this error.

aaronlevy: Otherwise it could just say "timeout reached" but not what it was doing

pkg/util/etcdutil/migrate.go
hongchaodeng: This is called after etcd cluster TPR is created. We don't need to check not found here, right?

hongchaodeng:
@aaronlevy 
fixed. PTAL

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +138,36 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
-	//   Once we can talk to it, we are sure that etcd cluster is created.
-	// - Then we list members and see if it's been reduced to 1 member. Because when we
-	//   can talk to the etcd cluster, we are certain there are 2 members at the beginning,
-	//   and will reduce to 1 eventually. That's the timeline of expected events.
-	//   As long as 1 member cluster is reached, we are certain cluster has been migrated successfully.
-	err := wait.PollImmediate(10*time.Second, 60*time.Second, func() (bool, error) {
-		cfg := clientv3.Config{
-			Endpoints:   []string{fmt.Sprintf("http://%s:2379", etcdServiceIP)},
-			DialTimeout: 5 * time.Second,
+	err := wait.Poll(10*time.Second, 300*time.Second, func() (bool, error) {
```
yifan-gu: make consts here instead of hardcode?

yifan-gu: Also line 143

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +138,36 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
+			return false, err
 		}
-		if len(m.Members) != 1 {
-			glog.Infof("Still migrating boot etcd member, retrying...")
-			return false, nil
+		switch e.Status.Phase {
+		case spec.ClusterPhaseRunning:
+			return true, nil
+		case spec.ClusterPhaseFailed:
+			return false, errors.New("failed to create etcd cluster")
```
yifan-gu: You can also use fmt.Errorf() if you want to be consistent, but you can leave as is.

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +138,36 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
-	// - Then we list members and see if it's been reduced to 1 member. Because when we
-	//   can talk to the etcd cluster, we are certain there are 2 members at the beginning,
-	//   and will reduce to 1 eventually. That's the timeline of expected events.
-	//   As long as 1 member cluster is reached, we are certain cluster has been migrated successfully.
-	err := wait.PollImmediate(10*time.Second, 60*time.Second, func() (bool, error) {
-		cfg := clientv3.Config{
-			Endpoints:   []string{fmt.Sprintf("http://%s:2379", etcdServiceIP)},
-			DialTimeout: 5 * time.Second,
+	err := wait.Poll(10*time.Second, 300*time.Second, func() (bool, error) {
+		uri := fmt.Sprintf("%s/apis/coreos.com/v1/namespaces/kube-system/etcdclusters/%s", host, etcdClusterName)
```
yifan-gu: better to make const for this or have a function to create the uri.

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +138,36 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
-			return false, nil
+		switch e.Status.Phase {
+		case spec.ClusterPhaseRunning:
+			return true, nil
+		case spec.ClusterPhaseFailed:
+			return false, errors.New("failed to create etcd cluster")
 		}
-		glog.Infof("etcd cluster is up. Member: %v", m.Members[0].Name)
-		return true, nil
+		return false, nil
```
yifan-gu: what can be the other status? is this an error? maybe return something like fmt.Errorf("unknown error type")?

yifan-gu: If we don't want to treat as an error, we can still print a warning here to catch this unexpected result

yifan-gu:
@hongchaodeng A couple nits.

pkg/util/etcdutil/migrate.go
aaronlevy: Still need to have this name changed (kube-etcd)

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +140,50 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
+	err := wait.Poll(10*time.Second, waitEtcdClusterRunningTime, func() (bool, error) {
+		res := restclient.Get().RequestURI(makeEtcdClusterURI(host, etcdClusterName)).Do()
+		if res.Error() != nil {
+			return false, res.Error()
 		}
-		etcdcli, err := clientv3.New(cfg)
-		if err != nil {
-			glog.Errorf("fail to create etcd client, retrying...: %v", err)
-			return false, nil
+		fmt.Println("haha")
```
aaronlevy: ?

aaronlevy:
Started a review, but I think this is still being tested. Ping me when I should take another look

pkg/util/etcdutil/migrate.go
hongchaodeng: The CI failed. I'm trying to print out something. But the CI didn't show anything.

hongchaodeng: @aaronlevy Do you know where I can get the bootkube logs from the CI?

aaronlevy: I don't think you can right now. @pbx0 is looking into general failures right now -- it may be unrelated to this PR

pbx0:
retest this please

hongchaodeng:
@aaronlevy 
Ready for review

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +140,41 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
-	//   Once we can talk to it, we are sure that etcd cluster is created.
-	// - Then we list members and see if it's been reduced to 1 member. Because when we
-	//   can talk to the etcd cluster, we are certain there are 2 members at the beginning,
-	//   and will reduce to 1 eventually. That's the timeline of expected events.
-	//   As long as 1 member cluster is reached, we are certain cluster has been migrated successfully.
-	err := wait.PollImmediate(10*time.Second, 60*time.Second, func() (bool, error) {
-		cfg := clientv3.Config{
-			Endpoints:   []string{fmt.Sprintf("http://%s:2379", etcdServiceIP)},
-			DialTimeout: 5 * time.Second,
+	err := wait.Poll(10*time.Second, waitEtcdClusterRunningTime, func() (bool, error) {
```
xiang90: change this to 5 minute per aaron's request?

hongchaodeng: Yes, it's 300s

xiang90: awesome.

aaronlevy:
How is this being tested? Without additional changes - it doesn't seem like I could test this with any of the hack/* installation methods  or existing tests.

xiang90:
@aaronlevy We need to ensure the etcd TPR status is running after the destructive tests. I think either @pbx0 or @hongchaodeng can add that checking bit.

aaronlevy:
@xiang90 @hongchaodeng As this PR stands, this code won't even be exercised at all (as far as I can tell) -- like none of the installation methods or tests use it -- so how is it being verified manually?

For reference, these were the changes I had to make last time I was testing to get it to launch a cluster at all: https://github.com/kubernetes-incubator/bootkube/compare/master...aaronlevy:sh-etcd

- `--etcd-quorum-read=true` on bootkube and self-hosted apiserver (https://github.com/kubernetes-incubator/bootkube/commit/f43a1ba02976314e83ed05b2bb581702a0c8ba51#diff-1a342de84d1bbb83d110cdca78afa143R144 & https://github.com/kubernetes-incubator/bootkube/commit/f43a1ba02976314e83ed05b2bb581702a0c8ba51#diff-b2cbe19f6b4883d7cdc511917ceccad1R92)
- modifying `bootkube render` and `bootkube start` flags (https://github.com/kubernetes-incubator/bootkube/commit/f43a1ba02976314e83ed05b2bb581702a0c8ba51#diff-4999be0b076d54f1ccb80eb692d76e9e) then also changing the status check to just use first endpoint (https://github.com/kubernetes-incubator/bootkube/compare/master...aaronlevy:sh-etcd#diff-4b418a1b1ee3602f77393018ab019352R27
- May not be an issue anymore - but I had to explicitly ignore notFound errors: https://github.com/kubernetes-incubator/bootkube/compare/master...aaronlevy:sh-etcd#diff-2d3c8cde0a58be225d2f6d6b5d64059fR68

aaronlevy:
@xiang90 @hongchaodeng I added a PR against this branch:

https://github.com/kubernetes-incubator/bootkube/pull/265

That should help with testing against local multi-node installation. But as far as I got was this still needing: https://github.com/kubernetes-incubator/bootkube/compare/master...aaronlevy:sh-etcd#diff-4b418a1b1ee3602f77393018ab019352R27

Otherwise it fails because multiple etcd servers are specified during `bootkube start`

xiang90:
@aaronlevy With all your change merged in, I think we should expect self hosted etcd to work by specifying self-hosted flag to the muti node setup. You already got it works the other day. The issue what the operator failed to pick up the migrated cluster after a restart. It should be fixed now. So if we try to restart the node, everything should come up correctly assuming checkpoint works.

aaronlevy:
The checkpointing still won't work until we "checkpoint" the iptables rules too (can probably start with just enabling iptables-store.service / iptables-restore.service on host)

xiang90:
> May not be an issue anymore - but I had to explicitly ignore notFound errors: 

This still need to be fixed. Talking directly to TPR outside the operator programmatically is annoying. I am working on an etcd operator client. So we can hide TPR completely from operator user point of view.

Makefile
```diff
@@ -66,6 +67,10 @@ vendor:
 	@mv $@/k8s.io/kubernetes/vendor/* $(abspath $@)
 	@cd $@/k8s.io/ && ln -sf kubernetes/staging/src/k8s.io/client-go client-go
 	@rm -rf $@/k8s.io/kubernetes/vendor $@/k8s.io/kubernetes/.git
+	@mkdir -p $@/github.com/coreos/etcd-operator
```
xiang90: we should only vendor the spec not the entire operator repo.

pkg/bootkube/bootkube.go
```diff
@@ -45,10 +45,10 @@ type bootkube struct {
 
 func NewBootkube(config Config) (*bootkube, error) {
 	apiServer := apiserver.NewServerRunOptions()
+	apiServer.GenericServerRunOptions.EnableGarbageCollection = false
```
hongchaodeng: ignore this part. We will remove it at the end

xiang90: can we remove this now? this should not be in this pr.

pkg/bootkube/bootkube.go
```diff
@@ -60,6 +61,7 @@ func NewBootkube(config Config) (*bootkube, error) {
 		"--cluster-cidr=10.2.0.0/16",
 		"--configure-cloud-routes=false",
 		"--leader-elect=true",
+		"--enable-garbage-collector=false",
```
xiang90: can we remove this?

pkg/util/etcdutil/migrate.go
```diff
@@ -136,41 +144,44 @@ func createMigratedEtcdCluster(restclient restclient.Interface, host, podIP stri
...
 		}
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-		defer cancel()
-		m, err := etcdcli.MemberList(ctx)
-		if err != nil {
-			glog.Errorf("fail to list etcd members, retrying...: %v", err)
+
+		e := &spec.EtcdCluster{}
+		if err := json.Unmarshal(b, e); err != nil {
+			glog.Errorf("Unmarshal fail: %v", err)
```
xiang90: we can clean up all the error messages and debugging info in this function?

hongchaodeng:
Finally we have the fix PR ready and tested again hack/multi-node
- have self hosted etcd
- have scaled etcd cluster from 1 -> 2

Unachieved items:
- scale >2 because we only have two nodes
- reboot doesn't work (no ip tables checkpointing)

Here's what the fix is:
- Since we changed to rest.Interface, we need to check error now instead of status code.
- We need to wait cluster TPR status.phase to switch to "Running". Otherwise, stopping apiserver will cause the restore of the cluster to fail.
- We need to wait boot etcd to be removed. Otherwise, after apiserver restarts, restore of the cluster won't remove the boot member.

If all is successful, we shall see from bootkube logs:
```
I0112 22:18:23.357220    1454 migrate.go:160] etcd cluster status.phase: Running
I0112 22:18:23.357278    1454 migrate.go:62] etcd cluster RUNNING...
I0112 22:18:33.365627    1454 migrate.go:196] Still waiting boot etcd to be deletd...
...
E0112 22:18:58.928740    1454 errors.go:63] apiserver received an error that is not an unversioned.Status: rpc error: code = 13 desc = transport is closing
I0112 22:19:03.360928    1454 migrate.go:67] boot etcd deleted...
```
We see that apiserver erroring because boot etcd is removed. At that point, in bootkube, it will detect that as well independently of apiserver using etcd API and etcd service IP. Then bootkube shall exit successfully.

We shall see
```
$ kubectl --kubeconfig cluster/auth/kubeconfig -n kube-system get pod
kube-etcd-0000                             1/1       Running           0          6m
```
No "kube-etcd-$(IP)" for the boot member.

After scaling up, we shall see pods:
```
kube-etcd-0000                             1/1       Running           0          6m
kube-etcd-0001                             0/1       PodInitializing   0          9s
```
```
kube-etcd-0000                             1/1       Running   0          6m
kube-etcd-0001                             1/1       Running   0          11s
```

cluster TPR:
```
$ kubectl --kubeconfig cluster/auth/kubeconfig -n kube-system get etcdcluster kube-etcd -o yaml
...
spec:
  ...
  size: 2
```

hongchaodeng:
There is prerequisite for above comment that I need to mention:

- In our testing, we disabled GC in apiserver and controller manager. This is known k8s 1.5 regression issue.
- This depends on #270 #271

aaronlevy:
I'll try and take a closer look tonight, but just skimmed your message. FWIW you can launch as many nodes as you want: https://github.com/kubernetes-incubator/bootkube/blob/master/hack/multi-node/Vagrantfile#L12-L15

pkg/util/etcdutil/migrate.go
```diff
@@ -33,6 +43,7 @@ func Migrate() error {
 	if err != nil {
 		return err
 	}
+	glog.Infof("TPR is setup...")
```
xiang90: etcd cluster TPR is setup...

pkg/util/etcdutil/migrate.go
```diff
@@ -41,11 +52,20 @@ func Migrate() error {
 	glog.Infof("boot-etcd pod IP is: %s", ip)
 
 	if err := createMigratedEtcdCluster(restClient, apiserverAddr, ip); err != nil {
-		glog.Errorf("fail to create migrated etcd cluster: %v", err)
-		return err
+		return fmt.Errorf("fail to create migrated etcd cluster: %v", err)
+	}
+	glog.Infof("created etcd cluster TPR")
```
xiang90: etcd migration cluster is created.

pkg/util/etcdutil/migrate.go
```diff
@@ -41,11 +52,20 @@ func Migrate() error {
...
+
+	if err := waitEtcdClusterRunning(restClient); err != nil {
+		return fmt.Errorf("wait etcd cluster running failed: %v", err)
 	}
+	glog.Info("etcd cluster RUNNING...")
```
xiang90: etcd migration cluster is now running.

pkg/util/etcdutil/migrate.go
```diff
@@ -41,11 +52,20 @@ func Migrate() error {
...
 
-	return checkEtcdClusterUp()
+	if err := waitBootEtcdDeleted(); err != nil {
+		return fmt.Errorf("wait boot etcd deleted failed: %v", err)
+	}
+	glog.Info("boot etcd deleted...")
```
xiang90: the boot etcd is removed from the migration cluster.

hongchaodeng:
cleaned up

xiang90:
@hongchaodeng Can you rebase the code with master?

@aaronlevy @pbx0 and I tested this with CI, it works. Can you give this a review and hopefully get this merged? Thank you!

aaronlevy:
Overall these changes lgtm

@xiang90 @hongchaodeng - one thing I'd like to see changed (which caused me issues when testing in the past):

Add an explicit tag to the etcd-operator manifest(https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L516) and then not modifying the image for that version tag after release.

aaronlevy:
Can the conflicts also be resolved (sorry about that -- bumped the k8s version)

hongchaodeng:
@aaronlevy 
Rebased.

We pushed a special tag 'bootkube' in etcd operator quay image for bootkube right now for stability purpose. After releasing beta of etcd operator image, we would change the tag to
that.

aaronlevy:
Is there a particular reason you want to use a "bootkube" tag rather than maybe the git hash? I have would have no good way of mapping a particular etcd-operator image back to the repo to debug

hongchaodeng:
@aaronlevy 
Pushed a new tag based on current etcd operator master's git hash.

aaronlevy:
When I test this locally, I get a running cluster but only see the boot-etcd pod (no self hosted pod):

```
aaron@aa:$ kubectl get pods --all-namespaces
kube-system   boot-etcd-172.17.4.101                     1/1       Running   0          2m
kube-system   checkpoint-installer-ftmw5                 1/1       Running   0          5m
kube-system   kube-apiserver-svrvp                       1/1       Running   2          5m
kube-system   kube-controller-manager-3654003418-kb6v8   1/1       Running   0          5m
kube-system   kube-controller-manager-3654003418-rsq0d   1/1       Running   0          5m
kube-system   kube-dns-4101612645-lr026                  4/4       Running   0          5m
kube-system   kube-flannel-9dclv                         2/2       Running   1          5m
kube-system   kube-flannel-gp2md                         2/2       Running   3          4m
kube-system   kube-proxy-l4bhq                           1/1       Running   0          5m
kube-system   kube-proxy-pkw0g                           1/1       Running   0          4m
kube-system   kube-scheduler-850181384-1bv9p             1/1       Running   0          5m
kube-system   kube-scheduler-850181384-rtl5m             1/1       Running   0          5m
kube-system   pod-checkpointer-172.17.4.101              1/1       Running   0          4m
```

aaronlevy:
Actually, ignore my previous comment this likely was my fault - testing again

xiang90:
> When I test this locally, I get a running cluster but only see the boot-etcd pod (no self hosted pod):

etcd operator is not even started. How did you test it?

xiang90:
> Actually, ignore my previous comment this likely was my fault - testing again

ah. ok. i was too fast.

PR #261, opened 2017-01-06
Enable TLS Bootstrapping / RBAC

Quentin-M:
This is an attempt at enabling TLS bootstrapping / RBAC.

```
=== all: Enable TLS Bootstrapping
Instead of Bootkube generating a certificate for kubelets
to authenticate against kube-api, each kubelet submits a
its own CSR to the kube-api, which is then signed by the
kube-controller-manager once approved.

Bootkube however still generates a certificate for the
administrator. Additionaly, an auth token is generated
by Bootkube and distributed to the kubelets. This token
mainly acts as a first gate preventing everyone from
submitting CSRs. The security solely depends on
the CSR approval flow adopted.

Bootkube automatically approves any pending CSRs during
the bootstrapping phase.

Ref: http://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/
```

```
=== all: Enable RBAC authorization modes
In order to improve overall security and constrain
the bootstrap token to submitting CSRs, this commit
enables both RBAC and introduces default RBAC bindings.
```


Quentin-M:
cc @aaronlevy @ericchiang @dghubble

pbx0:
Note that because of changes in bootkube setup I will need to update the tests for this PR so its safe to ignore bootkube-dev until I test it offline and incorporate these changes.

pkg/asset/internal/templates.go
```diff
@@ -172,6 +183,18 @@ spec:
...
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {"user": "admin",                                                                          "namespace": "*", "resource": "*",                          "apiGroup": "*"                                     }}
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {"user": "kubelet-bootstrap",                         "group": "system:kubelet-bootstrap", "namespace": "*", "resource": "certificatesigningrequests", "apiGroup": "certificates.k8s.io"                   }}
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {"user": "system:serviceaccount:kube-system:default",                                      "namespace": "*", "resource":"*", "nonResourcePath": "*",   "apiGroup":"*"                                      }}
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {                                                     "group": "system:masters",           "namespace": "*", "resource":"*", "nonResourcePath": "*",   "apiGroup":"*"                                      }}
```
Quentin-M: Must split into two rules: https://github.com/kubernetes/kubernetes/pull/29930

pkg/asset/internal/templates.go
```diff
@@ -172,6 +183,18 @@ spec:
...
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {                                                     "group": "system:unauthenticated",                     "nonResourcePath": "*",                                                      "readonly": true }}
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {"user": "admin",                                                                          "namespace": "*", "resource": "*",                          "apiGroup": "*"                                     }}
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {"user": "kubelet-bootstrap",                         "group": "system:kubelet-bootstrap", "namespace": "*", "resource": "certificatesigningrequests", "apiGroup": "certificates.k8s.io"                   }}
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {"user": "system:serviceaccount:kube-system:default",                                      "namespace": "*", "resource":"*", "nonResourcePath": "*",   "apiGroup":"*"                                      }}
```
Quentin-M: Must split into two rules: https://github.com/kubernetes/kubernetes/pull/29930

ericchiang:
why is the first commit here?

Quentin-M:
@ericchiang Just a mistake from my side during the latest rebase that I did in a rush to show you what's currently there. Thanks for the feedback you given me offline.

Quentin-M:
@aaronlevy @ericchiang Updated. All three `hack/` examples are working.

pkg/bootkube/bootkube.go
```diff
@@ -88,6 +90,10 @@ func makeAPIServerFlags(config Config) []string {
 		"--tls-private-key-file=" + filepath.Join(config.AssetDir, asset.AssetPathAPIServerKey),
 		"--tls-cert-file=" + filepath.Join(config.AssetDir, asset.AssetPathAPIServerCert),
 		"--client-ca-file=" + filepath.Join(config.AssetDir, asset.AssetPathCACert),
+		"--token-auth-file=" + filepath.Join(config.AssetDir, asset.AssetPathBootstrapAuthToken),
+		"--authorization-mode=ABAC,RBAC",
+		"--authorization-policy-file=" + filepath.Join(config.AssetDir, asset.AssetPathAPIServerAuthorizationPolicy),
+		"--authorization-rbac-super-user=admin",
```
ericchiang: Let's actually drop the use of this flag. It's being deprecated later. Uses of this should just go in the `system:masters` group.

The bootstrapping policies are here. They're a little hard to read (later releases check in yaml versions), but they summarize the idea.

https://github.com/kubernetes/kubernetes/blob/v1.5.2/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
https://github.com/kubernetes/kubernetes/blob/v1.5.2/pkg/auth/user/user.go#L69-L78

ericchiang: The "Organization" of a TLS cert can be used to place a client cert in a group btw.

https://kubernetes.io/docs/admin/authentication/#x509-client-certs

pkg/asset/tls.go
```diff
@@ -98,13 +109,13 @@ func newAPIKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey, altNa
...
-func newKubeletKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey) (*rsa.PrivateKey, *x509.Certificate, error) {
+func newAdminKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey) (*rsa.PrivateKey, *x509.Certificate, error) {
 	key, err := tlsutil.NewPrivateKey()
 	if err != nil {
 		return nil, nil, err
 	}
 	config := tlsutil.CertConfig{
-		CommonName:   "kubelet",
+		CommonName:   "admin",
 		Organization: []string{"kube-node"},
```
ericchiang: Add `system:nodes` here for the kubelet.

pkg/asset/internal/templates.go
```diff
@@ -25,6 +25,21 @@ contexts:
     user: {{ .UserName }}
 `)
 
+	KubeSystemSARoleBindingTemplate = []byte(`apiVersion: rbac.authorization.k8s.io/v1alpha1
```
ericchiang: how are these getting into the API?

pkg/asset/tls.go
Quentin-M: This `kubeconfig` is not used by kubelet at all. This kubeconfig is only used by the administrator of the cluster (i.e. to use `kubectl`). Does it still makes sense I add the `system:nodes` organization?

pkg/asset/internal/templates.go
Quentin-M: They are added to the default static assets in https://github.com/kubernetes-incubator/bootkube/pull/261/files#diff-4475804b8fcae522e91ffb4651f8f967R30, so are generated by the `render` command (`NewDefaultAssets`) then created by the `start` command (`CreateAssets`).

pkg/asset/tls.go
Quentin-M: Should it be `system:masters` instead?

Quentin-M: Is `kube-node` required? Otherwise we could drop it.

Also, right above is the cert generation for the API Server. It has the `kube-master` group. I feel like it should be renamed too.

hack/quickstart/init-master.sh
```diff
@@ -70,31 +70,36 @@ function init_master_node() {
...
 # After assets are available on the remote host, the script will execute itself in "local" mode.
 if [ "${REMOTE_HOST}" != "local" ]; then
     # Set up the kubelet.service on remote host
     scp -i ${IDENT} -P ${REMOTE_PORT} ${SSH_OPTS} kubelet.master core@${REMOTE_HOST}:/home/core/kubelet.master
     ssh -i ${IDENT} -p ${REMOTE_PORT} ${SSH_OPTS} core@${REMOTE_HOST} "sudo mv /home/core/kubelet.master /etc/systemd/system/kubelet.service"
 
-    # Copy self to remote host so script can be executed in "local" mode
+    # Copy the auto approver to remote host and execute it for 1h
+    scp -i ${IDENT} -P ${REMOTE_PORT} ${SSH_OPTS} ../auto_approver.sh core@${REMOTE_HOST}:/home/core/auto_approver.sh
+    ssh -i ${IDENT} -p ${REMOTE_PORT} ${SSH_OPTS} core@${REMOTE_HOST} "/home/core/auto_approver.sh 3600" &
```
aaronlevy: Why not have bootkube auto-approve the first node which tries to join? In case people didn't want to use the tls bootstrap -- you could have it auto-sign or wait for a node to join. Whichever comes first.

I'm a little wary of relying on more external components for the bootstrap process to work (especially a random backgrounded shell process).

If the intention is that we allow nodes to join for a longer period - I think it might be better to let bootkube start it's process after first node has joined (and it auto-approves), but once an api-server is available we deploy a pod/job/deployment for an autoapprover (not really sure what our desired functionality is here).

hack/quickstart/quickstart-aws.md
```diff
@@ -60,20 +60,24 @@ $ IDENT=k8s-key.pem ./init-master.sh <PUBLIC_IP>
...
 
 ```
-IDENT=k8s-key.pem ./init-worker.sh <PUBLIC_IP> cluster/auth/kubeconfig
+IDENT=k8s-key.pem ./init-worker.sh <PUBLIC_IP>
 ```
 
-**NOTE:** It can take a few minutes for each node to download all of the required assets / containers.
+**NOTE:**
+- If the master has been bootstrapped over an hour ago, each worker will have
+ to be [approved](../auto_approver.sh) manually.
```
aaronlevy: Can we give a concrete example of how an admin would do this? Not sure linking to the auto approver is immediately helpful.

image/checkpoint/build-image.sh
```diff
@@ -14,7 +14,7 @@ function image::build() {
     cp ${BOOTKUBE_ROOT}/image/checkpoint/Dockerfile ${TEMP_DIR}
     cp ${BOOTKUBE_ROOT}/image/checkpoint/checkpoint-install.sh ${TEMP_DIR}
     cp ${BOOTKUBE_ROOT}/image/checkpoint/checkpoint-pod.yaml ${TEMP_DIR}
-    sed -i "s#{{ REPO }}:{{ TAG }}#${IMAGE_REPO}:${VERSION}#" ${TEMP_DIR}/checkpoint-pod.yaml
+    sed -i.bk "s#{{ REPO }}:{{ TAG }}#${IMAGE_REPO}:${VERSION}#" ${TEMP_DIR}/checkpoint-pod.yaml
```
aaronlevy: Is there a particular reason you want to keep a backup?

pkg/asset/asset.go
```diff
@@ -13,45 +13,50 @@ import (
...
+	AssetPathKubeFlannel                 = "manifests/kube-flannel.yaml"
+	AssetPathKubeFlannelCfg              = "manifests/kube-flannel-cfg.yaml"
+	AssetPathAPIServerSecret             = "manifests/kube-apiserver-secret.yaml"
+	AssetPathAPIServer                   = "manifests/kube-apiserver.yaml"
+	AssetPathControllerManager           = "manifests/kube-controller-manager.yaml"
+	AssetPathControllerManagerSecret     = "manifests/kube-controller-manager-secret.yaml"
+	AssetPathScheduler                   = "manifests/kube-scheduler.yaml"
+	AssetPathKubeDNSDeployment           = "manifests/kube-dns-deployment.yaml"
+	AssetPathKubeDNSSvc                  = "manifests/kube-dns-svc.yaml"
+	AssetPathSystemNamespace             = "manifests/kube-system-ns.yaml"
```
aaronlevy: nit: I know not your change, but I don't believe this is used anymore if you wanted to drop it.

hack/quickstart/init-worker.sh
```diff
@@ -13,57 +13,41 @@ function usage() {
     exit 1
 }
 
-function extract_master_endpoint (){
```
aaronlevy: How is the master location determined now?

pkg/asset/k8s.go
```diff
@@ -99,6 +138,7 @@ func newControllerManagerSecretAsset(assets Assets) (Asset, error) {
 	secretAssets := []string{
 		AssetPathServiceAccountPrivKey,
 		AssetPathCACert, //TODO(aaron): do we want this also distributed as secret? or expect available on host?
+		AssetPathCAKey,  // ^
```
aaronlevy: I'm not really sure what that TODO means -- these are being distributed as secrets...

hack/quickstart/init-master.sh
```diff
@@ -7,8 +7,8 @@ CLUSTER_DIR=${CLUSTER_DIR:-cluster}
 IDENT=${IDENT:-${HOME}/.ssh/id_rsa}
 SSH_OPTS=${SSH_OPTS:-}
 
-BOOTKUBE_REPO=${BOOTKUBE_REPO:-quay.io/coreos/bootkube}
-BOOTKUBE_VERSION=${BOOTKUBE_VERSION:-v0.3.1}
+BOOTKUBE_REPO=${BOOTKUBE_REPO:-quay.io/quentin_m/bootkube}
+BOOTKUBE_VERSION=${BOOTKUBE_VERSION:-bootstrap_tls}
```
Quentin-M: Yeap, thanks. I didn't mean to include that and probably forgot it during the last rebase.

image/checkpoint/build-image.sh
Quentin-M: This is literally so it can [work](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/sed.1.html) on Darwin. I can omit that change and keep it to myself.

pkg/asset/asset.go
Quentin-M: Yeap this isn't used, Goglang complains about that dead constant.

hack/quickstart/init-worker.sh
Quentin-M: The generated (bootstrap-)kubeconfig already contains the Private IP of the API server. We don't need to do anything else now that we are removing the `--api-servers` parameter: Kubelet gets it from kubeconfig.

pkg/asset/k8s.go
Quentin-M: You meant that maybe we should distribute these through another channel.

image/checkpoint/build-image.sh
aaronlevy: ah. You could do something like

```
if darwin
sedcmd='sed -i ""'
else
sedcmd='sed -i'
```

hack/quickstart/quickstart-aws.md
Quentin-M: I may either link https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/ (look at the last section) or https://github.com/gtank/csrctl, whichever you think is more meaningful.

hack/quickstart/init-master.sh
Quentin-M: Not sure if I get what you just suggested?

I believe we should not have an auto-approver by default in Bootkube elsewhere than in the `hack/` examples. Production users should have control over which nodes are allowed to join, and be able to define their own approval policy, regardless if it is the first or the 10th node.

We could however simply have that script as a systemd unit on masters? Or maybe bootkube should have an integrated command to manage approvals (list, approve, reject, auto-approve).. since kubectl doesn't have one.

aaronlevy: > regardless if it is the first or the 10th node

We may have to special-case at least the first node (e.g. bootkube auto-approves first node). Otherwise nothing starts (no node is available for scheduling, so no pods run).

Another option here is when switching to using static pods for the bootstrap control plane (https://github.com/kubernetes-incubator/bootkube/issues/168) - One other pod could be the initial approver (e.g. the approval mechanism is the ability to prove that it's running on the same node).

> I believe we should not have an auto-approver by default in Bootkube elsewhere than in the hack/ examples.

I agree -- but even in those examples I want to be moving away from on-host configuration (which this script ultimately is). Running it as a backgrounded bash job also makes this more difficult to debug.

> We could however simply have that script as a systemd unit on masters?

Same issue - we should move away from on-host configuration (e.g. no new systemd units unless unavoidable). Maybe we change bootkube to auto approve X nodes, or all nodes while running, or? Past that point I think we either need to provide example approvers which can run as pods in the cluster (maybe just an example "approve all") & documentation (curl?) on how to approve new nodes manually.

> Or maybe bootkube should have an integrated command to manage approvals

I want to keep bootkube as a "run only once" tool, after bootstrap we should only require using pure upstream tooling. In a pinch, maybe curl + good docs (and putting effort into getting necessary features into kubeadm).

hack/quickstart/init-worker.sh
aaronlevy: ack

hack/quickstart/quickstart-aws.md
aaronlevy: Probably both with examples of using gtank's csrctl

ericchiang:
@aaronlevy I discussed this a bit out in person with @Quentin-M.

Proposed progression of this feature:

* For this PR, bootkube auto-approves everything (in it's Go code) as it bootstraps. It also creates an auto-approver pod that runs for some period of time after bootkube exits (e.g. an hour).
* As https://github.com/kubernetes-incubator/bootkube/issues/168 (temp control-plane as static pods) gets implemented, that auto-approver replaces the Go code as a static pod/deployment. May need some sort of bootstrapping kubelet to run the static pods because kubelets blocks on TLS bootstrapping before launching static pods.
* Users can later customize their approval behavior by replacing the auto-approver in the temporary control-plane.

aaronlevy:
@ericchiang @Quentin-M 

In general sgtm - but still wondering about a couple things:

- What does the 1hr auto-approver get us / solve? Seems somewhat unnecessary. Auto-approve all nodes during bootstrap (in go code), then documenting how to approve additional nodes later seems like it should cover most bootstrapping scenarios initially (e.g. launch cloud-formation with 4 nodes, and want them all initially part of the cluster).

- Not sure the complexity of moving this "bootstrap" approving into a static pod will get us. Bootkube is still going to be sitting there running (inspecting state via api) even if we use static pods (because bootkube will eventually need to remove them). So seems reasonable to leave the "auto-approve during bootstrap" code in bootkube (but this is just my initial hunch).

- As far as user customization, agree - I'd like to see some generally useful examples (and this may occur upstream) like AWS/GCE auto-approvers (which use metadata/ IAM / etc). But we can always provide an "Auto-approver" pod as an example (which we don't really recommend using).

So mostly I would think the move-forward steps are to just add "auto-approve while running" into bootkube's code...

Quentin-M:
SGTM. Thanks for the input.

My main concern regarding to embedding an auto-approver into bootkube itself is that every nodes/kubelet might not be up during the limited life duration of bootkube, even during the automated QuickStart examples, which could then be troublesome. Users might end up having to manually approve and automated processes would have to run their own approver anyways.

The second reason is that I didn't want a default "auto-approve all" running at any given time outside of the examples - and instead give full control to the end-user (i.e a malicious/unauthorized node sneaks in during the bootstrapping process). However one may consider the auth bootstrap token as secure as the original kubeconfig - not cryptographically speaking but because they are distributed in the same manner. Within this assumption, the second reason becomes less relevant.

ericchiang:
> Not sure the complexity of moving this "bootstrap" approving into a static pod will get us. Bootkube is still going to be sitting there running (inspecting state via api) even if we use static pods (because bootkube will eventually need to remove them). So seems reasonable to leave the "auto-approve during bootstrap" code in bootkube (but this is just my initial hunch).

I think getting the approval logic out of bootkube helps us focus on the plug-ability of this feature. You'll always need something in case a worker tries to join after bootkube exits. If we have to write that component, I'd rather just have it do all of the bootstrapping, and ensure bootkube works when something else is admitting the initial nodes.

Quentin-M:
> I think getting the approval logic out of bootkube helps us focus on the plug-ability of this feature. You'll always need something in case a worker tries to join after bootkube exits. If we have to write that component, I'd rather just have it do all of the bootstrapping, and ensure bootkube works when something else is admitting the initial nodes.

That was exactly my point in the code comments above.

aaronlevy:
My concern with that approach is that you now must coordinate another tool on the host during bootstrap. Running `bootkube start` will not result in a working cluster. You must also run some kind of approval process (it can't be started as a static pod, because there is a chicken/egg issue).

This might not be a complete deal-breaker, but it does seem rather unfortunate that the tool no longer works in isolation...

I'm also not advocating that this same code isn't put into its own pod (e.g. example "approval all pod") that can run in-cluster, it's just that if we do not put similar/same code in bootkube we now have 2 required moving parts -- and I think that is worse than maybe 20 lines of code being duplicated. The systemd unit to run the temporary approver will probably be longer than the code itself.

As far as the concern that all initial nodes won't be available during bootstrap phase -- I think running for an hour is just as arbitrary (and less user-friendly). After bootstrap - the user should decide how they want to approve, or have deployed an approver pod as part of the bootstrap process itself. If you're really set on the hour window, for example: I would suggest adding "approve first node" code to bootkube, then deploying a job in the cluster which approves nodes for an hour.

ericchiang:
Yeah I think we all agree that another process should take over admitting nodes after bootkube exits. Where I think we're differing is what admits nodes during `bootkube start`. Or at least, form our side, what's the value of having something other than bootkube admin nodes during the bring-up.

Until https://github.com/kubernetes-incubator/bootkube/issues/168 is resolved though, the initial admitter logic has to be bootkube. So let's just have bootkube admit all for now and discuss customization once #168 is in.

Sound reasonable?

aaronlevy:
Sure -- but to be fair I don't see how https://github.com/kubernetes-incubator/bootkube/issues/168 changes anything. You still have a chicken / egg problem and bootkube will have to approve at least the first node that tries to join - otherwise no static pods will start... unless I'm missing something.

The other option being that you're running a separate approver binary (not as a static pod) -- and I just don't see any benefit to that over letting bootkube approve at least first node.

aaronlevy:
Also @ericchiang what are "bootkube admin nodes"?

aaronlevy:
In an ideal future I'd like bootkube to only approve only the first node, and one of the manifests it then deploys is an "approver" application (e.g. an AWS based approver) - which takes over after that. But we don't have any development work towards that yet (that I'm aware of).

ericchiang:
> Also @ericchiang what are "bootkube admin nodes"?

Sorry I meant "admit"

> Sure -- but to be fair I don't see how #168 changes anything. You still have a chicken / egg problem and bootkube will have to approve at least the first node that tries to join - otherwise no static pods will start... unless I'm missing something.

There's going to be a chicken and egg problem that kubelets need to bootstrap before they'll run static pods anyway.

> In an ideal future I'd like bootkube to only approve only the first node, and one of the manifests it then deploys is an "approver" application (e.g. an AWS based approver) - which takes over after that. But we don't have any development work towards that yet (that I'm aware of).

Yeah I think this is fine.

aaronlevy:
So for clarity I still don't understand:

> Until #168 is resolved though, the initial admitter logic has to be bootkube. So let's just have bootkube admit all for now and discuss customization once #168 is in.

What does #168 change? The initial admit logic has to live somewhere (and not in a static pod -- because it will never start).

ericchiang:
> What does #168 change? The initial admit logic has to live somewhere (and not in a static pod -- because it will never start).

If we wanted to admit other nodes while the temporary control plane was running using logic that wasn't admit all. Maybe I'm just misunderstanding the steps though.

aaronlevy:
Ah k. It wouldn't need to be a static pod -- just a normal deployment will do -- not many reasons we can't start to run normal workloads during bootstrap (we do this already for dns, heapster, even the control plane components themselves).

- Bootkube starts (static control plane or not)
- Bootkube injects all manifests into api (deployments, daemonsets, services, etc.)
    - One of those manifests is a deployment for "custom approver app"
- Bootkube approves first node (or maybe only "self" node)
- Pods start ending up on that node -- one of which is "custom approver app"
- "custom approver app" pod is now responsible for approving csrs.

aaronlevy:
But **something** has to at least approve the first node so we can start workloads. I'd prefer for that something to not be a random binary/systemd unit on the host, and it can't be a static pod due to chicken/egg

Quentin-M:
@aaronlevy The idea that @ericchiang had as a long-term plan is that the kubelet would still be able to run static pods even tho it hasn't been bootstrapped yet, in a world where bootkube would simply be responsible for moving around static manifests for that kubelet - including an approver (instead of starting the temporary control plan itself).

aaronlevy:
I still see very little (if any) benefit to using a static pod. As long as you require an external coordinator to shuffle the static pods around on disk -- you may as well have that external coordinator approve the first node (itself). The static pod is just another random piece you have to delete then re-create as a real object. And starting the pod sync loop before the bootstrap phase seems like a lot of work in the kubelet with basically no benefit (but increased complexity).

Quentin-M:
I updated the PR. Notable changes since the last revision are:
- Embedded an auto-approver into bootkube, that is ran during `bootkube start`
- Documented approving manually a node
- Tweaked the TLS Assets generation slightly to fit the default User/Group that are defined by the RBAC Cluster Roles Bindings [created](https://github.com/deads2k/kubernetes/blob/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go) by Kubernetes itself.

I did not include the creation of an approver pod asset here as we have to create an approver tool independently in the near future, that might be open-source and generic enough to fit in here. TBD.

ericchiang:
rbac stuff lgtm

pkg/asset/tls.go
```diff
@@ -98,18 +109,55 @@ func newAPIKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey, altNa
...
+		return nil, err
+	}
+
+	token := fmt.Sprintf("%s,kubelet-bootstrap,10001,\"system:kubelet-bootstrap\"\n", bearer)
+	return []byte(token), nil
+}
+
+// getBootstrapAuthTokenFromFile retrieves the first bearer token from the
+// given data of a static token file.
+func getBootstrapAuthTokenFromFile(data []byte) (string, error) {
```
aaronlevy: This doesn't actually get anything from a file. Maybe `parseBootstrapAuthToken`?

pkg/asset/tls.go
```diff
@@ -98,18 +109,55 @@ func newAPIKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey, altNa
...
 	return key, cert, err
 }
+
+// newBootstrapAuthTokenFile creates a static token file, containing a single
+// randomly-generated bearer token mapped to the user and group
+// 'kubelet-bootstrap'.
+//
+// This token file is meant to be used by kubelet to bootstrap its TLS
+// certificate as described on k8s.io/docs/admin/kubelet-tls-bootstrapping.
+func newBootstrapAuthTokenFile() ([]byte, error) {
```
aaronlevy: Doesn't actually create a file. Maybe just `newBootstrapAuthToken`?

pkg/bootkube/approve.go
```diff
@@ -0,0 +1,88 @@
...
+		Type:    certificates.CertificateApproved,
+		Reason:  "AutoApproved",
+		Message: "Auto approving of all kubelet CSRs is enabled on bootkube",
+	}
+
+	for {
+		request.Status.Conditions = append(request.Status.Conditions, condition)
+
+		if _, err := client.UpdateApproval(request); err != nil {
+			if strings.Contains(err.Error(), "the object has been modified") {
```
aaronlevy: I think you can use `isConflict(error)` here from` pkg/api/errors`

pkg/asset/tls.go
Quentin-M: I consciously named this function with the `File` prefix to make a distinction between a simple token / bearer (e.g. _487afc487_ ) and something that follows the token file format defined by Kubernetes' `plugin/pkg/auth/authenticator/token/tokenfile/...`. Renaming and making clear that we follow that format in the comment.

pkg/bootkube/approve.go
Quentin-M: I used it initially. doesn't work. The returned `err` doesn't implement the `APIStatus` interface and thus, `reasonForError` returns `metav1.StatusReasonUnknown` rather than `metav1.StatusReasonConflict`, therefore `isConflict(err)` returns `false`.

So I literally inspired myself from `test/e2e/kubectl.go:runKubectlRetryOrDie` and simply tested `strings.Contains()`. I could compare to the `OptimisticLockErrorMsg` constant but was afraid that the message could slightly differ one day for one or another reason if they ever modify it.

hack/quickstart/quickstart-aws.md
```diff
@@ -60,20 +60,47 @@ $ IDENT=k8s-key.pem ./init-master.sh <PUBLIC_IP>
...
 
 ```
-$ kubectl --kubeconfig=cluster/auth/kubeconfig get nodes
+# Find the CSR to approve.
+$ kubectl --kubeconfig=cluster/auth/admin-kubeconfig get csr
+NAME        AGE       REQUESTOR           CONDITION
+csr-46rdv   12m       kubelet-bootstrap   Pending
+
+# Proxy the Kubernetes API to localhost.
+$ kubectl proxy --port=8080 &
```
Quentin-M: Will propose an easy-to-use script on Monday!

Quentin-M:
Updated. I included a `csrctl.sh` script, that parses the provided kubeconfig and enables listing, getting, approving or denying CSRs.

f0:
Hi,

just FYI, i added this patch on top of the current master. 
Bootstrapped successfully a cluster with RBAC, works really good :+1: 

Thanks,
Florian

derekparker:
@Quentin-M this patch needs a rebase, but after that @pbx0 feel free to merge (I'll likely be offline by the time you see this, on Belgium time right now).

Quentin-M:
Thanks! Just rebased and re-tested. The etcd3 and disruption budgets PRs that got merged this morning created a conflict. 

/cc @pbx0

hack/single-node/user-data.sample
```diff
@@ -26,7 +26,10 @@ coreos:
         ExecStartPre=/bin/mkdir -p /var/lib/cni
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --kubeconfig=/etc/kubernetes/kubeconfig \
+          --experimental-bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubeconfig \
+          --cert-dir=/etc/kubernetes/secrets \
           --require-kubeconfig \
+          --cert-dir=/etc/kubernetes/secrets \
```
pbx0: duplicated flag

Quentin-M: Fixed~

pkg/bootkube/approve.go
```diff
@@ -0,0 +1,88 @@
...
+		v1.NamespaceAll,
+		fields.Everything(),
+	)
+
+	_, controller := cache.NewInformer(
+		watchList,
+		&certificates.CertificateSigningRequest{},
+		time.Second*5,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
```
aaronlevy: Nit: Do you think there is a possibility that a CSR has already been submitted before this watch is implemented? If that's the case, we may miss the add message(s). An option here would be to also use the UpdateFunc, which will fire for every CSR every informer resync period (5 seconds). Not a blocker, just wanted to call out the potential race.

pkg/bootkube/approve.go
aaronlevy: ack

pkg/bootkube/approve.go
Quentin-M: Sounds unlikely but worth doing! Good catch. I added a new commit and verified that:
1) It validates CSRs that were submitted before the watcher started.
2) It doesn't validate CSRs that have already been approved or denied.

Quentin-M:
test this

pkg/bootkube/approve.go
```diff
@@ -0,0 +1,102 @@
...
+				}
+			},
+		},
+	)
+
+	go controller.Run(wait.NeverStop)
+	return <-errCh
+}
+
+func approveCSR(client v1alpha1.CertificateSigningRequestInterface, request *certificates.CertificateSigningRequest, errCh chan error) {
```
liggitt: other things can submit CSRs... if the intent of this is to approve kubelet certs, it should verify it was submitted by the kubelet bootstrap user and that the requested cert matches the expected shape for a kubelet cert

see https://github.com/kubernetes/kubernetes/tree/master/pkg/controller/certificates/groupapprove.go#L40

Quentin-M:
test this

pkg/bootkube/approve.go
Quentin-M: Fixed. Thanks.

PR #263, opened 2017-01-06
Rename label master=true to kubernetes.io/role=master & Pin controller services

### LINKED ISSUE #246: use upstream labels for member roles
`kubernetes.io/role=master` and `kubernetes.io/role=node`.

https://github.com/kubernetes/kubernetes/pull/35975/files#diff-c8063ac1e636c387ca8169633ddee1bdR54

Quentin-M:
```
=== all: Rename label master=true to kubernetes.io/role=master  …
This commit reconciles with the upstream Kubernetes' preferences
regarding to role labeling (https://git.io/vMckM). This also enables
users to find the nodes' role using `kubectl describe`
(https://git.io/vMckb).

Additionally, label the minions with the `node` role in the
multi-node example.
```

```
=== pkg/asset: Pin Controller-Manager, Scheduler and KubeDNS to master nodes  …
Because master nodes are considered first-class citizens and more stable
than worker nodes, it is safer to schedule these services onto them.
This is also a step forward to actually making master nodes controller
nodes.
```

Fixes #246.

### DIFF
--- a/Documentation/upgrading.md
+++ b/Documentation/upgrading.md
@@ -10,10 +10,10 @@ Let's upgrade a self-hosted Kubernetes v1.4.1 cluster to v1.4.3 as an example.
 Show the control plane daemonsets and deployments which will need to be updated.
 
     $ kubectl get daemonsets -n=kube-system
-    NAME             DESIRED   CURRENT   NODE-SELECTOR   AGE
-    kube-apiserver   1         1         master=true     5m
-    kube-proxy       3         3         <none>          5m
-    kubelet          3         3         <none>          5m
+    NAME             DESIRED   CURRENT   NODE-SELECTOR                AGE
+    kube-apiserver   1         1         kubernetes.io/role=master    5m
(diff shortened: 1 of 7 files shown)


Quentin-M:
bootkube-dev will only be green with https://github.com/coreos-inc/pluton/pull/23.

Quentin-M:
Worth noting: https://github.com/kubernetes/kubernetes/pull/39112

Quentin-M:
Because we are not going to support migrations to 1.5.2, it might be worth merging the labeling part of this PR now. The pinning part will be handled by https://github.com/kubernetes-incubator/bootkube/pull/276.

However, instead of switching to what Kubernetes suggests using now, we might want to take a step ahead and use what's proposed in https://github.com/kubernetes/kubernetes/pull/39112 instead.

Quentin-M:
**DO NOT MERGE NOR REVIEW**

Waiting on https://github.com/kubernetes/kubernetes/pull/39112.

Issue #262, opened 2017-01-06
master reboot test is failing


pbx0:
This seemed to slip into the master branch during when the tests were temporarily broken over some of the self-hosted-flannel changes. If I reboot a master node, the api-server doesn't come back up. @aaronlevy is already on it and suspects what the problem is.

aaronlevy:
There are a few different issues at play here - but the core issue seems to be that the checkpointer is relying on determining local state from the kubelet `/pods` api - but that api may not reflect accurate local state.

Unfortunately that kubelet api endpoint will just report state from the last time it was able to successfully contact an api-server. Essentially a cache of the last state reported to api.

We didn't have this same issue with the old checkpointer - because it would determine "api is running" by reaching out to an api -- but that isn't reliable in multi-master, or for generic checkpointing.

This is a rather disappointing discovery -- as it seems there is no way to easily determine the local pod state if an api-server is not available.

Some options moving forward:

1. Go back to hard-coded expectations (e.g. `curl localhost:8080`)
1. Extract liveness probes from pod specs and try to use those if available
1. Add some kind of "last seen" annotation to checkpoint parent pods, where we expect to see an update within a particular time window.
1. Longer term option: make use of CRI interface and inspect state ourselves

/cc @yifan-gu @dghubble

Quentin-M:
Just to be sure, does this affect 0.3.2 only or 0.3.1 too?

Quentin-M:
I am not too familiar with the checkpointer itself and the contraints that surround it but I believe that relying on liveness probes could be a decent solution. It is simple, users are familiar with the concept and they would expect the checkpointer to rely on them. It is also user customizable in a way. Additionally, the project also vendors Kubernetes so the functions that achieve that could be re-used directly. 

The downside is that if we do not want to add the requirement of defining probes, this is only part of the solution.

aaronlevy:
The other part of this issue is that api-server behavior has changed in v1.5:

 In earlier versions, the api-server would continually re-try to bind to particular addresses (:443 & :8080). If they were already in use, it would just try again in 15 seconds.

The behavior in v1.5 is that the api-server will exit immediately if it is not able to listen on those addresses (and would rely on external mechanisms of systemd/kubelet/etc to restart it again).

In terms of the checkpointer - we need a reliable way to determine "real api-server is running, or it is trying to be run". The "trying to be run" is important, because we need to remove an active checkpoint in this situation, so the real server can actually start (and bind on 443/8080).

[...]

PR #264, opened 2017-01-09
pkg/asset/internal: Introduce a Flock around the API server

Quentin-M:
This commit represents a workaround for #262. By maintaining
a file lock while the API server is running (either temporary
or self-hosted), we prevent the self-hosted API server from
starting and trying to bind ports, until the temporary one is
stopped. Therefore, we avoid the loop where the self-hosted
API server would crash as soon as it is brought up due to the
ports already being bound by the stopping temporary server.

cc @aaronlevy

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -134,6 +134,10 @@ spec:
       - name: kube-apiserver
         image: quay.io/coreos/hyperkube:v1.5.1_coreos.0
         command:
+        - /usr/bin/flock
+        - --exclusive
+        - --timeout=30
+        - /var/lock/api-server.lock
         - /hyperkube
         - apiserver
         - --bind-address=0.0.0.0
@@ -164,6 +168,9 @@ spec:
         - mountPath: /etc/kubernetes/secrets
           name: secrets
           readOnly: true
+        - mountPath: /var/lock
+          name: var-lock
+          readOnly: false
       volumes:
       - name: ssl-certs-host
         hostPath:
@@ -171,6 +178,9 @@ spec:
       - name: secrets
         secret:
           secretName: kube-apiserver
+      - name: var-lock
+        hostPath:
+          path: /var/lock
 `)
 	CheckpointerTemplate = []byte(`apiVersion: "extensions/v1beta1"
 kind: DaemonSet


Quentin-M:
@aaronlevy

aaronlevy:

pkg/asset/internal/templates.go
```diff
@@ -134,6 +134,10 @@ spec:
       - name: kube-apiserver
         image: quay.io/coreos/hyperkube:v1.5.1_coreos.0
         command:
+        - flock
```
aaronlevy: nit: full path to binary

pkg/asset/internal/templates.go
```diff
@@ -164,13 +168,18 @@ spec:
         - mountPath: /etc/kubernetes/secrets
           name: secrets
           readOnly: true
+        - mountPath: /var/lock
+          name: var-lock
```
aaronlevy: nit: `readOnly: false` (default, but also explicit)

dghubble:
sgtm

Quentin-M:
Fixing nits now.

pkg/asset/internal/templates.go
Quentin-M: Fixed.

aaronlevy:
For posterity (was added in side-discussion):

Between v1.4.x and v1.5.x there was a change in api-server behavior:

In v1.4 if the apiserver failed to listen on a particular address (e.g. :8080 & :443) - it would keep running but keep re-trying to bind.

In v1.5 the apiserver will immediately exit if it is unable to listen on an address. The behavior changed in: kubernetes/kubernetes#35109

The existing functionality allowed us to start the self-hosted api-server while the checkpoint was still running. However, now we can get in failure loops because the checkpoint/parent will conflict with eachother - sometimes the transition works, and sometimes not.

The lockfile allows us to start the self-hosted parent, even while the checkpoint is still running. All of the needed assets can be downloaded (e.g. secrets/configMaps) without the pod dying, then gives us a more predictable timeframe to clean up the checkpointed copy- allowing the pivot to the self-hosted pod.

aaronlevy:

pbx0:

Quentin-M:
Test failed for an unrelated reason: `Error:  client: etcd cluster is unavailable or misconfigured`

aaronlevy:
@Quentin-M we took a look and it seems to be a selinux issue

aaronlevy:
Gonna go ahead a merge. We did manual tests until the CI tests are fixed.

luxas:
cc @pires ^

@aaronlevy Do we need to do this for the scheduler and/or the controller-manager also?
Or do they have the "old" behaviour to wait nicely?

Quentin-M:
@luxas This only required for the API server, this is the only pod we checkpoint.

aaronlevy:
@luxas also little more detail: controller-manager and scheduler do not use hostNetwork, so there would be no port conflict anyway (and are leader-elected).

PR #265, opened 2017-01-10
hack/multi-node: Add option to launch with self-hosted etcd

aaronlevy:
This is against yet to be merged code (https://github.com/kubernetes-incubator/bootkube/pull/259)

This is to help test that PR more easily.

Usage is:

```
cd hack/multi-node
rm -rf cluster
vagrant destroy -f
SELF_HOST_ETCD=true ./bootkube-up
```


xiang90:
lgtm

Issue #267, opened 2017-01-11
Will not recover if you delete the last api-server pod


aaronlevy:
We should be able to recover from deletion of the only api-server pod (as long as it is ultimately managed by a higher-level object like a daemonset).

However, with the change to the new checkpointer in v0.3.1 there is an issue where the local kubelet state will only be updated if it can contact an api-server. So checkpointer asks kubelet "is the api-server running" and it responds "yes". If that is the last api-server pod that was just deleted -- the local state will never be updated to reflect that it is not in fact running.

The old checkpointer would work around this because it would just try and hit "localhost:8080" to determine if the api-server is running -- but that has downsides that it means the checkpointer is no longer a generic tool (has to know about certain pods). And it also isn't actually accurate (it could be that the checkpoint copy is running and listening on :8080).

But it would be a short term option to keep allowing that same functionality.

/cc @Quentin-M

Quentin-M:
We discussed about potential solutions at the office. The two that sounded the most reasonable, given the existing constraints and the desire of keeping a generic checkpointer are:

- Extracting the liveness probe from a pod specification and querying it in order to determine the pod state. The good point is that we give some control to the developer over what the checkpointer should do with his pod. The drawback is that we could only do HTTP probes (and that someone might eventually ask for script support..) and that it might not be reliable in every cases.

- Actually contacting the local container runtime for the state of the pods. The advantage is that it is a quite simple and reliable solution. It can be done either by implementing the calls to existing runtimes (supposedly low amount of code and maintenance - there are few runtimes, rarely new ones) or by leveraging the Kubelet's code (which means we need to vendor it...).

aaronlevy:
Just to note: the workaround to this in the interim is to use an older version of the checkpointer (until a fix is merged):  quay.io/coreos/pod-checkpointer:b4f0353cc12d95737628b8815625cc8e5cedb6fc

Also the fix I plan for is using the `/runningpods/` api from the kubelet to determine the needed information.

PR #268, opened 2017-01-12
checkpoint: Get local pod running state from /runningpods/ endpoint

aaronlevy:
Fixes https://github.com/kubernetes-incubator/bootkube/issues/267

/cc @Quentin-M

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -20,7 +21,6 @@ import (
 	"k8s.io/kubernetes/pkg/api/v1"
 	clientset "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_5"
 	"k8s.io/kubernetes/pkg/client/unversioned/clientcmd"
-	clientcmdapi "k8s.io/kubernetes/pkg/client/unversioned/clientcmd/api"
 	"k8s.io/kubernetes/pkg/fields"
 	"k8s.io/kubernetes/pkg/runtime"
 )
@@ -28,7 +28,11 @@ import (
 const (
 	nodeNameEnv = "NODE_NAME"
 
-	kubeletAPIPodsURL = "http://127.0.0.1:10255/pods"
+	// We must use both the :10255/pods and :10250/runningpods/ endpoints, because /pods endpoint could have stale data.
+	// The /pods endpoint will only show the last cached status which has successfully been written to an apiserver.
+	// However, if there is no apiserver, we may get stale state (e.g. saying pod is running, when it really is not).
+	kubeletAPIPodsURL        = "http://127.0.0.1:10255/pods"
+	kubeletAPIRunningPodsURL = "https://127.0.0.1:10250/runningpods/"
 
 	activeCheckpointPath   = "/etc/kubernetes/manifests"
 	inactiveCheckpointPath = "/srv/kubernetes/manifests"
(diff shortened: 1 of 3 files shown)


cmd/checkpoint/main.go
```diff
@@ -113,6 +124,10 @@ func process(localParentPods, apiParentPods, activeCheckpoints, inactiveCheckpoi
 			// If the inactive checkpoint still has a parent pod, do nothing.
 			// This means the kubelet thinks it should still be running, which has the same scheduling info that we do --
 			// so we won't make any decisions about its checkpoint.
+			// TODO(aaron): This is a safety check, and may not be necessary -- question is do we trust that the api state we recieved
+			//              is accurate -- and that we should ignore our local state (or assume it could be innacurate). For example,
+			//              local kubelet pod state will be innacurate in the case that we can't contact apiserver (kubelet only keeps
```
derekparker: nit: s/innacurate/inaccurate/

cmd/checkpoint/main.go
```diff
@@ -311,6 +334,28 @@ func getLocalParentPods() (map[string]*v1.Pod, error) {
...
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response: %v", err)
+	}
+
+	var podList v1.PodList
+	if err := json.Unmarshal(b, &podList); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal podlist: %v", err)
+	}
+	return filterPodList(&podList, func(*v1.Pod) bool { return true }), nil
```
derekparker: This never filters anything, is this code just reliant on a side effect of `filterPodList`?

aaronlevy: Yeah, kinda poorly named. I'll try and update to make more clear of intentions.

cmd/checkpoint/main.go
```diff
@@ -311,6 +334,28 @@ func getLocalParentPods() (map[string]*v1.Pod, error) {
 	return podListToParentPods(&podList), nil
 }
 
+// getLocalRunningPods uses the /runningpods/ kubelet api to retrieve the local container runtime pod state
+func getLocalRunningPods() (map[string]*v1.Pod, error) {
+	// TODO(aaron): The kubelet api is currently secured by a self-signed cert. We should update this to actually verify at some point
```
ericchiang: So a couple points on this.

- Can the checkpointer manifests include secrets? We could include the CA later there if we needed to.
- One of the options for kubelet API auth(n/z) is the "token webhook" mode. Which let's you use things like service account tokens to query the kubelet API, but it requires the kubelet to be able to query the API server to determine if the token is valid and has the correct permissions. (see: `--authentication-token-webhook` and `--authorization-mode=Webhook`)
- This means we'll have to use client certs here if we ever want to do kubelet auth.

This isn't show stopping but something we should be aware of.

aaronlevy: - Because the checkpointer is run as a static manifest - the checkpointer cannot rely on secrets or service accounts because it will run regardless of an api-server being available.

- Can't really rely on this because we can't rely on an api-server being available - this has to function only relying on local information (but makes use of the api when available).

- The CA should ultimately be available somewhere on the host (the kubelet needs it). So as long as the kubelet cert is signed by that CA we should be fine (e.g. the checkpointer already relies on the kubeconfig provided to the kubelet in `/etc/kubernetes/kubeconfig` (we'll also need to figure out our options here when we switch to the kubelet TLS bootstrap - the location change might cause issue)

ericchiang: The checkpointer doesn't checkpoint its own secrets? :)

Yeah, it's not impossible to work around, but by relying on the kubelet API when the API server's down means we can't use that feature. More of a mental note for me when we try to enable kubelet auth.

aaronlevy: I know it's in jest, but just to note: static pods can't use any api-server provided assets (secrets, config maps, etc.), so there's nothing to checkpoint even if it wanted to (also service accounts are secrets).

In the long term we have a few options. But ultimately what we need is a way to determine "are pods running" purely from local state:

- Create a `/runningpods/` like api on the insecure/read-only kubelet api
- Query the CRI interface (once used/available) to determine pod state directly from runtime
- Query runtime(s) directly

PR #271, opened 2017-01-12
hack/multi-node: self hosted etcd option

hongchaodeng:
Separated from #259.

We have tested it.

### DIFF
--- a/hack/multi-node/bootkube-up
+++ b/hack/multi-node/bootkube-up
@@ -7,9 +7,18 @@ if [ ${un} == 'Darwin' ]; then
     local_os="darwin"
 fi
 
+SELF_HOST_ETCD=${SELF_HOST_ETCD:-false}
+if [ ${SELF_HOST_ETCD} = "true" ]; then
+    etcd_render_flags="--etcd-servers=http://10.3.0.15:2379 --storage-backend=etcd3 --experimental-self-hosted-etcd"
+    etcd_start_flags="--etcd-server=http://172.17.4.101:12379,http://10.3.0.15:2379 --experimental-self-hosted-etcd"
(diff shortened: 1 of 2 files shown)


hack/multi-node/bootkube-up
```diff
@@ -10,7 +10,7 @@ fi
 SELF_HOST_ETCD=${SELF_HOST_ETCD:-false}
 if [ ${SELF_HOST_ETCD} = "true" ]; then
     etcd_render_flags="--etcd-servers=http://10.3.0.15:2379 --storage-backend=etcd3 --experimental-self-hosted-etcd"
-    etcd_start_flags="--etcd-server=http://172.17.4.101:12379,http://10.3.0.15:2379 --experimental-self-hosted-etcd"
+    etcd_start_flags="--etcd-server=http://172.17.4.101:12379 --experimental-self-hosted-etcd"
```
aaronlevy: But this won't work if the etcd pod ever leaves this one node...

aaronlevy: And in fact, this is the temporary etcd pod -- how does this work at all?

hongchaodeng: See my explanation in https://github.com/kubernetes-incubator/bootkube/pull/259#issuecomment-272307260

xiang90: basically, the temp api server only contacts the boot member. once the boot member dies, bootkube dies. we know the boot member dies by contacting the first in-cluster etcd member directly. (alternative, we know the boot member dies if temp api server cannot contact to the boot member anymore)

there is no point that the temp api server needs to contact the first in cluster member i think.

PR #274, opened 2017-01-17
hack/tests: add logging and a delay when launching GCE workers

aaronlevy:
Getting intermittent "Too many authentication failures" when launching GCE nodes. This seems to help and is a stop-gap until can dig into why this is occurring.

/cc @Quentin-M

### DIFF
--- a/hack/tests/conformance-gce.sh
+++ b/hack/tests/conformance-gce.sh
@@ -70,11 +70,16 @@ function add_master {
 function add_workers {
     #TODO (aaron): parallelize launching workers
     for i in $(seq 1 ${WORKER_COUNT}); do
+        echo "Launching worker"
         gcloud compute instances create ${GCE_PREFIX}-w${i} \
             --image-project coreos-cloud --image-family ${COREOS_CHANNEL} --zone us-central1-a --machine-type n1-standard-1
 
+        echo "Adding ssh-key to worker metadata"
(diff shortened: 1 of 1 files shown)


hack/tests/conformance-gce.sh
```diff
@@ -70,11 +70,15 @@ function add_master {
...
     for i in $(seq 1 ${WORKER_COUNT}); do
+        echo "Launching worker"
         gcloud compute instances create ${GCE_PREFIX}-w${i} \
             --image-project coreos-cloud --image-family ${COREOS_CHANNEL} --zone us-central1-a --machine-type n1-standard-1
 
+        echo "Adding ssh-key to worker metadata"
         gcloud compute instances add-metadata ${GCE_PREFIX}-w${i} --zone us-central1-a --metadata-from-file ssh-keys=/root/.ssh/gce-format.pub
 
+        echo "Getting worker public IP"
+        sleep 30 # Have seen "Too many authentication failures" in CI jobs -- this seems to help
```
derekparker: nit: would put this sleep before the echo, could be confusing to say you're going to do something and then sleep for a while, but won't block on this.

derekparker: Also might want to put in a TODO here or something, and open up another issue (or update the current issue) to follow up on digging into this.

aaronlevy:
Addressed nits and opened issue: https://github.com/kubernetes-incubator/bootkube/issues/277

PR #276, opened 2017-01-17
pkg/asset: Pin kube-controller-manager and kube-scheduler to masters

Quentin-M:
Because master nodes are considered first-class citizens and more stable
than worker nodes, it is safer to schedule these services onto them.
This is also a step forward to actually making master nodes controller
nodes.

This commit is extracted from https://github.com/kubernetes-incubator/bootkube/pull/263 as it doesn't require running any migrations and therefore is easier to merge adopt right now.

After this change, all three kube-apiserver, kube-controller-manager and kube-scheduler will run exclusively on master nodes.


Quentin-M:
test this

Quentin-M:
test cluster

Quentin-M:
please test

Quentin-M:
test this please

aaronlevy:
Until we have multi-master at deploy time, this just means we are forcing all copies onto the same node - so this will in many cases be less resilient.

For these components we could launch multiple nodes with this label, and they will work out-of-box right now -- but it would be a little disingenuous because the api-server would only be accessed on a single node.

sym3tri:
@aaronlevy what do you mean by "multi-master at deploy time" WRT to bootkube? Won't pinning the manifests just run on as many master nodes as are available in the cluster?

aaronlevy:
Unto itself I don't see what this PR solves. It's not isolating workloads (so why not label everything as master), and unless launching multiple nodes with master=true, it actually makes the cluster less resilient.

If we launch multiple nodes with master=true, why not just label them all like that? Really that label was just meant for forcing the api-server on certain nodes to make it easily discoverable from fixed IPs - not (yet) for implementing dedicated nodes (e.g. "only run this special workload here").

Then if we're labeling everything as master (or even a subset), the original intent (api-server addressability) is lost (it will run everywhere, but we will only address it on a specific node until we add AWS/GCE loadbalancer support).

Quentin-M:
Completely isolating workloads involves using taints/tolerations to lock down the master nodes to only having certain pods scheduled on them, which is inconceivable for bootkube without breaking the single node use-case (user workload can't be ran on the master node anymore) or introducing a special case.

We actually want to do that but don't expect this to be available in bootkube due to that reason. I imagine we should simply stop relying on _render_ and invest into #168 to completely dissociate bootkube from the actual resources we need (i.e using our own approval policies for #261 even during bootkube's bootstrapping process).

dghubble:
@aaronlevy There was a specific intent behind this change. At the moment, a user may add and then remove workers quickly enough that both controller managers or both schedulers are knocked out. The cluster cannot recover, there is no checkpointing of those components. It has happened.

I also see your point about forcing all copies onto the same node. In that case, when the controller reboots, the same issue should occur.

Since master reboots are a certainty and rapid worker scale down is more of an edge case, this does seem undesired without more of a plan.

Also agreed, this isn't about isolating workloads. Isolating workloads, taints, tolerances, and those things are separate, future items to address.

aaronlevy:
Just want to preface with -- I'm not outright opposed to this, it's just that we need to consider the implications.

@Quentin-M I'm actually less concerned about long-term support of the single-node installation. It really is just for quick development (or testing on memory constrained machines) -- but it's not what we should model any "ideal" installations after. I could be convinced to at some point completely remove it - or maybe special case it (e.g. remove the master taints or something).

@sym3tri @dghubble @Quentin-M -- I also think that we should strive toward a point of "ideal reference implementation" -- which in my mind (in the short term) is a "dedicated" 3 node control plane (with taints) which runs pods for api/sched/cm/etcd/dns. All deployments have anti-affinity to their own pods, replicas=3, and affinity to "masters". All master nodes sit behind some kind of LB / DNS - and api-server is universally referenced (in cluster) from that endpoint.

But we're not quite there yet, so it's somewhat of a balancing act at this stage -- where an isolated change like this could be questionable in utility.

I also think that we will need to continue to support a single-master installations. This is a common, and real (see all of GKE), installation method that we can't outright discount. Again, this is a balancing act of figuring out what would work and how do we still support this option.

@Quentin-M not entirely sure how https://github.com/kubernetes-incubator/bootkube/issues/168 would change any of this -- where the temporary control plane runs isn't much of an issue. This would be more about modifying the control-plane templates prior to `bootkube start` if you wanted customizations that are not present here. But that's already the case (modify as much as you like), static bootstrap manifests (I don't think) will help here.

@dghubble reboots should be fine even in the single-master use-case. As long as the node doesn't disappear, those pods are already scheduled to the node, so on reboot they just start again (don't need checkpointing).

@dghubble @Quentin-M @sym3tri so to get to the root of the issue: how do we protect against users destroying too many worker nodes, which might be running required components.

- One option is this PR, to pin to master nodes.
    - Some concerns: we might need to also add anti-affinity or replicas>1 doesn't make sense in single-master case (or change to replicas=<expected initial # of masters>).
    - `kubectl drain <master-node>` would kill the cluster in single-master, because those pods have nowhere to go. So if we go this route we might need to mitigate this by adding eviction policies around these pods.
- Another random option would be to add a systemd shutdown unit which tries to evict pods (can use eviction quotas to attempt to block until satisfied). The issue here is if the power is just pulled, it won't go through the eviction process. And, we very likely do not want to drain nodes on every reboot (but rather "i'm being destroyed") -- not sure if this is possible to determine, and likely cloud-specific (e.g. AWS-destruction-shutdown.service).
- Or...??

dghubble:
@aaronlevy great, that clears up my worry about controller reboots. I'm for continuing down the contorller/scheduler pinning and looking into eviction policies so master drains aren't deadly. Separately, I think we said # master's + 1 replicas for those components, to support rolling upgrades and single-master.

aaronlevy:
sgtm

Quentin-M:
Just added disruption budgets to avoid breaking the cluster where `kubectl drain` is called on a cluster that has a single master. 

```
$ kubectl --kubeconfig=cluster/auth/kubeconfig drain 172.17.4.100 --ignore-daemonsets --force --timeout=15m
node "172.17.4.100" cordoned
WARNING: Deleting pods not managed by ReplicationController, ReplicaSet, Job, DaemonSet or StatefulSet: pod-checkpointer-172.17.4.100; Ignoring DaemonSet-managed pods: checkpoint-installer-63147, kube-apiserver-ss9bv, kube-flannel-rqcdd, kube-proxy-v5w5j
pod "kube-controller-manager-2101915543-jnk2p" evicted
pod "kube-scheduler-2712781253-tz3f2" evicted
pod "kube-dns-4101612645-bpvj0" evicted
error: Drain did not complete within 15m
```

```
$ kubectl get pods -n=kube-system
NAME                                       READY     STATUS    RESTARTS   AGE       IP             NODE
checkpoint-installer-63147                 1/1       Running   0          18m       172.17.4.100   172.17.4.100
kube-apiserver-ss9bv                       1/1       Running   3          18m       172.17.4.100   172.17.4.100
kube-controller-manager-2101915543-g863l   0/1       Pending   0          4m        <none>
kube-controller-manager-2101915543-rzh6n   1/1       Running   0          19m       10.2.0.2       172.17.4.100
kube-dns-4101612645-jzxd7                  0/4       Pending   0          4m        <none>
kube-flannel-rqcdd                         2/2       Running   0          18m       172.17.4.100   172.17.4.100
kube-proxy-v5w5j                           1/1       Running   0          18m       172.17.4.100   172.17.4.100
kube-scheduler-2712781253-9z3w5            0/1       Pending   0          4m        <none>
kube-scheduler-2712781253-t43fz            1/1       Running   0          19m       10.2.0.6       172.17.4.100
pod-checkpointer-172.17.4.100              1/1       Running   0          18m       172.17.4.100   172.17.4.100
```

```
$ kubectl --kubeconfig=cluster/auth/kubeconfig uncordon 172.17.4.100

$ kubectl get pods -n=kube-system
NAME                                       READY     STATUS    RESTARTS   AGE       IP             NODE
checkpoint-installer-63147                 1/1       Running   0          29m       172.17.4.100   172.17.4.100
kube-apiserver-ss9bv                       1/1       Running   3          29m       172.17.4.100   172.17.4.100
kube-controller-manager-2101915543-g863l   1/1       Running   0          15m       10.2.0.7       172.17.4.100
kube-controller-manager-2101915543-rzh6n   1/1       Running   0          30m       10.2.0.2       172.17.4.100
kube-dns-4101612645-jzxd7                  4/4       Running   0          15m       10.2.0.8       172.17.4.100
kube-flannel-rqcdd                         2/2       Running   0          29m       172.17.4.100   172.17.4.100
kube-proxy-v5w5j                           1/1       Running   0          29m       172.17.4.100   172.17.4.100
kube-scheduler-2712781253-9z3w5            1/1       Running   0          15m       10.2.0.9       172.17.4.100
kube-scheduler-2712781253-t43fz            1/1       Running   0          30m       10.2.0.6       172.17.4.100
pod-checkpointer-172.17.4.100              1/1       Running   0          29m       172.17.4.100   172.17.4.100

```

I believe that we can add anti-affinity policies in a separate PR as there are not strictly necessary to this specific PR, and also because we may want to wait until Kubernetes 1.6 lands (depending on how high that is on the priority bucket) so their implementation reaches Beta and [settles](https://groups.google.com/forum/m/#!msg/kubernetes-dev/-fwfHQtm9GE/H1TU4yqiBAAJ) on the manifest format.

I may embed the disruption manifests directly into the kube-controller-manager and kube-scheduler templates if you think it makes sense and want to reduce the amount of rendered templates.

pbx0:
So, just to summarize: It seems like the main concerns here were based around what the long-term plan is for an HA reference implementation looks like and if these changes will have short-term impacts on single-master resiliency. 

Long-term it makes sense to me to special case master nodes as a a place for kube-system components to run. I imagine we will want the self-hosted etcd pods to run on nodes that are considered a little less ephemeral.

Looks like quentin has addressed the master draining issue and we have plans to add anti-affinity to these pods in 1.6.

Additionally, although scheduling apiservers to other master's without a load-balancer used to address it doesn't add us any resiliency, I also don't see this PR introducing another failure case here. If the original intent was for the master label to just distinguish where the single apiserver is located isn't that not really how it works right now? Since the apiserver location is essentially hard-coded so using the master label to change where the apiserver gets scheduled is already pretty inflexible.

This also solves the short-term use-case of distinguishing nodes that run kube-system components so that a user destroying worker nodes doesn't trample on these system components. Overall I think this is fine to merge. I'll let @derekparker give it a once over as well when he wakes up.

PR #280, opened 2017-01-18
checkpoint: re-use http client transport

aaronlevy:
/cc @ericchiang


aaronlevy:
For posterity: description of issue: https://github.com/golang/go/issues/16267#issuecomment-230540719

PR #287, opened 2017-01-20
	 SHE: remove boot etcd manifest after migration

### LINKED ISSUE #283: self hosted etcd: remove boot etcd manifest after migration
We create the boot etcd static pod manifest for creating the initial out-cluster etcd. After we successfully migrate etcd into Kubernetes cluster, we should remove that static manifest. 

Or the kubelet will try to start the boot etcd again when it restarts (even if the restart policy is Never).

One thing I do not understand is that the boot etcd will be started as expected by Kubelet after a reboot, but it is always pending.

hongchaodeng:
fixes #283

### DIFF
--- a/pkg/util/etcdutil/start_etcd.go
+++ b/pkg/util/etcdutil/start_etcd.go
@@ -9,8 +9,10 @@ import (
 	"github.com/golang/glog"
 )
 
+const bootEtcdFilePath = "/etc/kubernetes/manifests/boot-etcd.yaml"
+
 func StartEtcd(endpoint string) error {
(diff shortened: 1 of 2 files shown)


pkg/util/etcdutil/start_etcd.go
```diff
@@ -9,8 +9,10 @@ import (
 	"github.com/golang/glog"
 )
 
+const bootEtcdFilePath = "/etc/kubernetes/manifests/boot-etcd.yaml"
```
xiang90: bootEtcdManifestsFilePath

xiang90:
lgtm

PR #293, opened 2017-02-04
pkg: bump hyperkube to v1.5.2_coreos.2

squat:


Documentation/upgrading.md
```diff
@@ -33,7 +33,7 @@ In this case, Kubernetes is `v1.4.1+coreos.0` and our goal is to upgrade to `v1.
 
 ### kube-apiserver
 
-Edit the kube-apiserver daemonset. Change the container image name to `quay.io/coreos/hyperkube:v1.4.3_coreos.0`.
+Edit the kube-apiserver daemonset. Change the container image name to `quay.io/coreos/hyperkube:v1.5.2_coreos.2`.
```
pbx0: This is an example version so lets not bump it so it fits with the above parts of this guide

squat: :+1:

pbx0:
Sorry -- one more nit: Could you also bump this in the `Makefile` and  `hack/tests/conformance-test.sh` for completeness? Its not strictly necessary since this wasn't a k8s code change per-say, just a change to the Hyperkube build script but it makes future search and replaces easier.

PR #296, opened 2017-02-08
Add etcd-operator to requiredPods when selfHostedEtcd is true

adieu:
In some environments, pulling image from `gcr.io` is faster than `quay.io`. We might call `etcdutil.Migrate()` too early even before `etcd-operator` got started.

In this case, the user will have `fail to wait etcd TPR to be ready: timed out waiting for the condition` error and the bootstrap process got terminated.

This PR add `etcd-operator` to `requiredPods` when `selfHostedEtcd` is true. We'll make sure `etcd-operator` is running before calling `etcdutil.Migrate()`.  Making the bootstrap process less likely to fail.

### DIFF
--- a/pkg/bootkube/bootkube.go
+++ b/pkg/bootkube/bootkube.go
@@ -110,6 +110,9 @@ func (b *bootkube) Run() error {
 			errch <- err
 		}
 	}()
+	if b.selfHostedEtcd {
+		requiredPods = append(requiredPods, "etcd-operator")
+	}
 	go func() { errch <- WaitUntilPodsRunning(requiredPods, assetTimeout, b.selfHostedEtcd) }()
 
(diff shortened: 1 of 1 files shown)


pkg/bootkube/bootkube.go
```diff
@@ -110,6 +110,9 @@ func (b *bootkube) Run() error {
 			errch <- err
 		}
 	}()
+        if b.selfHostedEtcd {
```
xiang90: can you please fix the indentation here? `if` needs to be aligned with previous statements.

you can run gofmt to do it.

xiang90:
The code change makes sense. Please format the changed code.

pbx0:
OK to test

adieu:
fixed the format. I was editing the file directly on an linux server.

xiang90:
@aaronlevy can you please take a look?

PR #298, opened 2017-02-11
templates: ensure critical components dont coexist

squat:
This commit adds antiaffinity rules to the controller manager and the
sheduler to ensure that these deployments' pods do not all end up on the
same master node, especially now that they are pinned with node
selectors. To make this possible on small clusters, e.g. those with 1
master node, this is specified with "preferred" rather than "required".

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -270,6 +270,26 @@ spec:
     metadata:
       labels:
         k8s-app: kube-controller-manager
+      annotations:
+        scheduler.alpha.kubernetes.io/affinity: |
+          {
+            "podAntiAffinity": {
+              "preferredDuringSchedulingIgnoredDuringExecution": [
+                {
+                  "weight": 100,
+                  "labelSelector": {
+                    "matchExpressions": [
+                      {
+                        "key": "k8s-app",
+                        "operator": "In",
+                        "values": ["kube-controller-manager"]
+                      }
+                    ]
+                  }
+                }
+              ]
+            }
+          }
     spec:
       nodeSelector:
         master: "true"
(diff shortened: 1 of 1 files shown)


bzub:
On a running cluster this is an invalid manifest, for me.  The annotation is added under spec.template.metadata, and are you sure it doesn't belong under metadata?  That's how the examples show it here: https://kubernetes.io/docs/user-guide/node-selection/  That change made it valid.

squat:
@bzub thanks for pointing out that there was an error. The annotation is definitely supposed to be in `.spec.template.metadata` because the `scheduler.alpha.kubernetes.io/affinity` annotation is specific to pods (the link you provided shows that as well).

The reason it wasn't working was because I actually wrote the annotation wrong :sweat_smile: `preferredDuringSchedulingIgnoredDuringExecution` is supposed to be an array of weighted affinity structs, not a single affinity struct as I had it before. Thanks for pointing it out. I tried it out on my cluster and it is working. 

The reason it "worked" when you put it directly in `.metadata` was because `scheduler.alpha.kubernetes.io/affinity` is not a whitelisted deployment annotation so kubernetes did not recognize nor try to parse that annotation, so there was no error to return.

bzub:
Ah ok.  I appreciate the explaination.

Quentin-M:
@squat The [initial plan](https://github.com/kubernetes-incubator/bootkube/pull/276) was to wait until Kubernetes 1.6 to introduce this due to the upcoming [non-backward compatible changes](https://groups.google.com/d/msg/kubernetes-dev/-fwfHQtm9GE/H1TU4yqiBAAJ).

squat:
@Quentin-M sure, let's wait to merge this then

Quentin-M:
I think it should apply to kube-apiserver too.

aaronlevy:
If we're not planning on merging this, I'm going to close this PR for now (these won't be annotations in v1.6). But can re-open if this is something we want to add now and migrate to new form after v1.6.

@Quentin-M the api-server is a daemonset, so you can't have pods scheduled to same node - but if we change to a deployment, then yes, this should also be added for api-server.

squat:
@aaronlevy let me know when you'd like to re-open this. I updated the branch to use the new 1.6 affinity format. This PR uses `preferred` affinity, that is the _soft_ requirement, in order to allow two replicas to be scheduled when there is only one master. Branch updates to not show up because the PR is closed :(. I can make a new PR if necessary

aaronlevy:
@squat - now that v1.6 has merged, we should get this functionality merged. Feel free to open a new PR or we can re-open this one as well.

squat:
Thanks @aaronlevy. I reopened this as #421.

PR #307, opened 2017-02-15
Detect pod and service CIDR from manifests

dghubble:
* Temporary apiserver should use service CIDR detected from `kube-apiserver.yaml`
* Temporary controller-manager should use pod CIDR detected from `kube-controller-manager.yaml`
* Eliminates the need for coordination btw render and start and unblocks pod/service CIDR customization

Full discussion about why we're taking this approach over others: https://github.com/kubernetes-incubator/bootkube/issues/236#issuecomment-279797787

related. #236

#### Examples of edge cases

Missing assets:

```
$ bootkube start --asset-dir=empty
Error: can't read file manifests/kube-apiserver.yaml: open empty/manifests/kube-apiserver.yaml: no such file or directory
can't read file manifests/kube-apiserver.yaml: open empty/manifests/kube-apiserver.yaml: no such file or directory
```

kube-apiserver.yaml (or controller manager) named container not found.

```
Error: can't detect --service-cluster-ip-range in manifests/kube-apiserver.yaml
can't detect --service-cluster-ip-range in manifests/kube-apiserver.yaml
```

kube-apiserver.yaml (or controller manager) uses an invalid CIDR.

```
bootkube start --asset-dir=assets
Error: invalid --cluster-cidr CIDR: invalid CIDR address: pickles
invalid --cluster-cidr CIDR: invalid CIDR address: pickles
```

kube-apiserver.yaml (or controller manager) is missing the required flag entirely.

```
Error: invalid --cluster-cidr CIDR: invalid CIDR address: 
invalid --cluster-cidr CIDR: invalid CIDR address: 
```

kube-apiserver.yaml (or controller manager) isn't valid YAML

```
Error: cant unmarshal manifests/kube-apiserver.yaml: error converting YAML to JSON: yaml: line 1: did not find expected <document start>
cant unmarshal manifests/kube-apiserver.yaml: error converting YAML to JSON: yaml: line 1: did not find expected <document start>
```

### DIFF
--- a/pkg/bootkube/parse.go
+++ b/pkg/bootkube/parse.go
@@ -1,0 +1,87 @@
+package bootkube
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net"
+	"path/filepath"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/kubernetes-incubator/bootkube/pkg/asset"
+	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/pkg/apis/extensions/v1beta1"
+)
+
+const (
+	apiServerContainerName         = "kube-apiserver"
+	controllerManagerContainerName = "kube-controller-manager"
+)
+
+// detectServiceCIDR deserializes the '--service-cluster-ip-range' from the
+// kube-apiserver manifest
+func detectServiceCIDR(config Config) (string, error) {
+	b, err := ioutil.ReadFile(filepath.Join(config.AssetDir, asset.AssetPathAPIServer))
+	if err != nil {
+		return "", fmt.Errorf("can't read file %s: %v", asset.AssetPathAPIServer, err)
+	}
+	var apiServer v1beta1.DaemonSet
+	err = yaml.Unmarshal(b, &apiServer)
+	if err != nil {
+		return "", fmt.Errorf("cant unmarshal %s: %v", asset.AssetPathAPIServer, err)
+	}
+
+	containers := map[string]v1.Container{}
+	for _, container := range apiServer.Spec.Template.Spec.Containers {
+		containers[container.Name] = container
+	}
+
+	if container, exists := containers[apiServerContainerName]; exists {
+		cidr := findFlag("--service-cluster-ip-range", container.Command)
+		if _, _, err := net.ParseCIDR(cidr); err != nil {
+			return "", fmt.Errorf("invalid --cluster-cidr CIDR: %v", err)
+		}
+		return cidr, nil
+	}
+	return "", fmt.Errorf("can't detect --service-cluster-ip-range in %s", asset.AssetPathAPIServer)
+}
+
+// detectPodCIDR deserializes the '--cluster-cidr' from the
+// kube-controller-manager manifest.
+func detectPodCIDR(config Config) (string, error) {
+	b, err := ioutil.ReadFile(filepath.Join(config.AssetDir, asset.AssetPathControllerManager))
+	if err != nil {
+		return "", fmt.Errorf("can't read file %s: %v", asset.AssetPathControllerManager, err)
+	}
+	var manager v1beta1.Deployment
+	err = yaml.Unmarshal(b, &manager)
+	if err != nil {
+		return "", fmt.Errorf("can't unmarshal %s: %v", asset.AssetPathControllerManager, err)
+	}
+
+	containers := map[string]v1.Container{}
+	for _, container := range manager.Spec.Template.Spec.Containers {
+		containers[container.Name] = container
+	}
+
+	if container, exists := containers[controllerManagerContainerName]; exists {
+		cidr := findFlag("--cluster-cidr", container.Command)
+		if _, _, err := net.ParseCIDR(cidr); err != nil {
+			return "", fmt.Errorf("invalid --cluster-cidr CIDR: %v", err)
+		}
+		return cidr, nil
+	}
+	return "", fmt.Errorf("can't detect --cluster-cidr flag in %s", asset.AssetPathControllerManager)
+}
+
+func findFlag(flagName string, args []string) string {
+	for _, arg := range args {
+		if strings.HasPrefix(arg, flagName+"=") {
+			return strings.TrimPrefix(arg, flagName+"=")
+		}
+		if strings.HasPrefix(arg, flagName+" ") {
+			return strings.TrimSpace(strings.TrimPrefix(arg, flagName+" "))
+		}
+	}
+	return ""
+}
--- a/pkg/bootkube/bootkube.go
+++ b/pkg/bootkube/bootkube.go
@@ -47,22 +47,20 @@ func NewBootkube(config Config) (*bootkube, error) {
 	apiServer := apiserver.NewServerRunOptions()
 	fs := pflag.NewFlagSet("apiserver", pflag.ExitOnError)
 	apiServer.AddFlags(fs)
-	fs.Parse(makeAPIServerFlags(config))
+	flags, err := makeAPIServerFlags(config)
+	if err != nil {
+		return nil, err
+	}
+	fs.Parse(flags)
 
 	cmServer := controller.NewCMServer()
 	fs = pflag.NewFlagSet("controllermanager", pflag.ExitOnError)
 	cmServer.AddFlags(fs)
-	fs.Parse([]string{
-		"--master=" + insecureAPIAddr,
-		"--service-account-private-key-file=" + filepath.Join(config.AssetDir, asset.AssetPathServiceAccountPrivKey),
-		"--root-ca-file=" + filepath.Join(config.AssetDir, asset.AssetPathCACert),
-		"--cluster-signing-cert-file=" + filepath.Join(config.AssetDir, asset.AssetPathCACert),
-		"--cluster-signing-key-file=" + filepath.Join(config.AssetDir, asset.AssetPathCAKey),
-		"--allocate-node-cidrs=true",
-		"--cluster-cidr=10.2.0.0/16",
-		"--configure-cloud-routes=false",
-		"--leader-elect=true",
-	})
+	flags, err = makeControllerManagerFlags(config)
+	if err != nil {
+		return nil, err
+	}
+	fs.Parse(flags)
 
 	schedServer := scheduler.NewSchedulerServer()
 	fs = pflag.NewFlagSet("scheduler", pflag.ExitOnError)
@@ -81,8 +79,12 @@ func NewBootkube(config Config) (*bootkube, error) {
 	}, nil
 }
 
-func makeAPIServerFlags(config Config) []string {
-	res := []string{
+func makeAPIServerFlags(config Config) ([]string, error) {
+	serviceCIDR, err := detectServiceCIDR(config)
+	if err != nil {
+		return []string{}, err
+	}
+	return []string{
 		"--bind-address=0.0.0.0",
 		"--secure-port=443",
 		"--insecure-port=8080",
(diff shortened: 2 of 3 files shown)


pkg/bootkube/parse.go
```diff
@@ -0,0 +1,67 @@
...
+// detectServiceCIDR deserializes the '--service-cluster-ip-range' from the
+// kube-apiserver manifest
+func detectServiceCIDR(config Config) string {
+	b, err := ioutil.ReadFile(filepath.Join(config.AssetDir, asset.AssetPathAPIServer))
+	if err != nil {
+		UserOutput("Can't detect --service-cluster-ip-range in %s, using %s: %v\n", asset.AssetPathAPIServer, defaultServiceCIDR, err)
+		return defaultServiceCIDR
+	}
+	var apiServer v1beta1.DaemonSet
+	yaml.Unmarshal(b, &apiServer)
```
aaronlevy: We shouldn't ignore error here. Maybe the manifest is a deployment rather than daemonset (and unmarshal fails). Similar to above we should surface this information to the user.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,67 @@
...
+	defaultPodCIDR     = "10.2.0.0/24"
+)
+
+// detectServiceCIDR deserializes the '--service-cluster-ip-range' from the
+// kube-apiserver manifest
+func detectServiceCIDR(config Config) string {
+	b, err := ioutil.ReadFile(filepath.Join(config.AssetDir, asset.AssetPathAPIServer))
+	if err != nil {
+		UserOutput("Can't detect --service-cluster-ip-range in %s, using %s: %v\n", asset.AssetPathAPIServer, defaultServiceCIDR, err)
+		return defaultServiceCIDR
```
aaronlevy: I wonder if we should just hard fail instead. If we can't detect this information - chances are we are going to get it wrong...

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,67 @@
...
+// detectPodCIDR deserializes the '--cluster-cidr' from the
+// kube-controller-manager manifest.
+func detectPodCIDR(config Config) string {
+	b, err := ioutil.ReadFile(filepath.Join(config.AssetDir, asset.AssetPathControllerManager))
+	if err != nil {
+		UserOutput("Can't detect --cluster-cidr in %s, using %s: %v\n", asset.AssetPathControllerManager, defaultPodCIDR, err)
+		return defaultPodCIDR
+	}
+	var manager v1beta1.Deployment
+	yaml.Unmarshal(b, &manager)
```
aaronlevy: Same as above, don't ignore errors here.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		UserOutput("Can't detect --service-cluster-ip-range in %s, using %s: %v\n", asset.AssetPathAPIServer, defaultServiceCIDR, err)
+		return defaultServiceCIDR
+	}
+	var apiServer v1beta1.DaemonSet
+	yaml.Unmarshal(b, &apiServer)
+
+	containers := apiServer.Spec.Template.Spec.Containers
+	if len(containers) > 0 {
+		return findFlag("--service-cluster-ip-range", containers[0].Command)
```
aaronlevy: We might want to be more specific than just the first container. If we're already relying on fixed file names, I don't think it's much of a leap to rely on fixed container name.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		UserOutput("Can't detect --cluster-cidr in %s, using %s: %v\n", asset.AssetPathControllerManager, defaultPodCIDR, err)
+		return defaultPodCIDR
+	}
+	var manager v1beta1.Deployment
+	yaml.Unmarshal(b, &manager)
+
+	containers := manager.Spec.Template.Spec.Containers
+	if len(containers) > 0 {
+		return findFlag("--cluster-cidr", manager.Spec.Template.Spec.Containers[0].Command)
```
aaronlevy: nit: if you're assigning to `containers` above, may as well use here for consistency

aaronlevy: Also, same comment as above, that might be better to rely on fixed name rather than ordering.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	UserOutput("Can't detect --cluster-cidr, using %s\n", defaultPodCIDR)
+	return defaultPodCIDR
+}
+
+func findFlag(flagName string, args []string) string {
```
aaronlevy: This seems a little fragile. For example, it's not necessarily required that an `=` is between key and value.

Any reason not to just use `FlagSet.Parse(args)` and `FlagSet.Lookup(flagName)` or something similar?

dghubble:
We can introduce error handling to bootkube if you are ok with it being done here.

pkg/bootkube/parse.go
dghubble: They're not args, its the command. We could do it that way as well, if we split off the command name. Ugly either way.

aaronlevy: Hmm, thats a good point. One option might be to iterate through each arg, and once you see one which starts with "-" you can assume the remainder of the slice are arguments which can be passed to flagset.Parse()

aaronlevy:
Yeah I'm fine if the `makeXFlags` functions returned an error (and NewBootkube already returns error).

pkg/bootkube/parse.go
dghubble: fixed

pkg/bootkube/parse.go
dghubble: yup

pkg/bootkube/parse.go
dghubble: making a map with name keys now

pkg/bootkube/parse.go
dghubble: yup

pkg/bootkube/parse.go
dghubble: gotcha

pkg/bootkube/parse.go
dghubble: I'll add this one tmrw

dghubble: Hm, flagsets (both stdlib and spf13) expect all flags to be registered, which means we need to do that stripping and also get the appropriate flags for an apiserver and controller manager (similar to https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/bootkube/bootkube.go#L53). Doesn't seem like a great general flag parser unless I'm missing something.

dghubble: So that could look kinda like this, repeat for apiserver. It allocates an apiserver and controller manager:

```
...

if container, exists := containers[controllerManagerContainerName]; exists {
	cmServer := controller.NewCMServer()
	fs := flag.NewFlagSet("controllermanager", flag.ContinueOnError)
	cmServer.AddFlags(fs)
	fs.Parse(argsOnly(container.Command))
	if f := fs.Lookup("cluster-cidr"); f != nil {
		return f.Value.String(), nil
	}
}
return "", fmt.Errorf("can't detect --cluster-cidr in %s", asset.AssetPathControllerManager)

...

// argsOnly filters non-arguments from a command string.
func argsOnly(cmd []string) []string {
	args := []string{}
	for _, token := range cmd {
		if strings.HasPrefix(token, "-") {
			args = append(args, token)
		}
	}
	return args
}
```

dghubble: This parse also won't work bc its two commands actually.

```
        - /usr/bin/flock
        - --exclusive
        - --timeout=30
        - /var/lock/api-server.lock
        - /hyperkube
        - apiserver
        - --bind-address=0.0.0.0
```

aaronlevy: heh. Might just be best to do a `startsWith("--service-cluster-cidr"`.

dghubble: Using that now and added a test for the `=` vs `" "` case you mentioned.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,78 @@
...
+// kube-apiserver manifest
+func detectServiceCIDR(config Config) (string, error) {
+	b, err := ioutil.ReadFile(filepath.Join(config.AssetDir, asset.AssetPathAPIServer))
+	if err != nil {
+		return "", err
+	}
+	var apiServer v1beta1.DaemonSet
+	err = yaml.Unmarshal(b, &apiServer)
+	if err != nil {
+		return "", err
```
aaronlevy: We should either add error context here or in the caller. This bubbles up all the way to bootkube exit (as you helpfully showed in your PR description) -- some context of what we're trying to do would be helpful.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,78 @@
...
+func findFlag(flagName string, args []string) string {
+	for _, arg := range args {
+		if strings.HasPrefix(arg, flagName+"=") {
+			return strings.TrimPrefix(arg, flagName+"=")
+		}
+		if strings.HasPrefix(arg, flagName) {
+			return strings.TrimSpace(strings.TrimPrefix(arg, flagName))
+		}
+	}
+	return "unknown"
```
aaronlevy: This should be an error. Otherwise we're going to end up using something like `--cluster-cidr=unknown` and waiting for the bootstrap to fail - where we could have caught / surfaced here.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,78 @@
...
+// kube-controller-manager manifest.
+func detectPodCIDR(config Config) (string, error) {
+	b, err := ioutil.ReadFile(filepath.Join(config.AssetDir, asset.AssetPathControllerManager))
+	if err != nil {
+		return "", err
+	}
+	var manager v1beta1.Deployment
+	err = yaml.Unmarshal(b, &manager)
+	if err != nil {
+		return "", err
```
aaronlevy: Add error context

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return "", fmt.Errorf("can't detect --cluster-cidr in %s", asset.AssetPathControllerManager)
+}
+
+func findFlag(flagName string, args []string) string {
+	for _, arg := range args {
+		if strings.HasPrefix(arg, flagName+"=") {
+			return strings.TrimPrefix(arg, flagName+"=")
+		}
+		if strings.HasPrefix(arg, flagName) {
```
aaronlevy: nit: `HasPrefix(arg, flagname+" ")` -- unlikely, but would remove issue of something like `--cluster-cidr-foo=xyz`.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,78 @@
...
+		return "", err
+	}
+
+	containers := map[string]v1.Container{}
+	for _, container := range apiServer.Spec.Template.Spec.Containers {
+		containers[container.Name] = container
+	}
+
+	if container, exists := containers[apiServerContainerName]; exists {
+		return findFlag("--service-cluster-ip-range", container.Command), nil
```
aaronlevy: I think we should also minimally check that we're getting a valid cidr here (net.ParseCIDR() probably enough. Otherwise --cluster-cidr with no value gets through fine.

pkg/bootkube/parse.go
```diff
@@ -0,0 +1,78 @@
...
+		return "", err
+	}
+
+	containers := map[string]v1.Container{}
+	for _, container := range manager.Spec.Template.Spec.Containers {
+		containers[container.Name] = container
+	}
+
+	if container, exists := containers[controllerManagerContainerName]; exists {
+		return findFlag("--cluster-cidr", container.Command), nil
```
aaronlevy: minimally validate that we got a CIDR

pkg/bootkube/parse.go
aaronlevy: Or could still use this, and let the CIDR check fail - but using an arbitrary string here feels wrong

pkg/bootkube/parse.go
dghubble: what do you mean? `--cluster-cidr-foo=xyz` should match the earlier condition and return

pkg/bootkube/parse.go
dghubble: added

pkg/bootkube/parse.go
dghubble: added

pkg/bootkube/parse.go
dghubble: added

pkg/bootkube/parse.go
dghubble: added and updated edge case examples

pkg/bootkube/parse.go
dghubble: It gets caught in the CIDR check and this could just be an empty string to be more conventional

dghubble: fixed

pkg/bootkube/parse.go
ericchiang: Without ` HasPrefix(arg, flagname+" ")` doesn't this match?

```
findFlag("--foo", []string{"--foobar hello"})
```

E.g. what happens if you add the following test case

```
{"--cluster", defaultPodCIDR}, // Matches "--cluster-cidr"
```

dghubble: gotcha

aaronlevy: yeah, sorry. Broken example...

dghubble:
I don't have further changes. I'll implement render options in a followup PR.

aaronlevy:
@dghubble overall this lgtm. Unfortunately the PR tests are currently not working - have you manually verified this works?

dghubble:
Rebased for v1.5.3.

dghubble:
@aaronlevy, I spoke with @pbx0. Tests seem to be running, just not updating the status on Github and this is considered a CI pass. https://jenkins-tectonic.prod.coreos.systems/job/bootkube-dev/314

```
2017-02-17T21:10:24Z kola: --- PASS: bootkube.destruct.delete on gce (406.458s)
2017-02-17T21:10:43Z kola: --- PASS: bootkube.smoke on gce (423.244s)
2017-02-17T21:12:58Z kola: --- FAIL: bootkube.experimentaletcd.smoke on gce (556.723s)
2017-02-17T21:14:16Z kola: --- PASS: bootkube.destruct.reboot on gce (632.535s)
```

The etcd experimental failures are on master.

ericchiang:
code lgtm. @aaronlevy cool to merge or are we blocked on the etcd test?

aaronlevy:
I believe self-hosted etcd isn't working right now - so I'm fine to merge.

PR #309, opened 2017-02-16
hack/tests: update conformance scripts for tls-bootstraping

pbx0:
Forgot to update this when merging the tls bootstraping feature. 

cc @Quentin-M


pbx0:
I think this won't work until we script in an auto-approver. We can copy-pasta some of the current interactive approval script into working as an auto-approver for now or add add the auto-approver pod Quentin is working on. Going to focus on getting the release out before this since I have another way of running conformance tests at the moment.

Quentin-M:
I am adding the approver now.

PR #317, opened 2017-02-17
hack/quickstart: bump bootkube image to v0.3.8

pbx0:


aaronlevy:
Should also bump the kubelet version in `quickstart/kubelet.{master,worker}`

PR #322, opened 2017-02-22
pkg/assets/tls: Update organization name

dghubble:
nit: Organization kube-aws is from the kube-aws era. The old name can cause user confusion.


pbx0:
rktbot run tests

PR #323, opened 2017-02-22
hack/*: GC kubelet rkt images after run.

### LINKED ISSUE #218: GC kubelet rkt images
Restarting the kubelet (via kubelet-wrapper) will create a new container filesystem. Eventually this could fill the disk. We should remove the rkt containers when systemd unit exits.

See change here to resolve same issue: https://github.com/coreos/coreos-kubernetes/commit/ea3bd80f9cc380df67b7c20033b44f1f7c79d5a0

yifan-gu:
This PR changes the cloud-config data in the hack dir so that it
stores the rkt UUID for the kubelet pod, and gc the stopped ones.

Fix #218 

/cc @aaronlevy @derekparker @pbx0


dghubble:
LGTM

PR #331, opened 2017-02-24
pkg/bootkube: use full asset path in log messages

aaronlevy:
/cc @dghubble


pbx0:
rktbot run tests

pbx0:
rktbot run tests

PR #320, opened 2017-02-18
Add support for configmap in checkpointer

dhawal55:
Fixes #124

PR #345, opened 2017-03-01
use a client cert for API server to kubelet communication

ericchiang:
This PR turns on kubelet authentication[0], restricting access to the kubelet API running on port 10250. API servers deployed by bootkube now use a client cert to issue requests to this API (used for exec, logs, and port-forward functionality).

Because kubelets don't have a way to bootstrap a serving certificate with the correct SANs in many deployments (e.g. an autoscaling group), we'll continue to omit the API server CA check on the kubelet. We hope to remedy this later with TLS bootstrapping[1] once the kubelet can request a serving cert.

On the kubelet side, deployments should enable the following flags on the kubelet:

```bash
--client-ca-file=/etc/kubernetes/ca.crt # Placed in assets/tls/ca.crt by 'bootkube render'
--anonymous-auth=false
```

If these flags aren't enabled, the kubelet will continue to allow any request.

Since the checkpointer uses the kubelet API to get the current list of running pods, this PR does not turn on kubelet authorization because authorization requires an available API server. 

[0] https://kubernetes.io/docs/admin/kubelet-authentication-authorization/
[1] https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/

TODO:

- [x] Add flags to API server and kubelet.
- [ ] Update CI to use new kubelet flags.
- [x] Update checkpointer to use kubelet client certs to talk to the kubelet API. (#348)
- [x] Do a checkpointer release and update this PR to use it.
- [x] Add docs about on host requirements.

cc @aaronlevy @pbx0

### DIFF
--- a/hack/quickstart/init-worker.sh
+++ b/hack/quickstart/init-worker.sh
@@ -34,6 +34,9 @@ function init_worker_node() {
     # Setup kubeconfig
     mkdir -p /etc/kubernetes
     cp ${KUBECONFIG} /etc/kubernetes/kubeconfig
+    # Pulled out of the kubeconfig in extract_master_endpoint. Other installations should
+    # place the root CA here manually.
+    cp /home/core/ca.crt /etc/kubernetes/ca.crt
 
     sed "s/{{apiserver}}/${MASTER_PRIV}/" /home/core/kubelet.worker > /etc/systemd/system/kubelet.service
     rm /home/core/kubelet.worker
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -167,6 +167,8 @@ spec:
         - --runtime-config=api/all=true
         - --tls-cert-file=/etc/kubernetes/secrets/apiserver.crt
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
+        - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
+        - --kubelet-client-key=/etc/kubernetes/secrets/apiserver.key
         - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
         - --client-ca-file=/etc/kubernetes/secrets/ca.crt
         - --authorization-mode=RBAC
@@ -216,7 +218,7 @@ spec:
       hostNetwork: true
       containers:
       - name: checkpoint-installer
-        image: quay.io/coreos/pod-checkpointer:5b585a2d731173713fa6871c436f6c53fa17f754
+        image: quay.io/coreos/pod-checkpointer:417b8f7552ccf3db192ba1e5472e524848f0eb5f
         command:
         - /checkpoint-installer.sh
         volumeMounts:
--- a/hack/multi-node/user-data.sample
+++ b/hack/multi-node/user-data.sample
@@ -22,6 +22,8 @@ coreos:
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --kubeconfig=/etc/kubernetes/kubeconfig \
           --require-kubeconfig \
+          --client-ca-file=/etc/kubernetes/ca.crt \
+          --anonymous-auth=false \
           --cni-conf-dir=/etc/kubernetes/cni/net.d \
           --network-plugin=cni \
           --lock-file=/var/run/lock/kubelet.lock \
--- a/hack/quickstart/init-master.sh
+++ b/hack/quickstart/init-master.sh
@@ -52,6 +52,7 @@ function init_master_node() {
     chown -R core:core /home/core/assets
     mkdir -p /etc/kubernetes
     cp /home/core/assets/auth/kubeconfig /etc/kubernetes/
+    cp /home/core/assets/tls/ca.crt /etc/kubernetes/ca.crt
 
     # Start the kubelet
     systemctl enable kubelet; sudo systemctl start kubelet
--- a/hack/quickstart/kubelet.master
+++ b/hack/quickstart/kubelet.master
@@ -15,6 +15,8 @@ ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/run/kubelet-pod.uuid
 ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --api-servers=https://${COREOS_PRIVATE_IPV4}:443 \
   --kubeconfig=/etc/kubernetes/kubeconfig \
+  --client-ca-file=/etc/kubernetes/ca.crt \
+  --anonymous-auth=false \
   --cni-conf-dir=/etc/kubernetes/cni/net.d \
   --network-plugin=cni \
   --lock-file=/var/run/lock/kubelet.lock \
--- a/hack/quickstart/kubelet.worker
+++ b/hack/quickstart/kubelet.worker
@@ -13,6 +13,8 @@ ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/run/kubelet-pod.uuid
 ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --api-servers=https://{{apiserver}}:443 \
   --kubeconfig=/etc/kubernetes/kubeconfig \
+  --client-ca-file=/etc/kubernetes/ca.crt \
+  --anonymous-auth=false \
   --cni-conf-dir=/etc/kubernetes/cni/net.d \
   --network-plugin=cni \
   --lock-file=/var/run/lock/kubelet.lock \
--- a/hack/single-node/user-data.sample
+++ b/hack/single-node/user-data.sample
@@ -30,6 +30,8 @@ coreos:
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --kubeconfig=/etc/kubernetes/kubeconfig \
           --require-kubeconfig \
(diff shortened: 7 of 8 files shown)


ericchiang:
Since this has dependencies on the on-host kubelet, maybe we should feature flag this?

hack/quickstart/init-worker.sh
```diff
@@ -34,6 +34,7 @@ function init_worker_node() {
     # Setup kubeconfig
     mkdir -p /etc/kubernetes
     cp ${KUBECONFIG} /etc/kubernetes/kubeconfig
+    cp /home/core/assets/tls/ca.crt /etc/kubernetes/ca.crt
```
aaronlevy: This won't exist on the worker node - it needs to be copied over (service and kubeconfig are copied here: https://www.pivotaltracker.com/story/show/140894109)

However, I dislike needing to copy another moving part. Right now a worker just needs a kubelet and a valid kubeconfig to be a fully-fledged node.

Just a thought (not sure it's a good idea) - but what about trying to extract CA from the kubeconfig instead (it's already provided in-line)?

pkg/asset/internal/templates.go
```diff
@@ -152,6 +152,8 @@ spec:
         - --runtime-config=api/all=true
         - --tls-cert-file=/etc/kubernetes/secrets/apiserver.crt
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
+        - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver-client.crt
+        - --kubelet-client-key=/etc/kubernetes/secrets/apiserver-client.key
```
aaronlevy: The apiserver cert we have been generating has both client/server key usages - do we really want to separate them? Not necessarily opposed, but it is more moving parts, and I'm not sure how necessary it really is.

pkg/asset/tls.go
```diff
@@ -111,3 +118,19 @@ func newKubeletKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey) (
...
 	return key, cert, err
 }
+
+func newAPIServerClientKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey) (*rsa.PrivateKey, *x509.Certificate, error) {
+	key, err := tlsutil.NewPrivateKey()
+	if err != nil {
+		return nil, nil, err
+	}
+	config := tlsutil.CertConfig{
+		CommonName:   "system:api-server",
```
aaronlevy: Not super familiar with the conventions here - but based on kubelet example (https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/util/csr/csr.go#L40-L43) seems like the CN should be something like `system:foo:api-sever` (but not sure what 'foo' would become). Then also, would we want this assigned to a particular Organization/Group?

pkg/tlsutil/tlsutil.go
```diff
@@ -102,7 +102,7 @@ func ParsePEMEncodedPrivateKey(pemdata []byte) (*rsa.PrivateKey, error) {
 	return x509.ParsePKCS1PrivateKey(decoded.Bytes)
 }
 
-func NewSignedCertificate(cfg CertConfig, key *rsa.PrivateKey, caCert *x509.Certificate, caKey *rsa.PrivateKey) (*x509.Certificate, error) {
+func newSignedCertificate(cfg CertConfig, key *rsa.PrivateKey, caCert *x509.Certificate, caKey *rsa.PrivateKey, usage []x509.ExtKeyUsage) (*x509.Certificate, error) {
```
aaronlevy: nit: `newSignedClientAndServerCertificate` to keep consistent with `newSignedClientCertificate`

aaronlevy:
Were you able to verify this has the intended behavior?

hack/quickstart/init-worker.sh
ericchiang: > Just a thought (not sure it's a good idea) - but what about trying to extract CA from the kubeconfig instead (it's already provided in-line)?

Yeah let's just do that for now.

pkg/asset/internal/templates.go
ericchiang: Sure, let's just reuse the existing cert.

hack/quickstart/init-worker.sh
aaronlevy: Hmm thinking about this more though -- any instructions we provide are still going to have to say "copy over the CA cert", so not sure it's really much benefit to try and parse the kubeconfig in bash (and we're just obscuring the needed steps).

So maybe we just make sure to copy the ca.crt in this step (I had wrong link in previous comment): https://github.com/kubernetes-incubator/bootkube/blob/master/hack/quickstart/init-worker.sh#L54-L55

ericchiang:
@aaronlevy the kubelet is now denying everything, even API server requests. Will look into it more.

ericchiang:
@aaronlevy got this working using matchbox[0] (cc @dghubble). I think I mixed up a render step during my last testing. Need to follow up with @pbx0 about the Jenkins testing. Would be good to add a sanity test where we try to grab logs using kubectl.

From what I can see this change works. There's actually no breaking change because if a user doesn't specify the kubelet flags, it'll behave as before. But I'll follow up with docs about recommend the kubelet host setup.

[0] https://github.com/ericchiang/matchbox/commit/5a32f1a7791e6fa565cd08e40549ee6e683abe85

ericchiang:
> There's actually no breaking change, because if a user doesn't specify the kubelet flags, it'll behave as before.

Actually, if this is true then the CI shouldn't be failing. Let me investigate.

ericchiang:
rktbot run tests

pbx0:
@ericchiang If the host-kubelet needs flags, testing code will still need to change. I plan to address host-kubelet changes causing this problem today.

ericchiang:
Updated the checkpointer to use client certs to talk to the kubelet API. One issue is that this PR will require those changes to the checkpointer, and hence a release. Going to:

* Open another PR to have the checkpointer use client certs.
* Get that merged and do a release of the checkpointer.
* Update this PR to use that release.

pbx0:
rktbot run checkpointer tests will help with that inception problem by first building a checkpointer container from the PR, then it will build a bootkube container as is from the PR but with said checkpointer container included. I'll still need to update CI regardless.

pbx0:
rktbot run checkpointer tests

cmd/checkpoint/main.go
```diff
@@ -60,21 +60,64 @@ func main() {
...
+	if err != nil {
+		glog.Fatalf("Failed to load kubeconfig: %v", err)
+	}
+	client := clientset.NewForConfigOrDie(kubeConfig)
+
+	load := func(rawData []byte, filepath string) []byte {
+		if len(rawData) > 0 {
+			return rawData
+		}
+		data, err := ioutil.ReadFile(filepath)
```
aaronlevy: IIRC the kubelet client certs are in-lined in the kubeconfig (not as files on disk).

Another option is to just extract this info directly from the kubeconfig. An example of getting an http client from client in etcd-operator: Another option is to just 

An example might be how etcd-operator handles this for TPRs: https://github.com/coreos/etcd-operator/blob/master/pkg/util/k8sutil/k8sutil.go#L300

ericchiang: This is extracting it straight from the kubeconfig. I didn't want to assume that the kubeconfig inlined the certs.

aaronlevy: Ah, sorry. Misread

ericchiang:
Tested the checkpointer by deleting the API server and it looks like everything's working.

ericchiang:
Rebased the PR. Per conversation with @aaronlevy I think the updates to the `hack` directory qualify as docs for now (will open other PRs for actual docs). Still blocked on testing.

pkg/asset/internal/templates.go
```diff
@@ -216,7 +218,7 @@ spec:
       hostNetwork: true
       containers:
       - name: checkpoint-installer
-        image: quay.io/coreos/pod-checkpointer:5b585a2d731173713fa6871c436f6c53fa17f754
+        image: quay.io/coreos/pod-checkpointer:e054c5b5a95993baa15ed94e2e2083cc826d4d49
```
aaronlevy: I cut a new release of checkpointer which also contains support for configMaps (https://github.com/kubernetes-incubator/bootkube/pull/320). We could bump to that release here if you're not opposed (we can also wait and bump separately).

aaronlevy:
Overall this looks good to me. Do you mean you're blocked because the CI tests aren't passing (I'll re-trigger)? Or was there more testing you wanted to add/do?

pkg/asset/internal/templates.go
ericchiang: not opposed. Will update

aaronlevy:
rktbot run tests

ericchiang:
@aaronlevy to the best of my knowledge the CI still doesn't actually turn on client cert auth on the kubelet. Was going to wait for that.

aaronlevy:
ah, k.

@pbx0 how far off are we from CI tests consuming the hack scripts for host configuration?

ericchiang:
Updated to use the checkpointer with configmap support.

aaronlevy:
test failure is unrelated and has to do with self-hosted etcd scaling test flaking

PR #348, opened 2017-03-02
cmd/checkpoint: use kubelet's client certs for the kubelet API

ericchiang:
Use the client certs in the kubelet's kubeconfig to talk to the
kubelet API. This has no effect currently, but will once the
kubelet enables API auth. We assume the kubelet's client certs
originate from the same CA that cert auth trusts.

These are checkpointer changes split out from #345 so we can
do a release then import it later once we turn on kubelet auth.

cc @pbx0

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -60,21 +60,64 @@ func main() {
 	}
 
 	glog.Infof("Starting checkpointer for node: %s", nodeName)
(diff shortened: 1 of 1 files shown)


pbx0:
rktkbot run checkpointer tests

pbx0:
rktbot run checkpointer tests

ericchiang:
@pbx0 all green!

PR #350, opened 2017-03-03
support self-hosted etcd in quickstart scripts

pbx0:
Add support for self-hosted etcd within master-init script. Simplify service files and init-worker script a bit which makes them easier to configure and re-use by other tools. 

cc @aaronlevy

### DIFF
--- a/hack/quickstart/init-master.sh
+++ b/hack/quickstart/init-master.sh
@@ -6,6 +6,7 @@ REMOTE_PORT=${REMOTE_PORT:-22}
 CLUSTER_DIR=${CLUSTER_DIR:-cluster}
 IDENT=${IDENT:-${HOME}/.ssh/id_rsa}
 SSH_OPTS=${SSH_OPTS:-}
+SELF_HOST_ETCD=${SELF_HOST_ETCD:-false}
 
 BOOTKUBE_REPO=${BOOTKUBE_REPO:-quay.io/coreos/bootkube}
 BOOTKUBE_VERSION=${BOOTKUBE_VERSION:-v0.3.9}
@@ -37,16 +38,25 @@ function init_master_node() {
     systemctl daemon-reload
     systemctl stop update-engine; systemctl mask update-engine
 
+    etcd_render_flags=""
+    etcd_start_flags=""
+
     # Start etcd.
-    configure_etcd
-    systemctl enable etcd-member; sudo systemctl start etcd-member
+    if [ "$SELF_HOST_ETCD" = true ] ; then
+        echo "WARNING: THIS IS NOT YET FULLY WORKING - merely here to make ongoing testing easier"
+        etcd_render_flags="--etcd-servers=http://10.3.0.15:2379 --experimental-self-hosted-etcd"
+        etcd_start_flags="--etcd-server=http://${COREOS_PRIVATE_IPV4}:12379 --experimental-self-hosted-etcd"
+    else
+        configure_etcd
+        systemctl enable etcd-member; sudo systemctl start etcd-member
+    fi
 
     # Render cluster assets
     /usr/bin/rkt run \
         --volume home,kind=host,source=/home/core \
         --mount volume=home,target=/core \
         --trust-keys-from-https --net=host ${BOOTKUBE_REPO}:${BOOTKUBE_VERSION} --exec \
-        /bootkube -- render --asset-dir=/core/assets --api-servers=https://${COREOS_PUBLIC_IPV4}:443,https://${COREOS_PRIVATE_IPV4}:443
+        /bootkube -- render --asset-dir=/core/assets --api-servers=https://${COREOS_PUBLIC_IPV4}:443,https://${COREOS_PRIVATE_IPV4}:443 ${etcd_render_flags}
 
     # Move the local kubeconfig into expected location
     chown -R core:core /home/core/assets
@@ -61,8 +71,10 @@ function init_master_node() {
     /usr/bin/rkt run \
         --volume home,kind=host,source=/home/core \
         --mount volume=home,target=/core \
+        --volume manifests,kind=host,source=/etc/kubernetes/manifests \
+        --mount volume=manifests,target=/etc/kubernetes/manifests \
         --net=host ${BOOTKUBE_REPO}:${BOOTKUBE_VERSION} --exec \
-        /bootkube -- start --asset-dir=/core/assets
+        /bootkube -- start --asset-dir=/core/assets ${etcd_start_flags}
 }
 
 [ "$#" == 1 ] || usage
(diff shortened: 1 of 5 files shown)


hack/quickstart/init-master.sh
```diff
@@ -38,15 +39,26 @@ function init_master_node() {
...
+            --mount volume=home,target=/core \
+            --trust-keys-from-https --net=host ${BOOTKUBE_REPO}:${BOOTKUBE_VERSION} --exec \
+            /bootkube -- render --asset-dir=/core/assets --api-servers=https://${COREOS_PUBLIC_IPV4}:443,https://${COREOS_PRIVATE_IPV4}:443 \
+            --etcd-servers=http://10.3.0.15:2379  --experimental-self-hosted-etcd
+    else
+        /usr/bin/rkt run \
+            --volume home,kind=host,source=/home/core \
+            --mount volume=home,target=/core \
+            --trust-keys-from-https --net=host ${BOOTKUBE_REPO}:${BOOTKUBE_VERSION} --exec \
+            /bootkube -- render --asset-dir=/core/assets --api-servers=https://${COREOS_PUBLIC_IPV4}:443,https://${COREOS_PRIVATE_IPV4}:443
```
aaronlevy: Most of this is the same between both cases, the only differences is the flags. Could simplify just by parameterizing the flags themselves (e.g. https://github.com/kubernetes-incubator/bootkube/blob/master/hack/multi-node/bootkube-up#L12-L20)

hack/quickstart/init-master.sh
```diff
@@ -57,11 +69,20 @@ function init_master_node() {
...
+            --net=host ${BOOTKUBE_REPO}:${BOOTKUBE_VERSION} --exec \
+            /bootkube -- start --asset-dir=/core/assets \
+            --etcd-server=http://${COREOS_PRIVATE_IPV4}:12379 --experimental-self-hosted-etcd
+    else
+        /usr/bin/rkt run \
+            --volume home,kind=host,source=/home/core \
+            --mount volume=home,target=/core \
+            --net=host ${BOOTKUBE_REPO}:${BOOTKUBE_VERSION} --exec \
+            /bootkube -- start --asset-dir=/core/assets
+    fi
```
aaronlevy: Same comment as above - all the flags could be set in one place and we don't need these if/then blocks

hack/quickstart/init-master.sh
```diff
@@ -38,15 +39,23 @@ function init_master_node() {
...
 
     # Start etcd.
-    configure_etcd
-    systemctl enable etcd-member; sudo systemctl start etcd-member
+    if [ "$SELF_HOST_ETCD" = false ] ; then
+        configure_etcd
+        systemctl enable etcd-member; sudo systemctl start etcd-member
+    fi
+
+    if [ ${SELF_HOST_ETCD} = "true" ]; then
```
aaronlevy: nit: combine this and the baove block into a single if/else

hack/quickstart/init-master.sh
```diff
@@ -38,15 +39,21 @@ function init_master_node() {
     systemctl stop update-engine; systemctl mask update-engine
 
     # Start etcd.
-    configure_etcd
-    systemctl enable etcd-member; sudo systemctl start etcd-member
+    if [ "$SELF_HOST_ETCD" = false ] ; then
+        configure_etcd
+        systemctl enable etcd-member; sudo systemctl start etcd-member
+	else
```
aaronlevy: nit: fix indent

pbx0:
Fixed indent. DO NOT MERGE -- Need to test this a little more.

pbx0:
@aaronlevy PTAL. I've tested with self-hosted and non-self hosted via conformance script.

hack/quickstart/init-worker.sh
```diff
@@ -13,33 +13,18 @@ function usage() {
...
-    extract_master_endpoint
 
     # Setup kubeconfig
     mkdir -p /etc/kubernetes
     cp ${KUBECONFIG} /etc/kubernetes/kubeconfig
-    # Pulled out of the kubeconfig in extract_master_endpoint. Other installations should
-    # place the root CA here manually.
+    # Pulled out of the kubeconfig. Other installations should place the root
+    # CA here manually.
+    grep 'certificate-authority-data' ${KUBECONFIG} | awk '{print $2}' | base64 -d > /home/core/ca.crt
```
aaronlevy: If we just output this to `/etc/kubernetes/ca.crt` we can drop the `mv` step below

hack/tests/conformance-gce.sh
```diff
@@ -46,7 +48,7 @@ function init {
     curl https://sdk.cloud.google.com | bash
     source ~/.bashrc
     gcloud config set project coreos-gce-testing
-    gcloud auth activate-service-account ${GCE_PREFIX}@coreos-gce-testing.iam.gserviceaccount.com --key-file=/build/keyfile
+    gcloud auth activate-service-account ${GCE_ACCOUNT}@coreos-gce-testing.iam.gserviceaccount.com --key-file=/build/keyfile
```
aaronlevy: I think this was mean to replace the section after the `@`

hack/tests/conformance-gce.sh
```diff
@@ -32,6 +32,8 @@ BOOTKUBE_VERSION=${BOOTKUBE_VERSION:-}
 COREOS_CHANNEL=${COREOS_CHANNEL:-'coreos-stable'}
 WORKER_COUNT=4
 GCE_PREFIX=${GCE_PREFIX:-'bootkube-ci'}
+GCE_ACCOUNT=${GCE_ACCOUNT:-'bootkube-ci'}
```
aaronlevy: Why bootkube-ci here? Is this just placeholder?

pbx0: Just preserving what was here. 

It appears that we have a service account titled: `bootkube-ci@coreos-gce-testing.iam.gserviceaccount.com` That we don't want to change with the instance prefix change because it refers to a specific account. The default instance prefix just happened to be the same as the service account but if we actually want to chance the instance prefix we shouldn't be changing the account as well.

hack/quickstart/init-worker.sh
pbx0: done

pbx0:
Please wait to merge -- Going to run one last run of tests with the latest update.

pbx0:
@aaronlevy PTAL: I've fixed the gce specific options a bit and re-ran tests. Self-hosted and normal bootkube both pass all conformance tests using the scripts.

hack/tests/conformance-gce.sh
```diff
@@ -45,8 +49,8 @@ function cleanup {
 function init {
     curl https://sdk.cloud.google.com | bash
     source ~/.bashrc
-    gcloud config set project coreos-gce-testing
-    gcloud auth activate-service-account ${GCE_PREFIX}@coreos-gce-testing.iam.gserviceaccount.com --key-file=/build/keyfile
+    gcloud config set project ${GCE_PROJECT}
+    gcloud auth activate-service-account ${GCE_SERVICE_ACCOUNT}@coreos-gce-testing.iam.gserviceaccount.com --key-file=/build/keyfile
```
pbx0: done, but I think you meant ${GCE_PROJECT}

hack/quickstart/init-master.sh
```diff
@@ -37,16 +38,25 @@ function init_master_node() {
     systemctl daemon-reload
     systemctl stop update-engine; systemctl mask update-engine
 
+    etcd_render_flags=""
+    etcd_start_flags=""
+
     # Start etcd.
-    configure_etcd
-    systemctl enable etcd-member; sudo systemctl start etcd-member
+    if [ "$SELF_HOST_ETCD" = false ] ; then
```
pbx0: makes sense, done

pbx0:
rktbot run tests

pbx0:
rktbot run tests

pbx0:
@aaronlevy PTAL -- I tested once more to make sure that options are plumbed through correctly with the logic change

PR #356, opened 2017-03-07
SelfHostedEtcd: Add kenc as DS in master nodes

hongchaodeng:
`kenc` is an iptables checkpoint tool.
With it, service IP connectivity on master nodes can endure reboot.

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -200,6 +200,50 @@ spec:
         hostPath:
           path: /var/lock
 `)
+
+	KencTemplate = []byte(`apiVersion: "extensions/v1beta1"
+kind: DaemonSet
+metadata:
+  name: kenc
+  namespace: kube-system
+  labels:
+    k8s-app: kenc
+spec:
+  template:
+    metadata:
+      labels:
+        k8s-app: kenc
+      annotations:
+        checkpointer.alpha.coreos.com/checkpoint: "true"
+    spec:
+      nodeSelector:
+        master: "true"
+      hostNetwork: true
+      containers:
+      - image: quay.io/coreos/kenc:82343328b867a762ffca07c2877f0079a99c8f1a
+        name: kenc
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - mountPath: /etc/kubernetes/selfhosted-etcd
+          name: checkpoint-dir
+          readOnly: false
+        - mountPath: /var/lock
+          name: var-lock
+          readOnly: false
+        command:
+        - "/bin/sh"
+        - "-c"
+        - "/usr/bin/flock --exclusive --timeout=30 /var/lock/kenc.lock; kenc -r -m iptables && kenc -m iptables"
+      volumes:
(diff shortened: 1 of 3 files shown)


hongchaodeng:
Needed:
- [ ] enable disruption test suite to test this PR

pkg/asset/internal/templates.go
```diff
@@ -198,6 +198,43 @@ spec:
...
+      labels:
+        k8s-app: kenc
+      annotations:
+        checkpointer.alpha.coreos.com/checkpoint: "true"
+    spec:
+      nodeSelector:
+        master: "true"
+      hostNetwork: true
+      containers:
+      - image: gcr.io/coreos-k8s-scale-testing/kenc
```
aaronlevy: Prior to merging we should have a quay/coreos repo + release tag

pkg/asset/internal/templates.go
```diff
@@ -198,6 +198,43 @@ spec:
...
+        name: kenc
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - mountPath: /etc/kubernetes/selfhosted-etcd
```
aaronlevy: For pod checkpointing we use:

`/etc/kubernetes/checkpoint-configmaps`
`/etc/kubernetes/checkpoint-secrets`

Maybe for consistency we use `/etc/kubernetes/checkpoint-iptables`?

hongchaodeng: There is an option to configure it. Let me make it.

hongchaodeng: Actually we only need to change hostPath..

aaronlevy:
We may want to consider adding a flock around the execution of this pod - unless you're sure it's safe to potentially have two running at once. The checkpointing process does not guarantee that parent/child have some overlap. Example with api-server: https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L151-L155

hongchaodeng:
> The checkpointing process does not guarantee that parent/child have some overlap

What does it mean?

aaronlevy:
Sorry, not very helpful of description:

Both the original pod and the checkpointed copy of that pod could end up running at the same time. The checkpointer does not enforce that only one is running at a time, and there likely will be overlap.

hongchaodeng:
@aaronlevy 

Thanks for explaining. I think it's best practice for all such pods to do "flock", although for kenc overlapping doesn't matter. I will change it shortly.

hongchaodeng:
Added flock

hongchaodeng:
@aaronlevy 
Any other comment on this?

aaronlevy:
lgtm. Is this ready to be merged?

hongchaodeng:
@aaronlevy 

Yes. Ready for merge. We have also tested it.

Also see https://github.com/kubernetes-incubator/bootkube/pull/356#issuecomment-284857226 .
Enabling destruction test will be delayed because we need etcd operator to do a new release.

aaronlevy:
Gonna wait for PR test before merge

PR #360, opened 2017-03-09
quickstart: fix for ca.crt

aaronlevy:


aaronlevy:
rktbot run tests

pbx0:
Is there anyway to delete the `extract_master_endpoint` function entirely here and copy over the cert generated in the assets?

aaronlevy:
We still need the kubeconfig - so we could remove most of that function and just extract the ca.crt

aaronlevy:
But this is a fix because HEAD is currently broken - could do that in a separate PR (and I saw you had that in your hack+self-hosted etcd pr too)

aaronlevy:
@pbx0 it's newly necessary to keep the extracted CA because we're using it to turn on AuthN to the kubelet api

ericchiang:
FYI these changes worked using the AWS quickstart manually.

Issue #253, opened 2016-12-30
checkpointer: should GC itself if installer no longer scheduled


aaronlevy:
The checkpointer is deployed via a daemonset which "installs" a static manifest to the host.

If the checkpoint-installer is no longer scheduled to a node, the checkpointer should know how to GC itself (and all checkpoints).

yifan-gu:
Today I have an idea about using the checkpointer to checkpoint itself so that we can get rid of the checkpointer installer.

Here is what it will look like in several scenarios:

### Scenario A: checkpoint pod get scheduled on node A as a daemonset
_t0:_ Checkpointer (call it C1) gets running
_t1:_ C1 checkpoint *and activate* itself by reading its own spec and saving to `/etc/kubernetes/manifests`
_t2:_ The static pod version of checkpoint starts (call it C2), C2 will wait on a flock held by C1 to avoid race.
_t3:_ C1 runs in a loop to keep creating checkpoints for other pods if necessary.

### Scenario B: node A reboots, but cannot reach API server
_t0:_ Kubelet starts
_t1:_ The static pod version of checkpoint (C2) starts and grabs the flock
_t2:_ C2 starts running the same loop as C1, and kicks off other checkpointed pods.

### Scenario C: node A reached API server
_t0:_ Daemonset version of checkpoint (C1) gets scheduled and started on node A
_t1:_ C1 waits on a flock held by C2
_t2:_ C2 keeps running [1]
_t3:_ If C1 gets descheduled from node A, then C2 will find it out and remove all checkpoints

During [1], if checkpoint's spec is changed on the API server, then C2 will checkpoint the new spec, and be restarted by kubelet. So the running and on-disk checkpoint specs are always the latest.

This is actually not very different from the `checkpoint installer + checkpointer` model. But by using only one image/pod, we can have the checkpoint to get its own pod spec from API server and get rid of the hidden manifest (see #206 )

The imperfect part is that user than will see 2 checkpointer pods running on each master node, with one being active, and one being stand-by.

aaronlevy:
I think this is a really interesting idea!

As far as the 2 checkpointer pods - really we already have this problem. There is a "checkpoint-installer" and "pod-checkpointer" on every node.

[...]

PR #366, opened 2017-03-10
Checkpoint and activate itself.

yifan-gu:
This PR enables us to GC itself using the existing codepath.
Also it removes the needs for the checkpoint-installer, which
also enables us to update the checkpointer by just updating the
checkpointer's manifest.

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -240,10 +240,10 @@ func process(localRunningPods, localParentPods, apiParentPods, activeCheckpoints
 			if _, ok := apiParentPods[id]; !ok {
 				glog.V(4).Infof("API GC: should remove inactive checkpoint %s", id)
 
+				removeMap[id] = struct{}{}
 				if isPodCheckpointer(inactiveCheckpoints[id]) {
 					podCheckpointerID = id
-				} else {
-					removeMap[id] = struct{}{}
+					break
 				}
 
 				delete(inactiveCheckpoints, id)
@@ -256,14 +256,26 @@ func process(localRunningPods, localParentPods, apiParentPods, activeCheckpoints
 			if _, ok := apiParentPods[id]; !ok {
 				glog.V(4).Infof("API GC: should remove active checkpoint %s", id)
 
+				removeMap[id] = struct{}{}
 				if isPodCheckpointer(activeCheckpoints[id]) {
 					podCheckpointerID = id
-				} else {
-					removeMap[id] = struct{}{}
+					break
 				}
 
 				delete(activeCheckpoints, id)
 			}
+		}
+	}
+
+	// Remove all checkpoints if we need to remove the pod checkpointer itself.
+	if podCheckpointerID != "" {
+		for id := range inactiveCheckpoints {
+			removeMap[id] = struct{}{}
+			delete(inactiveCheckpoints, id)
+		}
+		for id := range activeCheckpoints {
+			removeMap[id] = struct{}{}
+			delete(activeCheckpoints, id)
 		}
 	}
 
@@ -293,8 +305,12 @@ func process(localRunningPods, localParentPods, apiParentPods, activeCheckpoints
 
 	// De-duped checkpoints to remove. If we decide to GC a checkpoint, we will clean up both inactive/active.
 	for k := range removeMap {
+		if k == podCheckpointerID {
+			continue
+		}
 		remove = append(remove, k)
 	}
+	// Put pod checkpoint at the last of the queue.
 	if podCheckpointerID != "" {
 		remove = append(remove, podCheckpointerID)
 	}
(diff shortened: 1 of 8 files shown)


yifan-gu:
Fix #253
Fix #206

Ready for a review.

I did several tests manually:

1. 
- Boot up 2 nodes.
- Schedule checkpointer on both.
- Schedule a test pod (checkpoints enabled) on the worker node.
- See 4 checkpointers, 2 running, 2 standby.
- Suspend the master, reboot the worker node.
- The standby checkpointer starts, which then starts checkpoints.
- Resume the master, see 4 checkpointers, 2 running, 2 standby, and the static checkpointer now is the effective one on the worker node.
- Checkpointer stops the test pod's checkpoints after the test pod is running.

2.
- Boot up 2 nodes.
- Schedule checkpointer on both.
- Schedule a test pod (checkpoints enabled) on the worker node.
- See 4 checkpointers, 2 running, 2 standby.
- Suspend the worker, delete the test pod, recreate the checkpointer daemonset so it only schedules on master node.
- Suspend the master
- Resume/reboot the worker
- The standby checkpointer starts, which then starts checkpoints.
- Resume the master
- The standby checkpointer cleans the checkpoints and itself.




/cc @aaronlevy @stuart-warren

cmd/checkpoint/main.go
```diff
@@ -295,6 +351,11 @@ func writeCheckpointManifest(pod *v1.Pod) error {
 	return writeAndAtomicRename(path, b, 0644)
 }
 
+// isPodCheckpointer returns true if the pod is the checkpointer itself.
+func isPodCheckpointer(pod *v1.Pod) bool {
+	return strings.HasPrefix(pod.Name, "pod-checkpointer")
```
aaronlevy: This seems a bit fragile. Is there another way we could determine this? We can know our "self" via the downward api + inject as env vars - maybe that's an option?

cmd/checkpoint/main.go
```diff
@@ -497,29 +558,35 @@ func checkpointConfigMap(client clientset.Interface, namespace, podName, configM
...
+		}
+
+		// Remove active checkpoints.
+		// We do this as the last step because we want to clean up
+		// resources before the checkpointer itself exits.
+		p = PodFullNameToActiveCheckpointPath(id)
+		if err := os.Remove(p); err != nil && !os.IsNotExist(err) {
+			glog.Errorf("Failed to remove active checkpoint %s: %v", p, err)
+			continue
+		}
```
aaronlevy: I'm a slightly worried about swapping the order of removal here. My thought being that we should stop the pods first, then remove the assets they rely on. It might not be an issue - because we weren't waiting for the pods to actually stop - but do you forsee issues in removing a secret / configMap for example from underneath a running pod?

cmd/checkpoint/main_test.go
```diff
@@ -104,6 +104,22 @@ func TestProcess(t *testing.T) {
 			activeCheckpoints: map[string]*v1.Pod{"AA": {}},
 			localParents:      map[string]*v1.Pod{"AA": {}},
 		},
+		{
+			desc:                "Inactive pod-checkpointer, local parent, local running, api parent: should start",
```
aaronlevy: May be worth adding another test to make sure it still starts even with a local parent (the checkpointed copy always needs to run)

cmd/checkpoint/main.go
yifan-gu: I tested this, it actually successfully removed without an issue. But I want to just double check and understand why it's not an issue and report back.

Alternatively, we can keep the order. and just leave the configmap/scret for the checkpointer untouched. And next time we create the configmap/scret again, we always overwrite the existing one. This seems safer and doesn't cause any issue except for two small files. WDYT?

yifan-gu: cc @aaronlevy ^

cmd/checkpoint/main.go
yifan-gu: @aaronlevy Which downward API? Getting the podname?

cmd/checkpoint/main_test.go
yifan-gu: Sure

cmd/checkpoint/main.go
yifan-gu: So actually removing the config, secret will always succeed, it's not a unmounting operaton.
But maybe that's not graceful to pods as they might have some prestop hooks or whatever graceful termination process that will require the secrets/configs.
I will revert this ordering change.

yifan-gu: Had more discussion on this. As for now we are not waiting the pod to be deleted anyway, so maybe it's fine to leave as is, and improve this later.

cmd/checkpoint/main.go
aaronlevy: Yeah. We could probably check that maybe:

```
pod.Name == self.Namespace
pod.Name == self.Name + "-" + self.NodeName // Because we care about the static pod, right?
```

aaronlevy: Hmm actually, it won't have the node-name suffix in the checkpoint manifest - so maybe just `pod.Name == self.Name` if we care about the parent pod

or, if we care about the checkpointed copy, something like:

```
pod.Name == strings.TrimSuffix(self.Name, self.NodeName) && pod.Name != self.Name
```

yifan-gu:
Comments addressed, PTAL @aaronlevy

aaronlevy:
This LGTM - but going to hold off on merging this to coordinate these changes at same time as k8s v1.6.0

yifan-gu:
Add one more commit for cleaning up all checkpoints when the pod checkpointer is removed.
Will squash if this passes the review @aaronlevy

PR #376, opened 2017-03-13
*: fix etcd and kenc container command

hongchaodeng:
1. get rid of `sh`
2. fix `flock` command

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -232,9 +232,10 @@ spec:
           name: var-lock
           readOnly: false
         command:
-        - "/bin/sh"
-        - "-c"
-        - "/usr/bin/flock --exclusive --timeout=30 /var/lock/kenc.lock; kenc -r -m iptables && kenc -m iptables"
+        - /usr/bin/flock
+        - /var/lock/kenc.lock
+        - -c
+        - "kenc -r -m iptables && kenc -m iptables"
       volumes:
       - name: checkpoint-dir
         hostPath:
--- a/pkg/util/etcdutil/start_etcd.go
+++ b/pkg/util/etcdutil/start_etcd.go
@@ -45,18 +45,16 @@ metadata:
 spec:
   containers:
   - command:
-    - /bin/sh
-    - -c
     - /usr/local/bin/etcd
-      --name boot-etcd
-      --listen-client-urls=http://0.0.0.0:12379
-      --listen-peer-urls=http://0.0.0.0:12380
-      --advertise-client-urls=http://$(MY_POD_IP):12379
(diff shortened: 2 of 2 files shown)


hongchaodeng:
@xiang90 @aaronlevy

pkg/util/etcdutil/start_etcd.go
```diff
@@ -46,7 +46,7 @@ spec:
   containers:
   - command:
     - /bin/sh
-    - -c
+    - -ec
```
aaronlevy: Why do we need to wrap this in a shell here? I think we could execute this directly. FWIW when you use $(FOO) the kubelet will template in those values (they are not being used as environment variables - note the `"()"` vs `"{}"`

pkg/asset/internal/templates.go
```diff
@@ -233,8 +233,8 @@ spec:
           readOnly: false
         command:
         - "/bin/sh"
-        - "-c"
-        - "/usr/bin/flock --exclusive --timeout=30 /var/lock/kenc.lock; kenc -r -m iptables && kenc -m iptables"
+        - "-ec"
```
aaronlevy: Aso doesn't seem like this needs to be wrapped in a shell. Can't we just execute flock directly?

pkg/util/etcdutil/start_etcd.go
hongchaodeng: Just my pick. I usually wrap everything in `sh -ec`. Then the whole command could be run in a single string.

I can change if you want.

aaronlevy: Yeah, if we don't need separate shell - seems simpler to just execute the binary directly.

hongchaodeng: Also the `$(..)` templating still works under `sh -ec`

hongchaodeng:
Updated

pbx0:
@hongchaodeng The change to the boot-etcd.yaml is causing things to fail. I'm not exactly sure why but the docker logs for that container says:
```
flag provided but not defined: -name boot-etcd
usage: etcd [flags]
       start an etcd server

       etcd --version
       show the version of etcd

       etcd -h | --help
       show the help information about etcd

       etcd --config-file
       path to the server configuration file
```

hongchaodeng:
fixed

pbx0:
rktbot run tests

hongchaodeng:
rktbot run tests

pbx0:
@hongchaodeng The CI tests look good to me. The self-hosted scale test is struggling these days, but the other tests all pass including the self-hosted smoke.

hongchaodeng:
@aaronlevy PTAL

aaronlevy:
lgtm

PR #381, opened 2017-03-14
Disable --insecure-port flag for api-server

### LINKED ISSUE #324: api-server: stop listening with insecure port
As discussed we should stop doing this with the API as we don't use it anymore.

cc @aaronlevy

jamiehannaford:
Fixes #324

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -162,28 +162,29 @@ spec:
         - /hyperkube
         - apiserver
         - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota
+        - --bind-address=0.0.0.0
+        - --secure-port=443
+        - --insecure-port=0
         - --advertise-address=$(POD_IP)
         - --allow-privileged=true
         - --anonymous-auth=false
-        - --authorization-mode=RBAC
-        - --bind-address=0.0.0.0
-        - --client-ca-file=/etc/kubernetes/secrets/ca.crt
-        - --cloud-provider={{ .CloudProvider  }}
 {{- if .EtcdUseTLS }}
         - --etcd-cafile=/etc/kubernetes/secrets/etcd-ca.crt
         - --etcd-certfile=/etc/kubernetes/secrets/etcd-client.crt
         - --etcd-keyfile=/etc/kubernetes/secrets/etcd-client.key
 {{- end }}
         - --etcd-servers={{ range $i, $e := .EtcdServers }}{{ if $i }},{{end}}{{ $e }}{{end}}
-        - --insecure-port=8080
+        - --client-ca-file=/etc/kubernetes/secrets/ca.crt
+        - --tls-ca-file=/etc/kubernetes/secrets/ca.crt
+        - --tls-cert-file=/etc/kubernetes/secrets/apiserver.crt
+        - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
         - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
         - --kubelet-client-key=/etc/kubernetes/secrets/apiserver.key
-        - --secure-port=443
         - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
         - --service-cluster-ip-range={{ .ServiceCIDR }}
         - --storage-backend=etcd3
-        - --tls-cert-file=/etc/kubernetes/secrets/apiserver.crt
-        - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
+        - --authorization-mode=RBAC
+        - --cloud-provider={{ .CloudProvider }}
         env:
         - name: POD_IP
           valueFrom:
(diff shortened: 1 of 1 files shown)


pbx0:
ok to test

pbx0:
rktbot run tests

pbx0:
rktbot run tests

aaronlevy:
I don't think removing the flag changes any behavior (the default is 8080). IIRC you could set it to `0` to disable (once upon a time) but this would need to be re-verified.

jamiehannaford:
So, for some weird reason, disabling the insecure port prevents any traffic from reaching kube-apiserver:

```
$ kubectl --kubeconfig=cluster/auth/kubeconfig get nodes
The connection to the server 172.17.4.100:443 was refused - did you specify the right host or port?
```

I initially thought that it was related to https://github.com/kubernetes/kubernetes/issues/13598 (the impossibility of disabling `--insecure-port`) but kube-apiserver _should_ allow you to disable it (see https://github.com/kubernetes/kubernetes/pull/31491). I also tried explicitly adding the `--kubeconfig` and `--master` flags to the controller manifest, but that didn't really fix anything. I'm not sure what's causing the inaccessibility... 

I SSHed into the vagrant box and copied the logs from the apiserver docker container: https://gist.github.com/jamiehannaford/6c3881860ffcd73d9c7d950a7468a100

jamiehannaford:
I also don't have access to the `bootkube-dev` CI build, but I assume it's the same problem as what's happening on my local box.

jamiehannaford:
@aaronlevy This is RFR again. Managed to get all tests passing.

pkg/asset/internal/templates.go
```diff
@@ -299,24 +300,31 @@ spec:
...
         - --configure-cloud-routes=false
         - --cluster-cidr={{ .PodCIDR }}
-        - --root-ca-file=/etc/kubernetes/secrets/ca.crt
-        - --service-account-private-key-file=/etc/kubernetes/secrets/service-account.key
+        - --root-ca-file=/etc/secrets/ca.crt
+        - --service-account-private-key-file=/etc/secrets/service-account.key
         - --leader-elect=true
-        - --cloud-provider={{ .CloudProvider  }}
+        - --cloud-provider={{ .CloudProvider }}
+        - --kubeconfig=/etc/kubernetes/kubeconfig
```
aaronlevy: This shouldn't be needed -- the controller-manager will use the service account injected into the pod -- and shouldn't require access to host files.

pkg/asset/internal/templates.go
```diff
@@ -157,22 +157,23 @@ spec:
...
         - --secure-port=443
-        - --insecure-port=8080
+        - --insecure-port=0
         - --advertise-address=$(POD_IP)
         - --etcd-servers={{ range $i, $e := .EtcdServers }}{{ if $i }},{{end}}{{ $e }}{{end}}
         - --storage-backend=etcd3
         - --allow-privileged=true
         - --service-cluster-ip-range={{ .ServiceCIDR }}
-        - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota
+        - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota
```
aaronlevy: We want this change -- but I don't think this is required for disabling the insecure port and we should move to a diff PR if so.

pkg/asset/tls.go
```diff
@@ -82,6 +82,7 @@ func newAPIKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey, altNa
 		"kubernetes.default",
 		"kubernetes.default.svc",
 		"kubernetes.default.svc.cluster.local",
+		"localhost",
```
aaronlevy: Will the admission controller only use localhost? I'd prefer not to have this as a SAN if possible

pkg/asset/internal/templates.go
jamiehannaford: I think it is needed because when removed the tests are failing. Although the service account should be used, it doesn't seem to be. Also, kubeadm mounts the host /etc/kubernetes dir in order to provide the `--kubeconfig` flag. See:

https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/master/manifests.go#L393
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/master/manifests.go#L238-L253
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/master/manifests.go#L96

pkg/asset/tls.go
jamiehannaford: Hmm. I remember seeing a bunch of SAN errors because _something_ was connecting with TLS over localhost. For the life of me I can't remember what it was though 😅  I'll try to dig deeper on this.

pkg/asset/internal/templates.go
jamiehannaford: Sure, I'll move it out

diegs: n.b. https://github.com/kubernetes-incubator/bootkube/pull/447

jamiehannaford:
@aaronlevy Thanks for the review. I addressed all those things you mentioned.

jamiehannaford:
Closing this since #459 has been merged instead.

PR #385, opened 2017-03-15
Use `podAntiAffinity` to schedule control plane pods to different nodes.

adieu:
Without `podAntiAffinity`, the scheduler might schedule `kube-scheduler` and `kube-controller-manager` to the same node. When this node is down, we might lose all `kube-scheduler` or `kube-controller-manager` at once. The cluster will have problem recover from this failure.

With this patch, the scheduler will schedule control plane pods into different nodes making the cluster more reliable.

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -282,9 +282,29 @@ metadata:
   labels:
     k8s-app: kube-controller-manager
 spec:
-  replicas: 2
+  replicas: 1
   template:
     metadata:
+      annotations:
+        scheduler.alpha.kubernetes.io/affinity: |
+          {
+            "podAntiAffinity": {
+              "requiredDuringSchedulingIgnoredDuringExecution": [
+                {
+                  "labelSelector": {
+                    "matchExpressions": [
+                      {
+                        "key": "k8s-app",
+                        "operator": "In",
+                        "values": ["kube-controller-manager"]
+                      }
+                    ]
+                  },
+                  "topologyKey": "kubernetes.io/hostname"
+                }
+              ]
+            }
+          }
       labels:
         k8s-app: kube-controller-manager
     spec:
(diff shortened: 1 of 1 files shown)


pkg/asset/internal/templates.go
```diff
@@ -282,9 +282,29 @@ metadata:
   labels:
     k8s-app: kube-controller-manager
 spec:
-  replicas: 2
+  replicas: 1
```
hongchaodeng: Why change replica?

pkg/asset/internal/templates.go
```diff
@@ -282,9 +282,29 @@ metadata:
...
   template:
     metadata:
+      annotations:
```
hongchaodeng: Maybe combine this with below into same util and only need one argument?

pbx0:
ok to test

aaronlevy:
Anti-affinity will no longer be based on an annotation in v1.6, so the plan was to wait to add this functionality until v1.6.0 ships (likely a couple weeks): See https://groups.google.com/forum/#!msg/kubernetes-dev/-fwfHQtm9GE/H1TU4yqiBAAJ

aaronlevy:
Added the issue to the milestone tracking the changes we want to get in with the v1.6.0 release: https://github.com/kubernetes-incubator/bootkube/milestone/3

pbx0:
rktbot run tests

pkg/asset/internal/templates.go
adieu: With `replicas=2` and one node, one replica pod will fail to be scheduled due to pod anti affinity. The error log might confuse the user so I changed the replica number to 1. We could scale this to 2 and etcd pods to 3 when we have multiple master nodes available.

pkg/asset/internal/templates.go
adieu: You mean use a function to set the annotation? I think it's ok to have some duplications in the template.

adieu:
@aaronlevy Sorry I didn't notice the issue. I can make some changes to this pr after v1.6 release if you want.

aaronlevy:
Sure. I'll leave this open to track that

squat:
This was implemented in https://github.com/kubernetes-incubator/bootkube/pull/421.

aaronlevy:
Closed in https://github.com/kubernetes-incubator/bootkube/pull/421

PR #394, opened 2017-03-22
Bump k8s version v1.5.5

aaronlevy:


squat:
@aaronlevy Looks good. Wanna bump the versions in hack/quickstart?

aaronlevy:
@squat quickstart uses the container images - so we need to wait until after a release to bump those (otherwise you're just bumping the kubelet version). Not the most intuitive thing - and need to make this more clear (it is mentioned in the release docs: https://github.com/kubernetes-incubator/bootkube/blob/master/RELEASING.md#updating-quickstart-guides

PR #414, opened 2017-04-04
conformance-gce.sh: Pin to a particular gcloud installer version.

yifan-gu:
The latest gcloud installer scripts will break the test, so let's pin to an old version.


aaronlevy:
LGTM

PR #416, opened 2017-04-05
hack/quickstart: bootkube v0.3.13 and kubelet v1.5.6

pbx0:


pbx0:
Tests look good besides known failing scale test.

PR #418, opened 2017-04-06
Fix apiserver etcd-servers flag when self-hosting.

### LINKED ISSUE #372: Bootkube prematurely exits if scheduler/controller-manager lose leader-election
This is most easily surfaced when using self-hosted etcd - because there is a period of time when pivoting from boot-etcd to self-hosted etcd that the compiled-in control plane is no longer able to contact the etcd cluster (and therefore give up leader status).

When either the scheduler or controller-manager give up leadership it calls an `os.Exit()` which kills bootkube entirely - and the bootstrap process may not be complete (e.g. boot-etcd still running, or in another incomplete state).

One fix might to to just provide the bootkube api-server with the expected addresses of both etcd's (boot-etcd, and the self-hosted etcd service ip) when running `bootkube start`. For example: https://github.com/kubernetes-incubator/bootkube/blob/master/hack/multi-node/bootkube-up#L16

Another solution in the future might be to use static manifests for control plane components, such that if they exit they won't affect bootkube.

/cc @hongchaodeng @xiang90

### LINKED ISSUE #411: `bootkube start` should allow specifying multiple etcd-servers
The current flag `--etcd-server` only allows one etcd endpoint to be specified. However, we should allow multiple etcd nodes to be used, which would mitigate some issues during the self-hosted etcd pivot (see: https://github.com/kubernetes-incubator/bootkube/issues/372).

diegs:
This adds the self-hosted etcd cluster to the temporary control plane
config, which should hopefully mitigate situations where the control
plane loses contact with etcd during the boot-etcd -> self-hosted-etcd
pivot and gives up leader status as a result.

I also fixed the hack/single-node config to not start local etcd
when running with self-hosting.

Fixes #411.

### DIFF
--- a/pkg/bootkube/parse.go
+++ b/pkg/bootkube/parse.go
@@ -91,6 +91,24 @@ func detectEtcdIP(assetDir string) (string, error) {
 	return service.Spec.ClusterIP, nil
 }
 
+// detectEtcdServiceURL deserializes the etcd-service URL.
+func detectEtcdServiceURL(assetDir string) (string, error) {
+	path := filepath.Join(assetDir, asset.AssetPathEtcdSvc)
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("can't read file %s: %v", path, err)
+	}
+	var service v1.Service
+	err = yaml.Unmarshal(b, &service)
+	if err != nil {
+		return "", fmt.Errorf("can't unmarshal %s: %v", path, err)
+	}
+	if numPorts := len(service.Spec.Ports); numPorts != 1 {
+		return "", fmt.Errorf("expected 1 etcd cluster port, found: %d", numPorts)
+	}
+	return fmt.Sprintf("http://%s:%d", service.Spec.ClusterIP, service.Spec.Ports[0]), nil
+}
+
 func findFlag(flagName string, args []string) string {
 	for _, arg := range args {
 		if strings.HasPrefix(arg, flagName+"=") {
--- a/hack/single-node/bootkube-up
+++ b/hack/single-node/bootkube-up
@@ -23,6 +23,9 @@ fi
 if [ ! -d "cluster" ]; then
   ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.100:443 ${etcd_render_flags}
   cp user-data.sample cluster/user-data
+	if [ ${SELF_HOST_ETCD} = "false" ]; then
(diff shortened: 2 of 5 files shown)


diegs:
Signed the CLA.

pkg/bootkube/parse.go
```diff
@@ -91,6 +91,24 @@ func detectEtcdIP(assetDir string) (string, error) {
...
+	}
+	var service v1.Service
+	err = yaml.Unmarshal(b, &service)
+	if err != nil {
+		return "", fmt.Errorf("can't unmarshal %s: %v", path, err)
+	}
+	if numPorts := len(service.Spec.Ports); numPorts != 1 {
+		return "", fmt.Errorf("expected 1 etcd cluster port, found: %d", numPorts)
+	}
+	return fmt.Sprintf("http://%s:%d", service.Spec.ClusterIP, service.Spec.Ports[0]), nil
```
yifan-gu: Will there be `https` ?
/cc @aaronlevy @xiang90

xiang90: it will be. but not for now.

pbx0:
rktbot run tests

pbx0:
rktbot run tests

pbx0:
rktbot run tests

pbx0:
rktbot run tests

pbx0:
rktbot run tests

pbx0:
rktbot run tests

pbx0:
rktbot run tests

pbx0:
rktbot run tests

diegs:
Manually ran bootkube.selfetcd.destruct.reboot and bootkube.selfetcd.scale, both passed.

pbx0:
rktbot run tests

pbx0:
tests look good to me

aaronlevy:
One random note - github will use "fixes X" in PR comment to auto-close the issue. In this case I think it fixes https://github.com/kubernetes-incubator/bootkube/issues/411 -- but lets leave https://github.com/kubernetes-incubator/bootkube/issues/372 open to track other possible cases this could still happen

diegs:
@aaronlevy done

@pbx0 is looking at the tests I believe. I had to invoke them manually, rkbot isn't working for some reason.

pbx0:
@aaron: looking into. At least there are still some failures happening as a result of container pulling freezing up, I'm working on reducing number of pulls per test run via quayctl which i'm hoping will bring tests in-line with what we are seeing when run locally which should help make any remaining flakes far more obvious.

PR #422, opened 2017-04-10
templates.go: Update manifest for Kubernetes 1.6

yifan-gu:


yifan-gu:
Not sure what other steps need to be done when we remove "master=true" node selector?
Also I haven't updated the DNS templates, it seems the DNS images we are using now is quite out-of-date (1.9 vs 1.14), do we want to upgrade that too? @aaronlevy

yifan-gu:
After OOB discussion, I am going to update the dns manifest to the latest. The `master=true` node selector will be kept for now until new labels are introduced.

pkg/asset/internal/templates.go
```diff
@@ -174,6 +178,13 @@ spec:
         - --storage-backend=etcd3
         - --tls-cert-file=/etc/kubernetes/secrets/apiserver.crt
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
+        livenessProbe:
```
aaronlevy: unfortunately we can't use a liveness probe here - it will kill the flock'ed api-server, which we need to keep running during checkpoint pivots. We may be able to come up with a solution later - but the short-term fix was to not use the liveness probe.

pkg/asset/internal/templates.go
```diff
@@ -406,20 +426,25 @@ spec:
...
 kind: DaemonSet
 metadata:
   name: kube-proxy
   namespace: kube-system
   labels:
-    k8s-app: kube-proxy
+    tier: node
+    component: kube-proxy
+  annotations:
+    scheduler.alpha.kubernetes.io/critical-pod: ''
```
aaronlevy: Should this be part of the podspec annotations (not the daemonset annotation)

pkg/asset/internal/templates.go
```diff
@@ -560,9 +600,12 @@ spec:
         resources:
           requests:
             cpu: 150m
-            memory: 10Mi
-      - name: dnsmasq-metrics
-        image: gcr.io/google_containers/dnsmasq-metrics-amd64:1.0
+            memory: 20Mi
+        volumeMounts:
+        - name: kube-dns-config
+          mountPath: /etc/k8s/dns/dnsmasq-nanny
```
aaronlevy: We don't create this directory (but we do create /etc/kubernetes). What is the dnsmasq nanny process / is this needed?

yifan-gu: I think this is a mount to the configmap volume `kube-dns-config`.
Looks like the `dnsmasq nanny` is a replacement for the old `kube-dnsmasq-amd64:1.4` container

yifan-gu: @aaronlevy ^

aaronlevy: ack. Also this is the volumeMount so that's the path in the container - so we should be fine here.

cmd/bootkube/start.go
```diff
@@ -35,9 +37,14 @@ func init() {
 	cmdStart.Flags().StringVar(&startOpts.etcdServer, "etcd-server", "http://127.0.0.1:2379", "Single etcd node to use during bootkube bootstrap process.")
 	cmdStart.Flags().StringVar(&startOpts.assetDir, "asset-dir", "", "Path to the cluster asset directory. Expected layout genereted by the `bootkube render` command.")
 	cmdStart.Flags().BoolVar(&startOpts.selfHostedEtcd, "experimental-self-hosted-etcd", false, "Self hosted etcd mode. Includes starting the initial etcd member by bootkube.")
+	cmdStart.Flags().StringVar(&startOpts.flockPath, "apiserver-flock-path", "/var/lock/api-server.lock", "The flock path of the API server, used to prevent API server from port conflicts with the bootkube, not required when switched to static files.")
```
aaronlevy: I might just suggest hard-coding this to keep it relatively self-contained. The flock will soon just end up in the static manifest once https://github.com/kubernetes-incubator/bootkube/pull/425 lands. Which we should also coordinate this same change there (if it resolves the issue). /cc @diegs

pbx0:
tests failing because it looks like the default locks aren't created.
```
 [   71.809234] bootkube-dev[5]: Error: Failed to grab the flock at "/var/lock/api-server.lock": no such file or directory
[   71.809836] bootkube-dev[5]: Failed to grab the flock at "/var/lock/api-server.lock": no such file or directory
```

aaronlevy:
Side question - is this actually failing without the flock, or do you just see multiple restarts of the api-server pod? The multiple restarts have always been the case - so not sure that should block this (but I'm also not opposed to putting in the flock).

pbx0:
Bootkube itself is nearly immediately exiting during  `func runCmdStart`

yifan-gu:
I am still investigating. I found the flock doesn't solve the issue yet, hmm.

aaronlevy:
rktbot run tests

yifan-gu:
Conformance test looks good after I fixedd the kubelet image. https://jenkins.coreos.systems/job/bootkube-development/164/
So let's wait for #428

yifan-gu:
rktbot run tests

yifan-gu:
rktbot run tests

yifan-gu:
Kicked off conformance test https://jenkins.coreos.systems/job/bootkube-development/171/

yifan-gu:
@pbx0 @squat I think the bootkube-dev tests failed because the test is using `k8s-app` label to select for api pods. We should change the test to use `component=kube-apiserver && tier=control-plane` or just remove the label selector.

PR #425, opened 2017-04-11
Use static assets for temporary control plane.

### LINKED ISSUE #168: Possibly use static pods for temporary control-plane rather than compiled-in components

diegs:
This change:
- Starts the bootstrap control plane using static manifests instead of
  launching them in-process.
- Auto-detects whether self-hosted etcd is being used in the run phase,
  rather than requiring synchronized flag usage.
- Simplifies a lot of code based on the above refactorings.

After this change clients can completely customize their bootstrap
environment, including changing the version of kubernetes that is
launched.

We might be able to un-vendor Kubernetes and pull in specific
dependencies after this change. I started to investigate but it
looks like it'd be a fairly big change to include here.

Fixes #168.


diegs:
rktbot run tests

hack/quickstart/init-master.sh
```diff
@@ -39,13 +39,11 @@ function init_master_node() {
...
 
     etcd_render_flags=""
-    etcd_start_flags=""
 
     # Start etcd.
     if [ "$SELF_HOST_ETCD" = true ] ; then
         echo "WARNING: THIS IS NOT YET FULLY WORKING - merely here to make ongoing testing easier"
-        etcd_render_flags="--etcd-servers=http://10.3.0.15:2379 --experimental-self-hosted-etcd"
-        etcd_start_flags="--etcd-server=http://${COREOS_PRIVATE_IPV4}:12379 --experimental-self-hosted-etcd"
+        etcd_render_flags="--etcd-servers=http://10.3.0.15:2379 --experimental-self-hosted-etcd"  #TODO(diegs): maybe add --bootstrap-etcd-server=http://${COREOS_PRIVATE_IPV4}:12379
```
diegs: Note my TODO here: I've currently hardcoded the bootstrap etcd server to be at http://127.0.0.1:12379. If we need more flexibility we will need a flag to adjust for that.

pkg/bootkube/bootstrap.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	for _, secret := range secrets {
+		if err := copyFile(filepath.Join(secretsDir, secret.Name()), filepath.Join(asset.BootstrapSecretsDir, secret.Name()), true); err != nil {
+			return err
+		}
+	}
+
+	// Copy the static manifests to the kubelet's pod manifest path.
+	manifestsDir := filepath.Join(assetDir, asset.AssetPathBootstrapManifests)
```
diegs: I think there's a possible race here when bootstrapping etcd: if the apiserver tries to come up and etcd isn't up yet it will die. We could mitigate this by setting the restartPolicy on the bootstrap api server (and perhaps other pods too) or by waiting for etcd to be ready before bringing up the other pods (this was effectively the previous behavior).

aaronlevy: I might start with leaving the restart policy as default (always). If we don't need to provide coordination points (e.g. x starts before y) it keeps bootkube somewhat more simple -- and the api-server should just keep retrying until it can access etcd. This is a common pattern across kubernetes.

diegs: Agreed, done.

diegs:
rktbot run tests

diegs:
Note: tests won't pass as-is due to the flag and quickstart changes.

I have a separate patch for pluton (https://github.com/coreos/mantle/pull/549) which allows the tests to pass.

pkg/bootkube/bootkube.go
```diff
@@ -31,141 +23,65 @@ var requiredPods = []string{
...
 		}
-	}()
-	// If any of the bootkube services exit, it means it is unrecoverable and we should exit.
-	err := <-errch
-	if err != nil {
-		UserOutput("Error: %v\n", err)
 	}
-	return err
+
+	if err = CleanupBootstrapControlPlane(b.assetDir, b.podManifestPath); err != nil {
```
pbx0: Can this be moved to the top of this function but using `defer` This way assets get cleaned up even if bootkube has an error before reaching this function.

pbx0:
Overall LGTM, just one nit about how the cleanup routine is called. Thanks for digging into the test change required here. Since tests pass with https://github.com/coreos/mantle/pull/549 locally, this PR should more or less be ready to go. Working on making that PR unnecessary as well.

diegs:
Local test results:

diegs@x260:bootkube [static-manifests %]% ~/go/bin/pluton run --parallel 2 --basename diegs-bootkube-dev --platform=gce --gce-image=projects/coreos-cloud/global/images/coreos-stable-1298-7-0-v20170401 --bootkubeRepo=quay.io/coreos/bootkube-dev --bootkubeTag=57489f653e7fc7b865259648e723f1dc99024c79 bootkube.*
=== RUN   bootkube.destruct.reboot
=== RUN   bootkube.selfetcd.destruct.delete
=== RUN   bootkube.smoke
=== RUN   bootkube.destruct.delete
=== RUN   bootkube.selfetcd.smoke
=== RUN   bootkube.selfetcd.scale
=== RUN   bootkube.selfetcd.destruct.reboot
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
--- PASS: bootkube.destruct.reboot (612.91s)
--- PASS: bootkube.selfetcd.smoke (649.16s)
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
--- PASS: bootkube.destruct.delete (455.37s)
--- PASS: bootkube.smoke (533.51s)
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
--- PASS: bootkube.selfetcd.destruct.reboot (687.68s)
--- PASS: bootkube.selfetcd.destruct.delete (705.94s)
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
--- FAIL: bootkube.selfetcd.scale (730.83s)
        run.go:107: creating cluster: final node check: cannot detect all nodes in kubectl output 
map[10.240.0.27:{}]
[0xc42041e200 0xc420334e00]
harness: test suite failed
FAIL

The selfetcd.scale test has been flaky but I'll look into the logs to make sure it isn't anything troubling.

pbx0:
@diegs: looks like another flake due to hyperkube downloads freezing: I suspect if you look at the journal logs for that test you'll find the worker machine taking forever to download hyperkube.

hack/quickstart/init-master.sh
aaronlevy: I think we might be able to also drop the `--etcd-servers` flag altogether. 10.3.0.15 is the etcd service IP - which is already being inferred elsewhere from the etcd service asset (`detectEtcdIP()`)

pkg/bootkube/bootstrap.go
```diff
@@ -0,0 +1,84 @@
...
+	manifests, err := ioutil.ReadDir(filepath.Join(assetDir, asset.AssetPathBootstrapManifests))
+	if err != nil {
+		return err
+	}
+	for _, manifest := range manifests {
+		if err := os.Remove(filepath.Join(podManifestPath, manifest.Name())); err != nil {
+			return err
+		}
+	}
+	if err := os.RemoveAll(asset.BootstrapSecretsDir); err != nil {
```
aaronlevy: Agree with @pbx0 about trying to clean up regardless of success -- but also specifically secrets - we should always try to remove these (even if steps above fail)

pkg/bootkube/bootstrap.go
```diff
@@ -0,0 +1,84 @@
...
+	if err := os.Mkdir(asset.BootstrapSecretsDir, os.FileMode(0755)); err != nil {
+		return err
+	}
+	secretsDir := filepath.Join(assetDir, asset.AssetPathSecrets)
+	secrets, err := ioutil.ReadDir(secretsDir)
+	if err != nil {
+		return err
+	}
+	for _, secret := range secrets {
+		if err := copyFile(filepath.Join(secretsDir, secret.Name()), filepath.Join(asset.BootstrapSecretsDir, secret.Name()), true); err != nil {
```
aaronlevy: Leaving the secrets as world readable feels a bit wrong - but I'm not immediately sure what expectations we should set of how bootkube is executed and/or if the temp control plane should have a specific user (or privileged). Now that bootkube doesn't have to bind on protected port -- it doesn't really need special permissions, so I think we could scope its access down more.

Maybe we just add a TODO for now to re-asses the temporary secret handling later? I don't think this really should block us right now - as it is only first node on first boot.

hack/quickstart/init-master.sh
aaronlevy: Sorry this is for the render side (not start). But I do think we should probably just pick the service IP for the user if they're using self-hosted etcd. It has to be within the service-cidr, and we already assume certain IPs for other services (e.g. DNS gets the .10).

That change can be follow-up/optimization though (maybe just add a TODO if not addressed here)

aaronlevy:
rktbot run tests

pkg/bootkube/bootkube.go
diegs: Excellent point, done.

pkg/asset/internal/templates.go
```diff
@@ -201,6 +201,51 @@ spec:
...
+	BootstrapAPIServerTemplate = []byte(`apiVersion: v1
+kind: Pod
+metadata:
+  name: bootstrap-kube-apiserver
+  namespace: kube-system
+spec:
+  hostNetwork: true
+  containers:
+  - name: kube-apiserver
+    image: quay.io/coreos/hyperkube:v1.6.0_coreos.0
```
diegs: Done.

pkg/asset/internal/templates.go
```diff
@@ -355,6 +400,42 @@ spec:
...
+	BootstrapControllerManagerTemplate = []byte(`apiVersion: v1
+kind: Pod
+metadata:
+  name: bootstrap-kube-controller-manager
+  namespace: kube-system
+spec:
+  hostNetwork: true
+  containers:
+  - name: kube-controller-manager
+    image: quay.io/coreos/hyperkube:v1.6.0_coreos.0
```
diegs: Done

pkg/asset/internal/templates.go
```diff
@@ -397,6 +478,23 @@ spec:
...
+	BootstrapSchedulerTemplate = []byte(`apiVersion: v1
+kind: Pod
+metadata:
+  name: bootstrap-kube-scheduler
+  namespace: kube-system
+spec:
+  hostNetwork: true
+  containers:
+  - name: kube-scheduler
+    image: quay.io/coreos/hyperkube:v1.6.0_coreos.0
```
diegs: Done

pkg/bootkube/bootstrap.go
diegs: Agreed. Switched the order to remove the secrets first. Overall the chaining of all this is a little brittle. I'll give some thought to a smarter way for handling the shared secrets.

pkg/bootkube/bootstrap.go
diegs: Is it safe to assume that bootkube and the kubelet are always both running as root (seems to be the case in container linux)? If so I think we can lock this down to 0700.

hack/quickstart/init-master.sh
diegs: The render code actually already does this: https://github.com/kubernetes-incubator/bootkube/blob/master/cmd/bootkube/render.go#L163

I'll just clean up the unnecessary uses in the scripts.

pkg/asset/internal/templates.go
diegs: Note: I also added the flock here based on other discussions to try to reduce the port contention churn. However, it often takes >30s for the self-hosted pods to come up, and definitely >30s to migrate etcd, and we don't tear down the temporary apiserver until all that is done. So, unless we increase the flock timeouts more we'll still see pod restarts for the self-hosted apiserver.

pkg/bootkube/bootstrap.go
aaronlevy: @dghubble do you know what our expectations are for how we're executing bootkube?

aaronlevy: Actually it looks like we do run it as root. If we keep that assumption we could change the permissions (0700 for dir 0600 for files)

diegs: Sounds good, I already verified that was true on the hack/ scripts, made the change, and ran it through the tests.

pkg/bootkube/bootstrap.go
```diff
@@ -0,0 +1,84 @@
...
+			return fmt.Errorf("file already exists: %v", dst)
+		} else if !os.IsNotExist(err) {
+			return err
+		}
+	}
+	data, err := ioutil.ReadFile(src)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(dst, data, os.FileMode(0644))
```
aaronlevy: If this is just for secrets, can we switch to `0600`

aaronlevy:
I just ran some manual smoke tests (also with self-hosted etcd). All looks good.

pkg/bootkube/bootstrap.go
diegs: It's also for manifests, but I think 0600 is good there too.

diegs:
Cool, I also ran it through manual tests (hack/multi-node) as well as invoking the modified pluton tests again.

diegs:
Ok, squashed and rebased. Running one more pass of tests now.

diegs:
All tests passed except for the etcd scale test. I think that's unrelated.

diegs@x260:bootkube [etcd-tls *]% ~/go/bin/pluton run --parallel 2 --basename diegs-bootkube-dev --platform=gce --gce-image=projects/coreos-cloud/global/images/coreos-stable-1298-7-0-v20170401 --bootkubeRepo=quay.io/coreos/bootkube-dev --bootkubeTag=0d6d03da5a50fb4abde3b10d0674bb052704e14d bootkube.*
=== RUN   bootkube.selfetcd.destruct.reboot
=== RUN   bootkube.smoke
=== RUN   bootkube.destruct.reboot
=== RUN   bootkube.destruct.delete
=== RUN   bootkube.selfetcd.smoke
=== RUN   bootkube.selfetcd.scale
=== RUN   bootkube.selfetcd.destruct.delete
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
--- PASS: bootkube.selfetcd.destruct.reboot (561.92s)
--- PASS: bootkube.selfetcd.smoke (579.02s)
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
--- PASS: bootkube.destruct.delete (523.29s)
--- PASS: bootkube.destruct.reboot (629.67s)
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
--- PASS: bootkube.smoke (463.82s)
--- PASS: bootkube.selfetcd.destruct.delete (409.79s)
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
--- FAIL: bootkube.selfetcd.scale (942.62s)
        scale.go:44: scaling down: Waited 150 seconds for etcd to scale: expected 1 etcd pods got 3: [Running Running Running]
harness: test suite failed
FAIL

pkg/bootkube/bootstrap.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	return nil
+}
+
+func copyFile(src, dst string, overwrite bool) error {
+	if !overwrite {
+		fi, err := os.Stat(dst)
+		if fi != nil {
+			return fmt.Errorf("file already exists: %v", dst)
+		} else if !os.IsNotExist(err) {
```
yifan-gu: nit: `else` is not required.

pkg/asset/asset.go
```diff
@@ -43,30 +44,39 @@ const (
 	AssetPathEtcdSvc                     = "manifests/etcd-service.yaml"
 	AssetPathKenc                        = "manifests/kenc.yaml"
 	AssetPathKubeSystemSARoleBinding     = "manifests/kube-system-rbac-role-binding.yaml"
+	AssetPathBootstrapManifests          = "bootstrap-manifests"
+	AssetPathBootstrapAPIServer          = "bootstrap-manifests/bootstrap-apiserver.yaml"
+	AssetPathBootstrapControllerManager  = "bootstrap-manifests/bootstrap-controller-manager.yaml"
+	AssetPathBootstrapScheduler          = "bootstrap-manifests/bootstrap-scheduler.yaml"
+	AssetPathBootstrapEtcd               = "bootstrap-manifests/bootstrap-etcd.yaml"
+	BootstrapSecretsDir                  = "/tmp/bootkube/secrets"
```
yifan-gu: Why this is in /tmp ?

diegs: Static manifests can only use filesystem mounts, not shared secrets. I need to put them somewhere temporarily--is there a more general place than /tmp to do so?

yifan-gu: @diegs My dumb question is why not just `bootstrap-secrets`/?

yifan-gu: nvm, I think it doesn't matter, we will remove the secrets anyway.

yifan-gu:
LGTM from my side.
Might need to wait for #429 and #422

PR #429, opened 2017-04-13
hack/quickstart/init-master.sh: Fix bootkube command.

yifan-gu:
A simple fix because the bootkube is not running in rkt container now.


yifan-gu:
Conformance test kicked off https://jenkins.coreos.systems/job/bootkube-development/170/console

yifan-gu:
Conformance tests passed, merging.

PR #430, opened 2017-04-13
Use /etc/kubernetes/inactive-manifests instead of /src/kubernetes/manifests

derekparker:
Using `/srv/kubernetes` is less intuitive, everything else lives under `/etc/kubernetes`, and it's possible the two mounts could be different filesystems which defeats the purpose of using an atomic `mv` internally.

This patch moves "inactive" checkpointed manifests to `/etc/kubernetes/inactive-manifests`.

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -239,7 +239,7 @@ spec:
       volumes:
       - name: checkpoint-dir
         hostPath:
-          path: /etc/kubernetes/checkpoint-iptables    
+          path: /etc/kubernetes/checkpoint-iptables
       - name: var-lock
         hostPath:
           path: /var/lock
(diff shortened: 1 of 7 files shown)


pkg/asset/internal/templates.go
```diff
@@ -265,7 +265,7 @@ spec:
       hostNetwork: true
       containers:
       - name: checkpoint
-        image: quay.io/coreos/pod-checkpointer:a936ee1149af95c7696a261e739b050ff00c6288
+        image: quay.io/coreos/pod-checkpointer:99ddc731171c0ee860ca482dd25db161a5ceb304
```
diegs: Just for my education: why do this in the same PR?

derekparker: Could be done in another PR, and we usually do, however I have become less familiar with the testing of bootkube as things have changed, and just wanted to make sure the new checkpointer is used in any / all tests ran.

derekparker: cc @pbx0 if my assumption / thinking here is correct.

derekparker:
rktbot run checkpointer tests

Issue #403, opened 2017-03-24
Support etcd+TLS


aaronlevy:
We should have options for both "bring your own etcd" and "self-hosted" etcd. And this should be the default installation method.

- The etcd TLS assets needed by the api-server should be populated in a secret
- The data that needs to end up on etcd nodes (non-self-hosted) will need to be plumbed through all `hack/*` examples (e.g. when launching etcd-member.service, it needs to have the generated TLS assets available).
- self-hosted etcd will likely require changes to the etcd-operator - as that is responsible for launching the etcd pods - and we would need to add secret volumes to distribute the TLS assets.

Related:
https://github.com/kubernetes-incubator/bootkube/issues/244
https://github.com/kubernetes-incubator/bootkube/issues/389

mfburnett:
Related: https://github.com/coreos/tectonic-forum/issues/86

hongchaodeng:
@aaronlevy @mfburnett 
Are there any well-defined use case for TLS setup? How does user want to generate certs?
Implementing TLS is simple, but is also permanent.

vmatekole:
@aaronlevy  and @mfburnett ... I attempted to enable auto-tls in etcd3  within the tectonic-controller.yaml.tmpl  located in the published Ignition assets from the installer to Provisioner:
```
systemd:
  units:
    - name: etcd-member.service
      enable: true
      dropins:
        - name: 40-etcd-cluster.conf
          contents: |
            [Service]
            Environment="ETCD_IMAGE_TAG=v3.1.2"
            {{- if index . "external_etcd"}}
            ExecStart=
            ExecStart=/usr/lib/coreos/etcd-wrapper gateway start \
                --listen-addr=0.0.0.0:2379 \
                --endpoints={{.etcd_endpoints}}
            {{- else}}
            Environment="ETCD_LISTEN_CLIENT_URLS=https://0.0.0.0:2379"
            ETCD_INITIAL_CLUSTER="node0=https://etcd0.seedbloom.it:2380,node1=https://etcd1.seedbloom.it:2380,node2=https://etcd2.seedbloom.it:2380
            Environment="ETCD_ADVERTISE_CLIENT_URLS=https://{{.domain_name}}:2379"
            Environment="ETCD_INITIAL_ADVERTISE_PEER_URLS=https://{{.domain_name}}:2380"
            Environment="ETCD_LISTEN_PEER_URLS=https://0.0.0.0:2380"
[...]

Issue #389, opened 2017-03-17
Etcd with tls


devent:
Hello,
How to use bootkube and a etcd cluster that is secured by TLS and client auth?
In the `kube-apiserver.yaml` I need to add 
```
        - --etcd-cafile=ca_cert.pem
        - --etcd-certfile=etcd_cert.pem
        - --etcd-keyfile=etcd_key.pem
```

But the `kube-apiserver.yaml` is generated from bootkube render. How to add those extra arguments? And how can the DaemonSet find the certificates?

I tried to patch kube-apiserver.yaml as follow, but no success.

```
apiVersion: "extensions/v1beta1"
kind: DaemonSet
metadata:
  name: kube-apiserver
  namespace: kube-system
  labels:
    k8s-app: kube-apiserver
spec:
  template:
    metadata:
      labels:
        k8s-app: kube-apiserver
      annotations:
        checkpointer.alpha.coreos.com/checkpoint: "true"
    spec:
      nodeSelector:
        master: "true"
      hostNetwork: true
      containers:
      - name: kube-apiserver
        image: quay.io/coreos/hyperkube:v1.5.4_coreos.0
        command:
        - /usr/bin/flock
        - --exclusive
        - --timeout=30
        - /var/lock/api-server.lock
        - /hyperkube
        - apiserver
        - --bind-address=0.0.0.0
        - --secure-port=443
        - --insecure-port=8080
        - --advertise-address=$(POD_IP)
        - --etcd-servers=https://xxx:2379
        - --etcd-cafile=/etc/etcd-ssl/ca_cert.pem
        - --etcd-certfile=/etc/etcd-ssl/client_cert.pem
        - --etcd-keyfile=/etc/etcd-ssl/client_key_insecure.pem
        - --storage-backend=etcd3
        - --allow-privileged=true
        - --service-cluster-ip-range=10.3.0.0/24
        - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota
        - --runtime-config=api/all=true
        - --tls-cert-file=/etc/kubernetes/secrets/apiserver.crt
        - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
        - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
        - --kubelet-client-key=/etc/kubernetes/secrets/apiserver.key
        - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
[...]

PR #433, opened 2017-04-14
Add support for generating and using etcd TLS assets.

### LINKED ISSUE #244: Add options to pass etcd certs to bootkube

diegs:
This change adds rendering options to allow the apiserver to connect
to etcd using TLS. This applies to both the temporary and self-hosted
control plane.

There are also some options (mostly intended for development) to help
generate the etcd (client/server) certificates. Obviously this is only
useful if etcd is not already up. I updated the quickstart/init-master.sh
script to use this to demonstrate the new functionality.

Self-hosted etcd is not supported at this time.

Partial fix for #403, fixes #244, #389.


diegs:
rktbot run tests

cmd/bootkube/render.go
```diff
@@ -63,6 +71,7 @@ func init() {
 	cmdRender.Flags().BoolVar(&renderOpts.selfHostKubelet, "experimental-self-hosted-kubelet", false, "(Experimental) Create a self-hosted kubelet daemonset.")
 	cmdRender.Flags().StringVar(&renderOpts.cloudProvider, "cloud-provider", "", "The provider for cloud services.  Empty string for no provider")
 	cmdRender.Flags().BoolVar(&renderOpts.selfHostedEtcd, "experimental-self-hosted-etcd", false, "(Experimental) Create self-hosted etcd assets.")
+	cmdRender.Flags().BoolVar(&renderOpts.etcdUseTLS, "etcd-use-tls", false, "If true, uses TLS for etcd. Implicitly true if --etcd-ca-path,--etcd-certificate-path,--etcd-private-key-path are set. If true but those flags are not set etcd TLS certificates will be generated. Not supported if --experimental-self-hosted-etcd=true.")
```
aaronlevy: Should we just force etcd to always use TLS? I think if the TLS assets are provided, we use them. And if not, we generate them for the user.

In the case of self-hosted etcd - we can't support this yet, but maybe we just warn the user if they try and set the asset flags and say it's not supported yet?

hack/quickstart/init-master.sh
```diff
@@ -16,16 +16,25 @@ function usage() {
...
         cat << EOF > /etc/systemd/system/etcd-member.service.d/10-etcd-member.conf
 [Service]
 Environment="ETCD_IMAGE_TAG=v3.1.0"
 Environment="ETCD_NAME=controller"
 Environment="ETCD_INITIAL_CLUSTER=controller=http://${COREOS_PRIVATE_IPV4}:2380"
 Environment="ETCD_INITIAL_ADVERTISE_PEER_URLS=http://${COREOS_PRIVATE_IPV4}:2380"
-Environment="ETCD_ADVERTISE_CLIENT_URLS=http://${COREOS_PRIVATE_IPV4}:2379"
-Environment="ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379"
+Environment="ETCD_ADVERTISE_CLIENT_URLS=https://${COREOS_PRIVATE_IPV4}:2379"
+Environment="ETCD_SSL_DIR=/etc/etcd/tls"
```
aaronlevy: Do we need this if we're providing the assets already?

aaronlevy:
or @heyitsanthony for etcd sanity check (others may be in diff timezone)

cmd/bootkube/render.go
diegs: Recapping our offline discussion: yes, we should always use TLS. I'll remove the flag and always either use or generate certs. I'll also generate a set of peer certs if none were provided, and update all the hack scripts to always use the certs.

For self-hosted, I'll simply fail for now if the flags are specified.

hack/quickstart/init-master.sh
diegs: Note sure I fully follow your question, but yes, we have to set the ETCD_SSL_DIR environment variable to mount /etc/etcd/tls directory under /etc/ssl/certs in the etcd container so etcd can see the certs (referenced below).

cmd/bootkube/render.go
dghubble: If I'm following along, this sounds like it might be an all or nothing hurdle for provisioning tools. We'll need to add path units to block etcd3 from starting until the TLS credentials are in place (when there is not secret store). Tectonic will need to distribute those credentials as well in ways that are appropriate for the platform.

If its reasonable, I'd recommend bootkube allow etcd TLS but not require it until v0.5.0.

aaronlevy: /cc @philips thoughts on this? I was approaching it from we only allow etcd+TLS (same as we only allow kubelet/api+TLS) -- but should we still allow this to be optional?

aaronlevy: Also, etcd would just be in a failure loop until the assets exist, don't know that we need to explicitly coordinate on the paths (but we could I guess)

diegs: @aaronlevy @dghubble the generation of etcd assets here is mainly intended as a convenience to help with testing. I was under the assumption that Tectonic would generate etcd certs upstream and simply pass them into bootkube. That's the main use production use case I would envision.

I can keep support for insecure etcd as well if it's a huge pain point for Tectonic, but I agree with Aaron that it feels like a stopgap we should try to discourage as much as possible.

dghubble: Sure, etcd can crash loop until the credentials are in place on the host. Bare-Metal can scp these into place. Cloud platforms can pull them from KMS. The story is similar to how the kubeconfig is placed. The challenges are not new. I just want to point out the risk here is that these clients need to make infrastructure specific changes to be able to upgrade to the new version of bootkube that ships with this change.

aaronlevy: It will be harder to add later -- so my vote would still be to require it (and that we make all the necessary changes in various installers as well).

dghubble: Ok

diegs: I reworked it so the render phase will accept `http://` etcd endpoints (and not set the TLS flags). However, the default flag still uses TLS, and when using TLS bootkube will create certificates for etcd if none were provided.

cmd/bootkube/render.go
```diff
@@ -55,6 +59,9 @@ func init() {
 	cmdRender.Flags().StringVar(&renderOpts.assetDir, "asset-dir", "", "Output path for rendered assets")
 	cmdRender.Flags().StringVar(&renderOpts.caCertificatePath, "ca-certificate-path", "", "Path to an existing PEM encoded CA. If provided, TLS assets will be generated using this certificate authority.")
 	cmdRender.Flags().StringVar(&renderOpts.caPrivateKeyPath, "ca-private-key-path", "", "Path to an existing Certificate Authority RSA private key. Required if --ca-certificate is set.")
+	cmdRender.Flags().StringVar(&renderOpts.etcdCAPath, "etcd-ca-path", "", "Path to an existing PEM encoded CA that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-certificate-path and --etcd-private-key-path, and must have etcd configured to use TLS with matching secrets.")
+	cmdRender.Flags().StringVar(&renderOpts.etcdCertificatePath, "etcd-certificate-path", "", "Path to an existing certificate that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-private-key-path, and must have etcd configured to use TLS with matching secrets.")
```
aaronlevy: Allowing this makes the use of these three flags (ca, cert, key) have varying behaviors. E.g. if `etcd-certificate-path` is provided, it's assumed that the key is for that cert, not the CA. But if it is not provided, the key is for the CA not the cert.

I'm thinking we could even remove this flag and only allow the option of providing a CA/key to use while generating certs for the etcd nodes. Only allowing the ca cert/key to be provided is more inline with how we generate the rest of the certs (user provides CA, which we use). Providing your own cert is something that is very easy to get wrong as well (wrong CN/SAN, etc.) which is why we've gone the route of "provide CA, but we'll generate).

Also, along those lines I'm wondering if we need to allow for providing a different etcd-specific CA altogether. The flexibility can't hurt -- just thinking aloud.

What about if we changed this to just allow:

```
--etcd-ca-certificate-path=
--etcd-ca-key-path=
```

Then use those to generate the assets we need.

Or, now that we're assuming cert generation based on etcd server scheme -- if we didn't care about supporting a separate etcd-specific CA - we wouldn't really need any new flags (just use the existing `--ca-certificate-x-path` flags).

Sorry for the back-forth on this. Obviously more considerations than I initially thought.

/cc @dghubble @ericchiang any thoughts here? Specifically, do we want a diff CA for etcd? Is it safe to assume that given a CA we can generate the assets in the default case?

ericchiang: > Specifically, do we want a diff CA for etcd?

For the "bring your own etcd" case we're supporting, I think it's important that these could be different.

diegs: This is definitely another mode of operation we could support, either in addition to or instead of the other modes.

I guess my thought was that if you were bringing your own etcd cluster you'd have the CA and certs for that already. The CA could certainly be the same one as what is used to generate the kubelet certs.

So, in total we have the following possible modes of operation:

1. user provides etcd ca, certs
2. user provides etcd ca, we use it to generate etcd certs
3. user provides kubelet ca, we use it to generate etcd certs
4. user provides no ca, we generate ca (shared for both kubelet and etcd) & use it to generate etcd certs

right now we support 1, 3, and 4. we could trivially add support for 2. what you describe would be to remove support for 1, and not support 2.

I'm cool with whatever you think makes sense :)

aaronlevy: Actually, in-person discussion - I misread this. The behavior is that all three "bring your own flags" must be set:

```
--etcd-ca-path
--etcd-private-key-path
--etcd-certificate-path
```

Or the we will generate the assets for you using the provided CA or we will generate one for you.

I think this is great -- covers the three common cases:

1. Bring your own etcd
1. Bring your own CA, but generate assets for me
1. Generate CA, generate all assets for me

tldr: ignore my previous comments.

PR #434, opened 2017-04-14
templates: add toleration to control-plane components

squat:
This PR adds taints and tolerations to the bootkube templates to ensure that node workloads do not get scheduled on master nodes.

cc @aaronlevy @yifan-gu

### DIFF
--- a/hack/multi-node/bootkube-up
+++ b/hack/multi-node/bootkube-up
@@ -22,7 +22,7 @@ if [ ! -d "cluster" ]; then
   ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.101:443 ${etcd_render_flags}
   cp user-data.sample cluster/user-data-worker
   cp user-data.sample cluster/user-data-controller
-  sed -i.bak -e '/--node-labels=node-role.kubernetes.io\/master/d' cluster/user-data-worker
+  sed -i.bak -e '/node-role.kubernetes.io\/master/d' cluster/user-data-worker
 fi
 
 # Start the VM
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -99,6 +99,11 @@ spec:
           mountPath: /rootfs
       hostNetwork: true
       hostPID: true
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        operator: Equal
+        value: ""
+        effect: NoSchedule
       volumes:
       - name: dev
         hostPath:
@@ -147,12 +152,6 @@ spec:
         checkpointer.alpha.coreos.com/checkpoint: "true"
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
-      tolerations:
-      - key: "CriticalAddonsOnly"
-        operator: "Exists"
-      nodeSelector:
-        node-role.kubernetes.io/master: ""
-      hostNetwork: true
       containers:
       - name: kube-apiserver
         image: quay.io/coreos/hyperkube:v1.6.1_coreos.0
@@ -197,6 +196,16 @@ spec:
         - mountPath: /var/lock
           name: var-lock
           readOnly: false
+      hostNetwork: true
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      tolerations:
+      - key: CriticalAddonsOnly
+        operator: Exists
+      - key: node-role.kubernetes.io/master
+        operator: Equal
+        value: ""
+        effect: NoSchedule
       volumes:
       - name: ssl-certs-host
         hostPath:
@@ -215,7 +224,6 @@ metadata:
   name: bootstrap-kube-apiserver
   namespace: kube-system
 spec:
-  hostNetwork: true
   containers:
   - name: kube-apiserver
     image: quay.io/coreos/hyperkube:v1.6.1_coreos.0
@@ -252,6 +260,7 @@ spec:
     - mountPath: /var/lock
       name: var-lock
       readOnly: false
+  hostNetwork: true
   volumes:
   - name: secrets
     hostPath:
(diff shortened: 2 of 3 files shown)


pkg/asset/internal/templates.go
```diff
@@ -599,6 +626,14 @@ spec:
         - name: etc-kubernetes
           mountPath: /etc/kubernetes
           readOnly: true
+      hostNetwork: true
+      tolerations:
+      - key: CriticalAddonsOnly
```
squat: This toleration was missing entirely from this critical pod, meaning it wouldn't get scheduled when it was dire

hack/multi-node/bootkube-up
```diff
@@ -22,7 +22,7 @@ if [ ! -d "cluster" ]; then
   ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.101:443 ${etcd_render_flags}
   cp user-data.sample cluster/user-data-worker
   cp user-data.sample cluster/user-data-controller
-  sed -i.bak -e '/--node-labels=node-role.kubernetes.io\/master/d' cluster/user-data-worker
+  sed -i.bak -e '/node-role.kubernetes.io\/master/d' cluster/user-data-worker
```
squat: Modifies the regex to ensure sed removes both the lines with `node-labels` and `register-with-taints`

squat:
failed on
```
Total time extracting prefetched containers to disk: 37.369150289s
--- FAIL: bootkube.selfetcd.smoke (179.01s)
        run.go:148: creating cluster: failed to request new GCE instance: googleapi: Error 503: Code: '3530860079175192029', backendError
```

squat:
rktbot run tests

pkg/asset/internal/templates.go
```diff
@@ -431,6 +445,13 @@ spec:
         - name: ssl-host
           mountPath: /etc/ssl/certs
           readOnly: true
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      tolerations:
```
yifan-gu: This manifest already deinifes a `tolerations` field, can you merge them? @squat

pkg/asset/internal/templates.go
```diff
@@ -540,7 +559,13 @@ spec:
             port: 10251  # Note: Using default port. Update if --port option is set differently.
           initialDelaySeconds: 15
           timeoutSeconds: 15
-
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      tolerations:
```
yifan-gu: ditto

pkg/asset/internal/templates.go
```diff
@@ -147,12 +152,6 @@ spec:
         checkpointer.alpha.coreos.com/checkpoint: "true"
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
-      tolerations:
-      - key: "CriticalAddonsOnly"
```
yifan-gu: Why this is removed from kube-apiserver? Or is this a previous fault?

yifan-gu:
Some small nits.

pkg/asset/internal/templates.go
squat: Good catch. fixed

pkg/asset/internal/templates.go
squat: fixed

pkg/asset/internal/templates.go
squat: it was accidentaly removed while adding other toleration. fixed

squat:
@yifan-gu thanks for the review. I addressed the nits and pushed the fixes

yifan-gu:
LGTM

pkg/asset/internal/templates.go
```diff
@@ -774,6 +795,18 @@ spec:
             memory: 20Mi
             cpu: 10m
       dnsPolicy: Default  # Don't use cluster DNS.
+      tolerations:
```
Quentin-M: no nodeSelector in that one?

yifan-gu: Good question, @aaronlevy

yifan-gu: I don't see why not put it on master nodes.

aaronlevy: Kind of up to us -- I don't have super strong feelings here.

yifan-gu: I would prefer to put them on master nodes as DNS pod is virtually part of the control plane, it's almost always turned on by default.

aaronlevy: wfm

squat: added the nodeselector and pushed the commit

yifan-gu:
LGTM

Quentin-M:
@squat Trying to use your PR in Tectonic (v1.6.1_coreos.0) ... getting:

```
Error: invalid argument "node-role.kubernetes.io/master:NoSchedule" for --register-with-taints=node-role.kubernetes.io/master:NoSchedule: invalid taint spec: node-role.kubernetes.io/master:NoSchedule
```

Quentin-M:
It has to be ```--register-with-taints=node-role.kubernetes.io/master=:NoSchedule```. Can you follow up?

xiang90:
I do not feel we should get this merged. Not all master components (etcd) now tolerate this taints. We should make sure everything can tolerate taints before moving forward with this solution.

also i do not think the test should ever pass. either we taints incorrectly or we use an old version of k8s for testing.

can we revert this back?

hongchaodeng:
Any testing on this?
IIUC, with this change, the testing still doesn't set any taints on node in testing. @pbx0 to confirm?
If so, we should revert this until testing is covered.

aaronlevy:
The tests pass due to a bug in flannel which strips the taints off the node objects (https://github.com/coreos/flannel/issues/667). Once we merge a fix for that (https://github.com/kubernetes-incubator/bootkube/issues/454) self-hosted etcd will stop working until the pods tolerate the taint.

Rather than reverting this -- it seems like adding the toleration should be a relatively trivial change -- can we just get that merged?

aaronlevy:
All other master components do already tolerate this taint - we should just need etcd pods as final step

xiang90:
@aaronlevy We should never merge change that breaks test. If it breaks existing stuff, we need to fix first, then merge. It should never go the other way around. I still think we should revert the change until we figure everything out.

PR #435, opened 2017-04-14
pkg/asset: Add critical pod annotations to control-plane

derekparker:
For more information: https://kubernetes.io/docs/concepts/cluster-administration/guaranteed-scheduling-critical-addon-pods/.

Since the hack setups (and I believe all cluster setups) only use a single master we could run the rescheduler, however for this PR I am not including it as we have concerns over the rescheduler not being leader elected, and potential issues with multi-master setups. See https://github.com/kubernetes/contrib/issues/2551.

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -145,7 +145,11 @@ spec:
         component: kube-apiserver
       annotations:
         checkpointer.alpha.coreos.com/checkpoint: "true"
+        scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
+      tolerations:
+      - key: "CriticalAddonsOnly"
+        operator: "Exists"
       nodeSelector:
         master: "true"
       hostNetwork: true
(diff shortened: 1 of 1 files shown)


pkg/asset/internal/templates.go
```diff
@@ -145,6 +145,7 @@ spec:
         component: kube-apiserver
       annotations:
         checkpointer.alpha.coreos.com/checkpoint: "true"
+        scheduler.alpha.kubernetes.io/critical-pod: ''
```
aaronlevy: Given this is a daemonset, should this also get the toleration? or should it not have the annotation at all? The taint is currently only scheduler side (not kubelet enforced) -- daemonsets don't use scheduler -- so my guess is that this would be ignored (unless daemon controller respects the taint?)

derekparker: The annotation also exists on `kube-proxy` but not the toleration, so I added it here for consistency. However, daemonsets do respect taints / tolerations so I don't think it hurts to add the toleration to the daemonsets as well. Thoughts?

aaronlevy: Ah right. The annotation is basically so it can check if it isn't being scheduled (and would evict other work).

Yeah, if the daemon controller respects it (and maybe later kubelet enforces taints) might not be bad to have the toleration already set.

derekparker: Added.

PR #436, opened 2017-04-14
hack,templates: rename `master` node label

squat:
Update the master node label from `master` to
`node-role.kubernetes.io/master` as documented in
https://github.com/kubernetes/kubernetes/pull/39112.

cc @aaronlevy @yifan-gu

### DIFF
--- a/hack/multi-node/bootkube-up
+++ b/hack/multi-node/bootkube-up
@@ -22,7 +22,7 @@ if [ ! -d "cluster" ]; then
   ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.101:443 ${etcd_render_flags}
   cp user-data.sample cluster/user-data-worker
   cp user-data.sample cluster/user-data-controller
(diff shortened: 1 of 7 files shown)


hack/multi-node/bootkube-up
```diff
@@ -22,7 +22,7 @@ if [ ! -d "cluster" ]; then
   ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.101:443 ${etcd_render_flags}
   cp user-data.sample cluster/user-data-worker
   cp user-data.sample cluster/user-data-controller
-  sed -i.bak -e '/--node-labels=master=true/d' cluster/user-data-worker
+  sed -i.bak -e '/--node-labels=node-role.kubernetes.io\/master=true/d' cluster/user-data-worker
```
aaronlevy: missed one

squat: ah thanks. good catch

PR #437, opened 2017-04-15
Change temporary secrets path to /etc/kubernetes/bootstrap-secrets.

diegs:
This reduces the number of paths that users need to make available to
bootkube, especially if running in a container.

I think this is a good change because the user only needs to guarantee
read/write access to one path (/etc/kubernetes) rather than two.


yifan-gu:
LGTM

dghubble:
:+1:

diegs:
rktbot run tests

diegs:
Tests won't pass without the following matching pluton change:

https://github.com/coreos/mantle/pull/561

PR #449, opened 2017-04-18
hack: fix taint value

squat:
getting error
```
Error: invalid argument "node-role.kubernetes.io/master:NoSchedule" for --register-with-taints=node-role.kubernetes.io/master:NoSchedule: invalid taint spec: node-role.kubernetes.io/master:NoSchedule
```

I found that the taint was missing the taint value. It needed a `=` so that the value is set to empty string `""`.

Fixes @Quentin-M's issues in https://github.com/kubernetes-incubator/bootkube/pull/434.

### DIFF
--- a/hack/multi-node/user-data.sample
+++ b/hack/multi-node/user-data.sample
@@ -32,7 +32,7 @@ coreos:
           --allow-privileged \
           --hostname-override=${COREOS_PUBLIC_IPV4} \
           --node-labels=node-role.kubernetes.io/master \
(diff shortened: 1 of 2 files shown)


Quentin-M:
LGTM! Thanks!

pbx0:
Is any reason why this isn't also in the hack/quickstart [service files](https://github.com/kubernetes-incubator/bootkube/blob/master/hack/quickstart/kubelet.master)? If not lets put it there too for consistency with PR testing. (tests source their service files from quickstart)

squat:
Good call, @pbx0. I just pushed that change.

Issue #463, opened 2017-04-24
Nodes unavailable during bootkube start when --cloud-provider=gce


dghubble:
Launching a GCE cluster with on-host Kubeletes with `--cloud-provider=gce` and creating assets with bootkube v0.4.0 and the `--cloud-provider=gce` render flag, bootkube start does not complete.

```
NAMESPACE     NAME                                                                        READY     STATUS             RESTARTS   AGE
kube-system   bootstrap-kube-apiserver-jakku-controller.c.dghubble-io.internal            1/1       Running            0          14m
kube-system   bootstrap-kube-controller-manager-jakku-controller.c.dghubble-io.internal   1/1       Running            0          13m
kube-system   bootstrap-kube-scheduler-jakku-controller.c.dghubble-io.internal            1/1       Running            0          14m
kube-system   kube-apiserver-9r5d4                                                        0/1       CrashLoopBackOff   10         14m
kube-system   kube-controller-manager-515890209-29xdg                                     0/1       Pending            0          14m
kube-system   kube-controller-manager-515890209-g08rc                                     0/1       Pending            0          14m
kube-system   kube-dns-2891576090-4q6pk                                                   0/3       Pending            0          14m
kube-system   kube-flannel-77r0p                                                          2/2       Running            0          14m
kube-system   kube-flannel-vn3wt                                                          2/2       Running            1          14m
kube-system   kube-proxy-j8ftt                                                            1/1       Running            0          14m
kube-system   kube-proxy-xnhbs                                                            1/1       Running            0          14m
kube-system   kube-scheduler-2387324182-7xvqb                                             0/1       Pending            0          14m
kube-system   kube-scheduler-2387324182-tfwkk                                             0/1       Pending            0          14m
kube-system   pod-checkpointer-f73x6                                                      1/1       Running            0          14m
kube-system   pod-checkpointer-f73x6-jakku-controller.c.dghubble-io.internal              1/1       Running            0          14m
```

bootstrap controller manager:

```
NodeController detected that all Nodes are not-Ready. Entering master disruption mode.
I0424 06:30:59.560436       1 taint_controller.go:180] Starting NoExecuteTaintManager
...
NodeController detected that some Nodes are Ready. Exiting master disruption mode.
```
Pending control plane components list no nodes as the reason they cannot be scheduled, though nodes were registered successfully according to apiserver logs

bootstrap scheduler:

```
FailedScheduling        no nodes available to schedule pods
```

[...]

PR #464, opened 2017-04-24
pkg/asset/internal: Add --cloud-provider to bootstrap manifests

dghubble:
~This was a solution to #463, not necessarily the right solution. I'm not familiar with the original reasons we set `configure-cloud-routes=true` to false (upstream default is true) or its impact on other providers.~

* Add --cloud-provider to bootstrap apiserver and controller-manager (missing)
~* Set --configure-cloud-routes true. Upstream description: "Should CIDRs allocated by allocate-node-cidrs be configured on the cloud provider. (default true)"~

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -255,6 +255,7 @@ spec:
     - --secure-port=443
     - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
     - --service-cluster-ip-range={{ .ServiceCIDR }}
+    - --cloud-provider={{ .CloudProvider }}
     - --storage-backend=etcd3
     - --tls-ca-file=/etc/kubernetes/secrets/ca.crt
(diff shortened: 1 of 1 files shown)


dghubble:
test this please

diegs:
rktbot run tests

pkg/asset/internal/templates.go
```diff
@@ -439,7 +440,7 @@ spec:
         - --allocate-node-cidrs=true
         - --cloud-provider={{ .CloudProvider }}
         - --cluster-cidr={{ .PodCIDR }}
-        - --configure-cloud-routes=false
+        - --configure-cloud-routes=true
```
aaronlevy: We assume the use of flannel (or another CNI plugin) - so we do not want the cloud-provider to also be trying to configure routes.

dghubble:
@aaronlevy I've scoped back this change just to set --cloud-provider on bootstrap manifests. It will no longer solve #463 however.

dghubble:
cc @aaronlevy

PR #468, opened 2017-04-24
*: make vendor

hongchaodeng:
I did a simple `make vendor` without any change.. It already had these updates.


pbx0:
I'm not very familiar with glide, but I think you will want to edit `glide.yaml` by hand and pin etcd and the etcd-operator to specific released versions. Then run `make vendor`

hongchaodeng:
closing this since we are still figuring how to vendor etcd v316

PR #473, opened 2017-04-25
pkg/bootkube: Remove dependency on github.com/coreos/etcd/pkg/fileutil.

yifan-gu:
glide.yaml is not touched until another etcd denpendency is fixed.


aaronlevy:
rktbot run tests

PR #477, opened 2017-04-25
Fix up some logging configuration.

diegs:
- Use the same logging config for all bootkube commands.
- Make sure flag.Parse() is called, otherwise glog complains.
  I checked and cobra does not call flag.Parse().


dghubble:
:+1:

PR #479, opened 2017-04-26
Change temporary etcd port to 4001.

diegs:
This is an actual (though deprecated) etcd port, as opposed
to a completely random one. Some docs recommend using this
port for migrating from http to https, so it feels slightly
more natural to use it.


diegs:
After discussion, abandoning since this port is <10,000.

In the future we can explore using pod networking for self-hosted etcd nodes, which would remove the port conflict.

PR #480, opened 2017-04-28
checkpoint: Fix garbage collection bug.

yifan-gu:
The apiParentPods is always non-nil, so we should test the length
instead of the nil pointer. Besides, testing the length will work
for nil pointer as well.


chancez:
Good catch.

PR #487, opened 2017-05-01
terraform: Add terraform module for bootkube render

dghubble:
This terraform module generates bootkube assets which aim to provide the same variables, defaults, features, and generated output that the command line `bootkube render` does. In practical terms, terraform users can use the module via

```
terraform apply
```

to do the equivalent of:

```
bootkube render --asset-dir=assets --api-servers=https://node1.example.com:443 --api-server-alt-names=DNS=node1.example.com --etcd-servers=http://127.0.0.1:2379
```

### Why

Supporting a module like allows vanilla bootkube clusters to be automated and tested with `terraform apply`. It unlocks generating assets as part of the apply (as opposed to installing bootkube locally) and using those assets in post-processing (such as copying the assets to a controller).

It will be directly useful for `bootkube` GCE/AWS testing and `matchbox` bare-metal examples.

### TODO

Note, this module does not yet have feature parity with the bootkube binary.

* Experimental manifests
* etcd TLS
* Self-hosted etcd

### DIFF
--- a/terraform/.gitignore
+++ b/terraform/.gitignore
@@ -1,0 +1,2 @@
+terraform.tfvars
+*.tfstate*
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -1,0 +1,53 @@
+# Terraform
+
+This Terraform module generates bootkube assets, just like using the bootkube binary to run `bootkube render`. It aims to provide the same variables, defaults, features, and generated output that the command line tool does.
+
+Terraform-based clusters can generate bootkube assets as part of `terraform apply`.
+
+## TODO
+
+Note, this module does not have feature parity with `bootkube render` yet.
+
+* Experimental manifests
+* etcd TLS
+* Self-hosted etcd
+
+## Usage
+
+Use the terraform module within your terraform configs. Check `variables.tf` for required and optional input variables and `terraform.tfvars.example` for examples.
+
+```hcl
+module "bootkube" {
+  source = "git://https://github.com/coreos/bootkube.git//terraform"
+
+  cluster_name = "example"
+  api_servers = ["node1.example.com"]
+  etcd_servers = ["http://127.0.0.1:2379"]
+  asset_dir = "/home/core/clusters/mycluster"
+}
+```
+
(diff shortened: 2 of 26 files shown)


rmenn:
sorry quick question, how would this be different from the tectonic installer?

dghubble:
@rmenn Tectonic is the CoreOS enterprise Kubernetes offering which includes a multi-user Tectonic console, experimental operators, has a GUI installer, etc. Tectonic renders similar, but different assets (such as integrating with identity) while still using bootkube to start the control plane. It requires a license, but its free for up to 10 nodes.

For bootkube itself, hack examples and conformance tests are based on assets produced by `bootkube render`.

aaronlevy:
This is great, but my general view is that bootkube shouldn't be a canonical source for how you launch infrastructure - but instead a simple tool that is used as a building block of other cluster deployment tools (e.g. terraform, or ansible, or cloud-formation, etc.)

Along those lines I don't think the various installation methods should live inside bootkube itself. That is unless we were replacing `bootkube render` with terraform for the asset generation step - but I'd be pretty hesitant about that right now (terraform's rendering isn't super featured either).

I do think this could be helpful as a stand-alone project potentially, or maybe even as a module within tectonic-installer (e.g. vanilla kubernetes)?

aaronlevy:
Along these lines, however, there are plans on using terraform to launch the compute resources for CI builds -- but this will just be used to bring up nodes, bootkube will still use the internal `render` for asset generation. The terraform scripts will also just be for testing infra and not as a recommended way of launching a real cluster.

dghubble:
I think this terraform module is only intended to create asset files, never to create infrastructure. The value would be that terraform users can use that module with their own infrastructure bring-up solutions to automate vanilla bootkube cluster creation.

That's ok if its out of scope for now. It exists as a separate project https://github.com/dghubble/bootkube-terraform and I'll be using it to provide a better bare-metal experience for terraform users of matchbox so it should be kept in sync. And you know where to find me if it is wanted later.

dghubble:
After discussion, maintaining this as an independent project.

PR #489, opened 2017-05-01
hack/terraform-quickstart: initial commit

pbx0:
This adds a basic way to use terraform to setup AWS compute along with the
quickstart scripts to initiate a cluster. This is designed for forming
the basis for test setup.

### DIFF
--- a/hack/terraform-quickstart/outputs.tf
+++ b/hack/terraform-quickstart/outputs.tf
@@ -1,0 +1,11 @@
+output "bootstrap_node_ip" {
+  value = "${aws_instance.bootstrap_node.public_ip}"
+}
+
+output "worker_ips" {
+  value = ["${aws_instance.worker_node.*.public_ip}"]
+}
+
+output "self_host_etcd" {
+  value = "${var.self_host_etcd}"
+}
--- a/hack/terraform-quickstart/start-cluster.sh
+++ b/hack/terraform-quickstart/start-cluster.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -euo pipefail
+
+export BOOTSTRAP_IP=`terraform output bootstrap_node_ip`
+export WORKER_IPS=`terraform output -json worker_ips | jq -r '.value[]'`
+export SELF_HOST_ETCD=`terraform output self_host_etcd`
+export SSH_OPTS="-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
+
+cd ../quickstart
+./init-master.sh $BOOTSTRAP_IP
+
+for IP in $WORKER_IPS
+do
+  ./init-worker.sh $IP cluster/auth/kubeconfig
+done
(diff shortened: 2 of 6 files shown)


hack/terraform-quickstart/resources.tf
```diff
@@ -0,0 +1,49 @@
+provider "aws" {
+  access_key = "${var.access_key_id}"
+  secret_key = "${var.access_key}"
+  region     = "us-east-1"
```
Quentin-M: why not configurable too 👍

hack/terraform-quickstart/resources.tf
```diff
@@ -0,0 +1,49 @@
...
+  ami           = "${data.aws_ami.coreos_ami.image_id}"
+  instance_type = "m3.medium"
+  key_name      = "${var.ssh_key}"
+
+  tags {
+    Name = "${var.instance_tags}"
+  }
+}
+
+resource "aws_instance" "worker_node" {
```
Quentin-M: you can use `count` to have multiple master/worker nodes!

hack/terraform-quickstart/start-cluster.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
Quentin-M: you may have TerraForm execute this for you if you wish.

hack/terraform-quickstart/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "bootstrap_node_ip" {
+  value = "${aws_instance.bootstrap_node.public_ip}"
+}
+
+output "worker_ips" {
```
Quentin-M: if you only have one worker, it should be singular I guess!

hack/terraform-quickstart/resources.tf
pbx0: Ah this will make it simpler to set a variable that determines the number of worker nodes.

hack/terraform-quickstart/outputs.tf
pbx0: I want this to be an array of IPs for the start-cluster.sh script to consume. So if I add the `count` field to the worker resource I think this will work out.

hack/terraform-quickstart/start-cluster.sh
pbx0: I don't have a strong desire to make this the case, but it would be nice. I looked at https://www.terraform.io/docs/provisioners/local-exec.html but I didn't see an easy way to trigger a script specifically until all the resources were up. It looks like in that example its tied to a specific resource.

hack/terraform-quickstart/resources.tf
pbx0: Yeah, I suppose why not. I'll add a default for this one.

pbx0:
Addressed feedback.  Number of worker nodes is now input variable.

hack/terraform-quickstart/start-cluster.sh
Quentin-M: There are several ways to define a dependency in TerraForm, but that's not like it needs to be done :) Just suggesting, maybe for later~

pbx0:
Thanks for the review!

Issue #488, opened 2017-05-01
etcd should use pod anti-affinity


sym3tri:
so multiple instances don't end up on same node

aaronlevy:
This is already true (etcd-operator will enforce this for self-hosted etcd pods)

PR #490, opened 2017-05-02
[do not merge] pkg/util: turn on antiAffinity for etcd pods

aaronlevy:
fixes https://github.com/kubernetes-incubator/bootkube/issues/488


aaronlevy:
Do not merge. This might be unneeded

hongchaodeng:
For self-hosted etcd, it has anti-affinity across nodes by default.
Because if two etcd pods stay on the same node, they will fight for port 2379.

aaronlevy:
Closing as anti-affinity is already set

Issue #432, opened 2017-04-14
Documentation: Disaster recovery scenarios


aaronlevy:
We should document recovery of:

- partial control-plane loss
    - Discussed in: https://github.com/kubernetes-incubator/bootkube/issues/112
- loss of all api-servers
    - Recover from a backup of the api-server checkpoint data (document the backup process as well)
- manual recovery from etcd backup (external etcd)
    - Extracting api-server manifest / secret data from etcd, then recovering using temporary static pods.
- manual recovery from etcd backup (self-hosted etcd).
    - Extracting etcd manifest & api-server manifests, then recovering using temporary static pods.

aaronlevy:
To expand on some of this a bit - and include strawman discussion of building a tool to handle some of the recovery mechanics:

## Partial loss of control plane:

We are assuming we still have running apiserver + etcd - and we just need to recover other control plane components that are no longer scheduled for some reason. More specific discussion is in https://github.com/kubernetes-incubator/bootkube/issues/112 -- but the pseudo process / UX might be:

```
bootkube recover pod-from deployment/kube-scheduler --target=node
```

or maybe

```
bootkube recover pod-from deployment/kube-scheduler --target-label=node-role.kubernetes.io/master
```

## loss of all api-servers

We have two main options here. One option would assume that we are also backing up checkpoints of "critical" pods (e.g. apiserver + anything it relies on like secrets). This seems relatively reasonable - but does make the backup/restore story have a few more moving pieces.

Another option would be to only expect access to etcd - and attempt recovery from there. This would be my initial preference because technically all needed state should exist directly in etcd.

The process would essentially be like a partial `bootkube start`, but you're pointing it at an existing etcd cluster that already has state that you want to extract:

All pseudo UX: 

* Extract the api-server manifest from etcd directly

`bootkube recover pod-from daemonset/kube-apiserver --etcd-server=http://foo --target=/etc/kubernetes/manifests/kube-apiserver.yaml`

This would:

1. Extract apiserver daemonset manifest + all secrets / configMaps directly from etcd
1. Place assets on disk (e.g. secrets/configmaps) then place the static pod at target
1. Wait for self-hosted apiserver to recover, then delete assets off disk?

The above UI might be trying to overload functionality too much -- we could also make this more explicit rather than generic:

`bootkube recover apiserver --etcd-server=http://foo --target=/etc/kubernetes/manifests`

## Recovery from etcd backup (external etcd)

[...]

PR #491, opened 2017-05-03
Preliminary recovery command for bootkube.

diegs:
The new `bootkube recover` command provides a framework for reading
control plane manifests out of a storage backend, serializing them to
disk, and constructing pod manifests for the bootstrap components that
are usable with `bootkube start` to re-bootstrap the control plane.

The command initially provides and etcdclient backend for use with a
running etcd cluster. In the future we can add support other methods,
such as an etcd backup directory or a running apiserver.

There is also documentation and an example script in hack/multi-node
that deletes a master node, recovers the assets from etcd, and then
re-bootstraps the control plane.

This is a first step in addressing the issues described in #432.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ bootkube start --asset-dir=my-cluster
 
 ### Recover a downed cluster
 
-In the case of a partial or total control plane outage (i.e. due to lost master nodes) an experimental `recover` command can extract and write manifests from a backup location. These manifests can then be used by the `start` to reboot the cluster. Currently recovery from a running etcd cluster is the only supported method.
+In the case of a partial or total control plane outage (i.e. due to lost master nodes) an experimental `recover` command can extract and write manifests from a backup location. These manifests can then be used by the `start` command to reboot the cluster. Currently recovery from an external running etcd cluster is the only supported method.
 
 To see available options, run:
 
@@ -72,7 +72,7 @@ Example:
 bootkube recover --asset-dir=recovered --etcd-servers=http://127.0.0.1:2379 --kubeconfig=/etc/kubernetes/kubeconfig
 ```
 
-For a complete recovery example please see the [hack/multi-node/bootkube-recover](hack/multi-node/bootkube-recover) script.
+For a complete recovery example please see the [hack/multi-node/bootkube-test-recovery](hack/multi-node/bootkube-test-recovery) script.
 
 ## Building
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ If you are interested in the design and details [see the Kubernetes self-hosted 
 
 ## Usage
 
-Bootkube has two main commands: `render` and `start`
+Bootkube has two main commands: `render` and `start`.
+
+There is a third, experimental command `recover` which can help reboot a downed cluster (see below).
 
 ### Render assets
 
@@ -53,6 +55,24 @@ Example:
 ```
 bootkube start --asset-dir=my-cluster
 ```
+
+### Recover a downed cluster
+
+In the case of a partial or total control plane outage (i.e. due to lost master nodes) an experimental `recover` command can extract and write manifests from a backup location. These manifests can then be used by the `start` to reboot the cluster. Currently recovery from a running etcd cluster is the only supported method.
+
+To see available options, run:
+
+```
+bootkube recover --help
+```
+
+Example:
+
+```
+bootkube recover --asset-dir=recovered --etcd-servers=http://127.0.0.1:2379 --kubeconfig=/etc/kubernetes/kubeconfig
+```
+
+For a complete recovery example please see the [hack/multi-node/bootkube-recover](hack/multi-node/bootkube-recover) script.
 
 ## Building
 
--- a/cmd/bootkube/recover.go
+++ b/cmd/bootkube/recover.go
@@ -11,8 +11,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/coreos/etcd/clientv3"
 	"github.com/kubernetes-incubator/bootkube/pkg/recovery"
+
+	"github.com/coreos/etcd/clientv3"
 	"github.com/spf13/cobra"
 )
 
@@ -43,7 +44,7 @@ func init() {
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdCAPath, "etcd-ca-path", "", "Path to an existing PEM encoded CA that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-certificate-path and --etcd-private-key-path, and must have etcd configured to use TLS with matching secrets.")
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdCertificatePath, "etcd-certificate-path", "", "Path to an existing certificate that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-private-key-path, and must have etcd configured to use TLS with matching secrets.")
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdPrivateKeyPath, "etcd-private-key-path", "", "Path to an existing private key that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-certificate-path, and must have etcd configured to use TLS with matching secrets.")
-	cmdRecover.Flags().StringVar(&recoverOpts.etcdServers, "etcd-servers", "", "List of etcd servers URLs including host:port, comma separated. Cannot be specified with --etcd-data-dir.")
+	cmdRecover.Flags().StringVar(&recoverOpts.etcdServers, "etcd-servers", "", "List of etcd servers URLs including host:port, comma separated.")
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdPrefix, "etcd-prefix", "/registry", "Path prefix to Kubernetes cluster data in etcd.")
 	cmdRecover.Flags().StringVar(&recoverOpts.kubeConfigPath, "kubeconfig", "", "Path to kubeconfig for communicating with the cluster.")
 }
@@ -67,16 +68,16 @@ func runCmdRecover(cmd *cobra.Command, args []string) error {
 
 func validateRecoverOpts(cmd *cobra.Command, args []string) error {
 	if recoverOpts.assetDir == "" {
-		return errors.New("Missing required flag: --asset-dir")
+		return errors.New("missing required flag: --asset-dir")
 	}
 	if (recoverOpts.etcdCAPath != "" || recoverOpts.etcdCertificatePath != "" || recoverOpts.etcdPrivateKeyPath != "") && (recoverOpts.etcdCAPath == "" || recoverOpts.etcdCertificatePath == "" || recoverOpts.etcdPrivateKeyPath == "") {
-		return errors.New("You must specify either all or none of --etcd-ca-path, --etcd-certificate-path, and --etcd-private-key-path")
+		return errors.New("you must specify either all or none of --etcd-ca-path, --etcd-certificate-path, and --etcd-private-key-path")
 	}
 	if recoverOpts.etcdPrefix == "" {
-		return errors.New("Missing required flag: --etcd-prefix")
+		return errors.New("missing required flag: --etcd-prefix")
 	}
 	if recoverOpts.kubeConfigPath == "" {
-		return errors.New("Missing required flag: --kubeconfig")
+		return errors.New("missing required flag: --kubeconfig")
 	}
 	return nil
 }
--- a/cmd/bootkube/recover.go
+++ b/cmd/bootkube/recover.go
@@ -1,0 +1,108 @@
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/kubernetes-incubator/bootkube/pkg/recovery"
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdRecover = &cobra.Command{
+		Use:          "recover",
+		Short:        "Recover a control plane from state stored in etcd.",
+		Long:         "",
+		PreRunE:      validateRecoverOpts,
+		RunE:         runCmdRecover,
+		SilenceUsage: true,
+	}
+
+	recoverOpts struct {
+		assetDir            string
+		etcdCAPath          string
+		etcdCertificatePath string
+		etcdPrivateKeyPath  string
+		etcdServers         string
+		etcdPrefix          string
+		kubeConfigPath      string
+	}
+)
+
+func init() {
+	cmdRoot.AddCommand(cmdRecover)
+	cmdRecover.Flags().StringVar(&recoverOpts.assetDir, "asset-dir", "", "Output path for writing recovered cluster assets.")
+	cmdRecover.Flags().StringVar(&recoverOpts.etcdCAPath, "etcd-ca-path", "", "Path to an existing PEM encoded CA that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-certificate-path and --etcd-private-key-path, and must have etcd configured to use TLS with matching secrets.")
+	cmdRecover.Flags().StringVar(&recoverOpts.etcdCertificatePath, "etcd-certificate-path", "", "Path to an existing certificate that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-private-key-path, and must have etcd configured to use TLS with matching secrets.")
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -54,6 +56,24 @@ Example:
 bootkube start --asset-dir=my-cluster
 ```
 
+### Recover a downed cluster
+
+In the case of a partial or total control plane outage (i.e. due to lost master nodes) an experimental `recover` command can extract and write manifests from a backup location. These manifests can then be used by the `start` to reboot the cluster. Currently recovery from a running etcd cluster is the only supported method.
```
xiang90: from an external running etcd cluster?

cmd/bootkube/recover.go
```diff
@@ -0,0 +1,108 @@
...
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/kubernetes-incubator/bootkube/pkg/recovery"
```
xiang90: separate internal and external imports?

cmd/bootkube/recover.go
```diff
@@ -0,0 +1,108 @@
...
+	as, err := recovery.Recover(context.Background(), recovery.NewEtcdBackend(etcdClient, recoverOpts.etcdPrefix), recoverOpts.kubeConfigPath)
+	if err != nil {
+		return err
+	}
+	return as.WriteFiles(recoverOpts.assetDir)
+}
+
+func validateRecoverOpts(cmd *cobra.Command, args []string) error {
+	if recoverOpts.assetDir == "" {
+		return errors.New("Missing required flag: --asset-dir")
```
xiang90: missing required flag (error message starts with lower letter)

pkg/recovery/etcd.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"context"
+	"fmt"
+	"path"
+	"reflect"
+	"strings"
+
+	"github.com/coreos/etcd/clientv3"
+
```
xiang90: remove this empty line?

pkg/recovery/etcd.go
```diff
@@ -0,0 +1,147 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/conversion"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/pkg/api"
+)
+
+// EtcdBackend is a backend that extracts a controlPlane from an etcd instance.
+type EtcdBackend struct {
```
xiang90: does this need to be a public struct?

pkg/recovery/etcd.go
```diff
@@ -0,0 +1,147 @@
...
+		decoder:    api.Codecs.UniversalDecoder(),
+		pathPrefix: pathPrefix,
+	}
+}
+
+// read implements Backend.read().
+func (s *EtcdBackend) read(ctx context.Context) (*controlPlane, error) {
+	cp := &controlPlane{}
+	for _, r := range []struct {
+		etcdName string
```
xiang90: etcdKeyName?

pkg/recovery/etcd.go
```diff
@@ -0,0 +1,147 @@
...
+
+	elems := make([][]byte, len(getResp.Kvs))
+	for i, kv := range getResp.Kvs {
+		elems[i] = kv.Value
+	}
+	return decodeList(elems, listPtr, s.decoder)
+}
+
+// decode decodes value of bytes into object.
+func decode(decoder runtime.Decoder, value []byte, objPtr runtime.Object) error {
```
xiang90: this func is not really specific to etcd backend. probably it should live in recovery/util.go

pkg/recovery/recover.go
```diff
@@ -0,0 +1,332 @@
...
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"path"
+	"reflect"
+	"strings"
+
+	"github.com/ghodss/yaml"
+
```
xiang90: remove the empty line here

pkg/recovery/recover.go
```diff
@@ -0,0 +1,332 @@
...
+)
+
+const (
+	kubeletKubeConfigPath = "/etc/kubernetes/kubeconfig"
+)
+
+var (
+	// bootstrapComponents contains the names of the components that we will extract to construct the
+	// temporary bootstrap control plane.
+	bootstrapComponents = map[string]bool{
```
xiang90: is this a set? or a map with boolean value? if it is a set, probably map[string]struct{} is more clear.

pkg/recovery/recover.go
```diff
@@ -0,0 +1,332 @@
...
+	addTypeMeta(&v1beta1.Deployment{}, v1beta1.SchemeGroupVersion)
+	addTypeMeta(&policyv1beta1.PodDisruptionBudget{}, policyv1beta1.SchemeGroupVersion)
+	addTypeMeta(&v1.Pod{}, v1.SchemeGroupVersion)
+	addTypeMeta(&v1.Secret{}, v1.SchemeGroupVersion)
+	addTypeMeta(&v1.Service{}, v1.SchemeGroupVersion)
+	addTypeMeta(&v1.ServiceAccount{}, v1.SchemeGroupVersion)
+}
+
+// Backend defines an interface for any backend that can populate a controlPlane struct.
+type Backend interface {
```
xiang90: why this interface needs to be public if all the methods are private?

pkg/recovery/recover.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	return as, nil
+}
+
+// renderBootstrap returns assets for a bootstrap control plane that can be used with `bootkube
+// start` to re-bootstrap a control plane. These assets are derived from the self-hosted control
+// plane that was recovered by the backend, but modified for direct injection into a kubelet.
+func (cp *controlPlane) renderBootstrap(kubeConfigPath string) (asset.Assets, error) {
+	// Extract pod specs from daemonsets and deployments.
+	var pods []*v1.Pod
```
xiang90: can we separate this large func to:

1. extractPods
2. fixupPods

?

pkg/recovery/recover.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+		}
+
+		// Add a mount for the kubeconfig.
+		pod.Spec.Volumes = append(pod.Spec.Volumes, v1.Volume{
+			VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: kubeletKubeConfigPath}},
+			Name:         "kubeconfig",
+		})
+
+		// Output the pod definition.
```
xiang90: this does not belong to pod fixup i feel. or we can change the comment at line 187.

pkg/recovery/etcd.go
```diff
@@ -0,0 +1,147 @@
...
+		obj:      &cp.secrets,
+	}, {
+		etcdName: "services/specs",
+		yamlName: "service",
+		obj:      &cp.services,
+	}, {
+		etcdName: "serviceaccounts",
+		yamlName: "service-account",
+		obj:      &cp.serviceAccounts,
+	},
```
aaronlevy: I would assume that we don't care about a few of these object types. If we're just trying to re-bootstrap a control-plane, we should only need daemonset/deployment (to extract podSpecs) and secrets/configmaps (converted to files/host mounts)

pkg/recovery/recover.go
```diff
@@ -0,0 +1,332 @@
...
+)
+
+var (
+	// bootstrapComponents contains the names of the components that we will extract to construct the
+	// temporary bootstrap control plane.
+	bootstrapComponents = map[string]bool{
+		"kube-apiserver":          true,
+		"kube-controller-manager": true,
+		"kube-scheduler":          true,
+		"pod-checkpointer":        true,
```
aaronlevy: I don't think we need the checkpointer to re-bootstrap the control plane (and once a control-plane is running, it will get scheduled normally).

pkg/recovery/recover.go
```diff
@@ -0,0 +1,332 @@
...
+		// Output the pod definition.
+		a, err := serializeYAML(path.Join(asset.AssetPathBootstrapManifests, pod.Name+".yaml"), pod)
+		if err != nil {
+			return nil, err
+		}
+		as = append(as, a)
+	}
+
+	// Output all the required secrets.
+	for _, secret := range cp.secrets.Items {
```
aaronlevy: We likely need to do the same thing for configMaps (change to file + hostmount). We don't use them by default - but a user could have added them to the manifests.

pkg/recovery/etcd.go
aaronlevy: Actually, looks to be slightly different behavior than I assumed.

On the one hand - the behavior of "extract everything" would enable stamping out cookie-cutter clusters from this source pretty easy (and this could be a nice feature). However, from a recovery standpoint, this feels pretty heavy-handed (why extract everything just to re-inject it where it already exists?).

From the purely recovery standpoint, I would think we only care about extracting enough to run a temporary control-plane - such that we get past the self-hosted chicken/egg issue and re-launch the cluster. All other assets will still exist and once the self-hosted control-plane is brought up - everything else should recover naturally.

The other situation we might want to address would be the potential for recovering from "bad" configuration that took down the control-plane (for whatever reason). This would mean we would need to extract some manifests to allow them to be modified, and then re-inject the fixed manifests into the cluster. However, I feel like this even could be scoped down to just core components (api, scheduler, controller-manager, proxy). This would also require a change to `bootkube start` such that it would replace existing assets (I believe `create` will error if object already exists).

hack/multi-node/bootkube-recover
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+GLOG_v=${GLOG_v:-1}
+HOST=${HOST:-c1}
+SELF_HOST_ETCD=${SELF_HOST_ETCD:-false}
+
+if [ ${SELF_HOST_ETCD} = "true" ]; then
```
aaronlevy: for clarity - not saying we should remove - just advocating for clearer script name

README.md
diegs: Done

cmd/bootkube/recover.go
diegs: Done

cmd/bootkube/recover.go
diegs: Done for all (I was following the other command definitions, but personally I prefer the go convention)

pkg/recovery/recover.go
diegs: Good point. Ok with a TODO for now?

pkg/recovery/etcd.go
aaronlevy: I guess my leaning toward being more specific about what we're recovering is that this wouldn't technically cover everything in a cluster - as we're only recovering some of the assets (e.g. there could scheduled jobs, stateful sets, namespaces, rbac rules, others that I'm not thinking of).

If we were really trying to recover all cluster assets / stamp out a new cluster (now that I'm revisiting this idea) we would need to get all the objects that *could* exist. Even then - we would need to introduce ordering to object creation (e.g. namespaces need to be created prior to objects that exist in those namespaces, or TPR types before instances of those types).

Regarding, recovering from etcd-backup - my assumption is that we would only support this for self-hosted etcd(?) - in which case the data would already be replicated into the cluster (essentially by re-executing the pivot into a new cluster), so we wouldn't necessarily need to re-create the assets (they're still there).

Ultimately, dynamically extracting all object types - then enforcing ordering constraints isn't insurmountable -- but it does seem like a decent amount of complexity vs starting out with a more focused set of assets to recover.

hack/multi-node/bootkube-recover
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
aaronlevy: sgtm

pkg/recovery/recover.go
aaronlevy: Todo is fine - but we should have a follow-up soon supporting configMaps. It's possible that the control-plane requires them (then the recovery wouldn't work).

pkg/recovery/recover.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	as = append(as, secrets...)
+	return as, nil
+}
+
+// extractBootstrapPods extracts bootstrap pod specs from daemonsets and deployments.
+func extractBootstrapPods(daemonSets []v1beta1.DaemonSet, deployments []v1beta1.Deployment) ([]v1.Pod, error) {
+	var pods []v1.Pod
+	for _, ds := range daemonSets {
+		if componentName := ds.Labels["component"]; bootstrapComponents[componentName] {
```
aaronlevy: We changed "component" to "k8s-app" to try and be more consistent (https://github.com/kubernetes-incubator/bootkube/pull/492) -- so I think this needs to be updated.

Also, for clarity might be better to define as a const.

pkg/recovery/recover.go
```diff
@@ -0,0 +1,370 @@
...
+			pod := v1.Pod{Spec: ds.Spec.Template.Spec}
+			if err := setBootstrapPodMetadata(&pod, ds.ObjectMeta); err != nil {
+				return nil, err
+			}
+			pods = append(pods, pod)
+		}
+	}
+	for _, ds := range deployments {
+		if componentName := ds.Labels["component"]; bootstrapComponents[componentName] {
```
aaronlevy: Same as above

pkg/recovery/etcd.go
```diff
@@ -0,0 +1,116 @@
...
+		yamlName:    "daemonset",
+		obj:         &cp.daemonSets,
+	}, {
+		etcdKeyName: "deployments",
+		yamlName:    "deployment",
+		obj:         &cp.deployments,
+	}, {
+		etcdKeyName: "poddisruptionbudgets",
+		yamlName:    "pod-disruption-budget",
+		obj:         &cp.podDisruptionBudgets,
```
aaronlevy: This can be dropped, we wouldn't need this to recover a control-plane (just an added protection from accidentally destroying one - but they should still exist in etcd anyway).

pkg/recovery/etcd.go
```diff
@@ -0,0 +1,116 @@
...
+		yamlName:    "secret",
+		obj:         &cp.secrets,
+	}, {
+		etcdKeyName: "services/specs",
+		yamlName:    "service",
+		obj:         &cp.services,
+	}, {
+		etcdKeyName: "serviceaccounts",
+		yamlName:    "service-account",
+		obj:         &cp.serviceAccounts,
```
aaronlevy: These could also be dropped - they will be auto-created as needed (by controller-manager) / already exist.

However, if we didn't want to require the use of a kubeconfig... we could use this instead (the service account will have the CA to trust as well as a signed bearer token, which can be used to auth against api).

That could be a follow-up to explore, as it does remove one more moving piece (kubeconfig).

pkg/recovery/recover.go
aaronlevy: Also based on the changes to labels -- this tool will only work if the cluster was launched with a particular version of bootkube (or a particular version of tectonic). Kind of makes me think that the label should be configurable? Or maybe we try both? Or we at least need to document...

pkg/recovery/recover.go
```diff
@@ -0,0 +1,370 @@
+// Package recovery provides tooling to help with control plane disaster recovery. Recover() uses a
```
aaronlevy: A lot of the helper functions in this file are good candidates for unit tests. Can be a todo/follow-up - but we haven't been super good in the past with test coverage on bootkube.

pkg/recovery/recover.go
```diff
@@ -0,0 +1,370 @@
...
+		Name:        "bootstrap-" + parent.Name,
+		Namespace:   parent.Namespace,
+	}
+	return nil
+}
+
+// fixUpBootstrapPods modifies extracted bootstrap pod specs to have correct metadata and
+// point to filesystem-mount-based secrets. It returns a set of secrets that must also be rendered
+// in order for the bootstrap pods to be functional.
+// TODO(diegs): also output the set of reqiured configMaps.
```
aaronlevy: After this merges, lets open a tracking issue for this

pkg/recovery/recover.go
diegs: Added a bunch of tests. It's not perfect but we can keep adding/updating as we improve the tool.

pkg/recovery/recover.go
diegs: SG. I was going to go ahead and add it but it's a bit of work, will follow up immediately after this PR.

pkg/recovery/etcd.go
diegs: I changed it to only extract the bootstrap manifests, and added a little bit of code in `bootkube start` to suppress a scary-sounding error if there are no self-hosted manifests to install.

pkg/recovery/etcd.go
diegs: Removed.

pkg/recovery/etcd.go
diegs: Removed for now, but took a note to explore in the near future.

pkg/recovery/recover.go
diegs: Done

pkg/recovery/recover.go
diegs: I'd say configurable might not be great, since we rely on other "bootkube conventions" such as various values we use for those labels.

Instead I added support for both the old and new label key, extracted to a function. Thanks for the nice catch.

PR #492, opened 2017-05-03
pkg/asset/internal: standardize around k8s-app label

pbx0:


pbx0:
We'll wait until the tectonic one goes in before merging this. Also the only tests that failed are the api-server deletion ones because the test relies on `component` labels. Will update the tests after this merges.

pbx0:
Tectonic one is in and they tested api-server deletion successful which is the one part of our suite down because of label change. So this should be OK to merge.

PR #493, opened 2017-05-03
templates: update etcd-operator to v0.2.6

hongchaodeng:
There is a change for self hosted etcd pods that make it more resilient
on anti-affinity guarantee. It doesn't break existing cluster, but have
stronger guarantee.

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -865,7 +865,7 @@ spec:
     spec:
       containers:
       - name: etcd-operator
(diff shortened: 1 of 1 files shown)


hongchaodeng:
rktbot run tests

hongchaodeng:
rktbot run tests

pbx0:
Looks like this release supports TLS as well. https://github.com/coreos/etcd-operator/blob/master/doc/user/cluster_tls.md

PR #494, opened 2017-05-05
Extract all image versions from templates.

diegs:
I just specified the values in a constant for now, but we can do
something more sophisticated if we want.


diegs:
rktbot run tests

diegs:
rktbot run tests

diegs:
rktbot run tests

PR #496, opened 2017-05-06
Move etcd TPR generation to render phase.

### LINKED ISSUE #476: etcd cluster TPR should be generated as part of the render step
Rather than hard-coded: https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/util/etcdutil/migrate.go#L121-L145

diegs:
This requires putting the boot etcd server behind a service so it can
have a pre-determined IP address (before this change the pod IP was
being dynamically detected).

Fixes #476


cmd/bootkube/render.go
```diff
@@ -21,6 +21,7 @@ const (
 	apiOffset                    = 1
 	dnsOffset                    = 10
 	etcdOffset                   = 15
+	bootEtcdOffset               = 200
```
hongchaodeng: What is this offset?

pkg/asset/internal/templates.go
```diff
@@ -913,21 +913,71 @@ spec:
     - --name=boot-etcd
     - --listen-client-urls=http://0.0.0.0:12379
     - --listen-peer-urls=http://0.0.0.0:12380
-    - --advertise-client-urls=http://$(MY_POD_IP):12379
-    - --initial-advertise-peer-urls=http://$(MY_POD_IP):12380
-    - --initial-cluster=boot-etcd=http://$(MY_POD_IP):12380
+    - --advertise-client-urls=http://{{ .BootEtcdServiceIP }}:12379
```
hongchaodeng: Why change this?

cmd/bootkube/render.go
diegs: It's just a safe place to put the temporary etcd service in the service CIDR.

pkg/asset/internal/templates.go
diegs: The self-hosted etcd cluster is now configured to migrate the boot etcd through its service IP instead of its actual (pod) IP. That way it doesn't have to look up the pod IP when doing the migration (which was kind of a hack)

cmd/bootkube/render.go
hongchaodeng: Why boot etcd needs a service?

pkg/asset/internal/templates.go
hongchaodeng: Why not keep using the downward API? Why change it?

diegs: Can you explain "downward API"? We would like to have all manifests created by `bootkube render`. This one was hardcoded into the etcd migration code.

Moving all assets into the render phase makes bootkube more flexible; users can customize the manifests before they are used by `bootkube start`.

cmd/bootkube/render.go
diegs: For below, if we render the TPR, which points to the migration cluster, ahead of time, we don't know what the pod IP will be. So instead we put it behind a service, which allows us to reserve an IP address ahead of time.

diegs:
This succeeds locally but is failing on pluton. I am going to investigate the test failures so please hold off on further reviews for the moment. Thanks!

cmd/bootkube/render.go
hongchaodeng: > we don't know what the pod IP will be

That's why we should use downward API. First of all, it doesn't need to reserve a service IP. Second, it's more generic on k8s platform than this method.

pkg/asset/internal/templates.go
hongchaodeng: > We would like to have all manifests created by bootkube render. This one was hardcoded into the etcd migration code.

Isn't the current code doing that? Why change it?

diegs: The current code compiles in the template here:

https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/util/etcdutil/migrate.go#L121-L145

Tracking issue is #476.

hongchaodeng: Yeah. I agree that the TPR manifest could be rendered. The discussion in this is about this specific line of change:

```
-    - --advertise-client-urls=http://$(MY_POD_IP):12379
+    - --advertise-client-urls=http://{{ .BootEtcdServiceIP }}:12379
```

aaronlevy: IIUC the underlying issue is that if these flags need to match how the etcd-operator reaches out to the seed-etcd, otherwise it will fail.

If we want to be able to render all the manifests we need for bootstrap (prior to ever running `bootkube start`), then we won't know what this IP will end up as - and can't pre-generate / populate the etcd TPR here: https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/util/etcdutil/migrate.go#L143

The service ip would allow us to use a consistent address between the etcd-operator and the seed so that we are able to pre-render all of the needed assets.

@hongchaodeng open to other ideas, but (assuming my understanding is correct) leaving this using the downward api means that the address populated in the etcd TPR will be incorrect.

The only other option I could see would be to enforce that the etcd-operator run on the same node as the bootstrap (or run a temp bootstrap etcd-operator as static pod) and then just use localhost?

hongchaodeng: > If we want to be able to render all the manifests we need for bootstrap (prior to ever running bootkube start), then we won't know what this IP will end up

We could query the API server for boot etcd's IP.

diegs: By we you mean the etcd-operator, right?

When the `render` phase runs no apiserver is running yet. We are laying out all of the bootstrap and self-hosted components ahead of time.

diegs: (In any event, though I've gotten this working on my own clusters the tests don't pass and I'm not sure why--so we can delay this for a bit until I figure that out, unless you have another suggested approach).

hongchaodeng: I see...

pkg/util/etcdutil/migrate.go
```diff
@@ -91,72 +89,29 @@ func waitEtcdTPRReady(restClient restclient.Interface, interval, timeout time.Du
...
-		}
-		return true, nil
-	})
-	return ip, err
+func createBootstrapEtcdService(restclient restclient.Interface, svcPath string) error {
+	svc, err := ioutil.ReadFile(svcPath)
+	if err != nil {
+		return err
+	}
+	return restclient.Post().RequestURI("/api/v1/namespaces/kube-system/services").SetHeader("Content-Type", "application/json").Body(svc).Do().Error()
```
aaronlevy: nit: We use the restclient for the TPR because the normal client isn't aware of the underlying type - but for services we could use something like `client.Core().Services().Update(service)`. Don't really need to change this, however. v1/service rest endpoint is not going anywhere.

pkg/util/etcdutil/migrate.go
```diff
@@ -208,6 +163,9 @@ func waitBootEtcdRemoved(etcdServiceIP string) error {
 	return err
 }
 
-func makeEtcdClusterURI(name string) string {
-	return fmt.Sprintf("/apis/%s/%s/namespaces/kube-system/%s/%s", spec.TPRGroup, spec.TPRVersion, spec.TPRKindPlural, name)
+func cleanupBootstrapEtcdService(restclient restclient.Interface) {
+	err := restclient.Delete().RequestURI("/api/v1/namespaces/kube-system/services/bootstrap-etcd-service").Do().Error()
```
aaronlevy: nit: It's a little fragile to rely on a fixed name here (bootstrap-etcd-service). It's possible the service manifest happened to have a diff name. If we load the actual object from the file, this would be easy enough to introspect prior to creating (and could also use regular client vs restclient). But this could also be a TODO

pkg/asset/internal/templates.go
```diff
@@ -913,21 +913,71 @@ spec:
...
+      {
+        "name": "peers",
+        "port": 12380,
+        "protocol": "TCP"
+      }
+    ]
+  }
+}`)
+
+	EtcdTPRTemplate = []byte(`{
```
xiang90: just a note: this needs to be regenerated for self hosted etcd recovery.

diegs:
Ok, I was able to debug this and found that the `bootstrap-etcd-service` was taking some time to become available. I added a wait loop and that appears to make the tests pass. PTAL when you have a moment. Thanks!

aaronlevy:
rktbot run tests

aaronlevy:
Also squash-before merge (and wait for tests being green...)

rimusz:
would really love to see this one merged

rimusz:
🥇

klausenbusk:
I just experienced a issue with the "hard-coding" of the `BootEtcdServiceIP`. When recovering the cluster (`bootkube start`, I got a `clusterIP` already in use conflict, as it was already allocated to another service, I ended up using a while loop to delete that service in my cluster as the first thing when `bootstrap-kube-apiserver` came up.


This is maybe a stupid question, but why can't we just set `bootMemberClientEndpoint` to `bootstrap-etcd-service.default`? and set `clusterIP: none` in the `bootstrap-etcd-service` service? Wouldn't that work?

PR #497, opened 2017-05-09
Add recovery support for ConfigMaps.

diegs:
This change also extracts and fixes up `ConfigMaps`.

It does not support the per-config-map mode bits that
`ConfigMapVolumeSource` supports.


diegs:
Note: I can't do a diffbase across my two branches, so this includes everything for now. Probably best to hold off reviewing until the upstream PR is merged.

PR #501, opened 2017-05-10
e2e: start of test framework

pbx0:
This commit forms the basis for moving tests in-repo, using the go test framework and a kubeconfig. Next steps would be moving over pluton tests, though much of the destructive abilities of those tests still need implementations through the k8s api. Longer-term this functionality may be implemented through some type of daemon-set. For now we can hack together what we need to reboot nodes and the like.

Looking for feedback on the basis of this before I get too far into porting tests.

cc @aaronlevy @derekparker @yifan-gu @diegs

### DIFF
--- a/e2e/framework/cluster.go
+++ b/e2e/framework/cluster.go
@@ -1,0 +1,71 @@
+package framework
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// Cluster represents an interface to test bootkube based kubernetes clusters.
+// The simplest way to write tests against it is to just write go tests that call
+// out to NewCluster at the beginning of each test. External tooling should first
+// provision the kubernetes cluster and output a kubeconfig.
+type Cluster struct {
+	Client kubernetes.Clientset
+
+	expectedNodes int
+}
+
+func NewCluster() *Cluster {
+	// uses the current context in kubeconfig
+	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	c := &Cluster{
+		Client:        *clientset,
+		expectedNodes: *expectedNodes,
+	}
+
+	return c
+}
+
+// Ready blocks until a Cluster is considered available. The current
+// implementation checks that the expected number of nodes are registered.
+func (c *Cluster) Ready() error {
+	f := func() error {
+		list, err := c.Client.CoreV1().Nodes().List(metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		if len(list.Items) != c.expectedNodes {
+			return fmt.Errorf("cluster is not ready, expected %v nodes got %v", c.expectedNodes, len(list.Items))
+		}
+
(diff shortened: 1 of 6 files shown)


e2e/framework/util.go
```diff
@@ -0,0 +1,20 @@
+package framework
+
+import "time"
+
+func Retry(attempts int, delay time.Duration, f func() error) error {
```
aaronlevy: There are some good helpers in apimachinery for these kinds of things: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L184

e2e/framework/setup.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"os"
+)
+
+var (
+	kubeconfig    = flag.String("kubeconfig", "../hack/quickstart/cluster/auth/kubeconfig", "absolute path to the kubeconfig file")
+	expectedNodes = flag.Int("nodes", 0, "the number of nodes to expect")
```
aaronlevy: maybe default to 1?

e2e/framework/cluster.go
```diff
@@ -0,0 +1,71 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// Cluster represents an interface to test bootkube based kubernetes clusters.
+// The simplest way to write tests against it is to just write go tests that call
+// out to NewCluster at the beginning of each test. External tooling should first
```
aaronlevy: Why can we not just create the cluster object once?

e2e/framework/setup.go
pbx0: It will hard-exit on 0. I figured if I default to a bad value it will make sure people use the flag correctly. If we change the default to something useful maybe it should be 2 because I think that will be the default for most tests. (which also means terraform-quickstart should have the same default as well)

e2e/framework/cluster.go
pbx0: It seems more uniform to make a call to the framework to get a client at every function then sharing globals in testing packages. No reason that NewCluster can't just be GetCluster that returns a global from the framework package assuming `Clientset` is safe for parallel use (docs didn't make this obvious if true). Its possible certain tests could be safe to run in parallel in the future.

pbx0:
@aaronlevy The main constructs I envision in addition to what we have is that the `getNodes(labelSelector)` function returns a `Node` type defined in the framework package. This type might have the following methods for example:
`func (n Node) Reboot() error`
`func (n Node) Destroy() error` (appropriately updates our `Ready()` function to expect less nodes)

Maybe one could do a reboot with a certain delay before starting the kubelet. Basically any functionality not normally provided directly by the k8s API could go in the framework package to be shared by all tests. I imagine this is mostly destructive in nature. 

Also any global state needed to be kept track of about the running cluster should go in the framework package and be exported through functions. Logging functionality might also be setup through the framework package.

Reflecting on this we really should probably only have 1 Cluster object per test run really instead of NewCluster at the start of every test.

e2e/framework/util.go
pbx0: I can switch to this. The Retry function here is just a copy of the pattern used all over the mantle (pluton/kola) code repo.

aaronlevy:
Lets start very simple with the actual tests we want to have -- and let that inform us of how/what kind of larger abstractions we need over time. The near-term goal should be functional tests (which we need badly) vs building out a more featured test framework (before we have tests).

For example, the first test we need is just a simple smoke test - can we start just with that and then as we add tests we can later look at building a framework for any painpoints / duplication?

pbx0:
I thought this was the minimal design for writing the smoke test we need. We can get a client and tell when the cluster is ready. We just need some package that coordinate that. The API can still be changed at will as needed.

Maybe if I include the smoke test in this PR it will make more sense what is and isn't needed yet?

PR #507, opened 2017-05-12
pkg/bootkube: increase count when api object already exists

xiang90:
If we do not increase this count, we will hit the error returns at line 132 in the same file.

Found when trying out recovery from a self hosted etcd cluster.

/cc @yifan-gu @aaronlevy


yifan-gu:
LGTM

xiang90:
@yifan-gu Do we need to wait (re-trigger) for e2e?

yifan-gu:
rktbot run tests

yifan-gu:
@xiang90 Doesn't have to, but I just kicked it off again.

xiang90:
@yifan-gu I will merge on green. thanks for the help.

Issue #508, opened 2017-05-12
recovery: do not try to infer etcd ip from service manifest 


xiang90:
We do not generate service manifest for recovery.

However, now bootkube infers etcd IP from service manifest here: https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/bootkube/parse.go#L14 during start.

We should either:

1. re-generate service manifest

1. get etcd IP from kuberentes API server

I tried both. Getting it from API server probably easier actually. Kubernetes itself has a bug when recreating the same service manifest. It does not return object already exists, but something else that is hard to parse.

/cc @aaronlevy @dghubble @hongchaodeng

PR #509, opened 2017-05-12
SelfHostedEtcd: get etcd service IP from APIServer

hongchaodeng:
Instead of parsing from static file, we change it to get etcd service IP from APIServer.

fix https://github.com/kubernetes-incubator/bootkube/issues/508


hongchaodeng:
@aaronlevy @xiang90

pkg/asset/k8s.go
```diff
@@ -12,6 +12,8 @@ import (
 )
 
 const (
+	EtcdServiceName = "etcd-service"
```
xiang90: comment on public const?

pkg/util/etcdutil/migrate.go
```diff
@@ -62,6 +64,11 @@ func Migrate(kubeConfig clientcmd.ClientConfig, etcdServiceIP string) error {
 	}
 	glog.Info("etcd cluster for migration is now running")
 
+	etcdServiceIP, err := getServiceIP(kubecli, api.NamespaceSystem, asset.EtcdServiceName)
```
xiang90: can we do this at line51? it is cleaner to get all prerequisites at one place before we start the migration.

xiang90:
looks good to me.

defer to other cc'd people. @dghubble is the author of this piece of code.

pkg/util/etcdutil/migrate.go
hongchaodeng: Isn't the code clearer to the variable closer to where it is used?

hongchaodeng:
fixed. PTAL

pkg/util/etcdutil/migrate.go
xiang90: if this is a simple variable declaration, then yes. but this is a func call that does some initializations.

hongchaodeng:
@aaronlevy @xiang90

dghubble:
Seems fine to me. Detecting the etcd IP from the service was just a means of getting away from the old hardcoded IP.

hongchaodeng:
fixed

xiang90:
LGTM

hongchaodeng:
rktbot run tests

ssh failure:
```
 run.go:147: creating cluster: populating cache: copying public keys: ssh: handshake failed: EOF: 
```

hongchaodeng:
rktbot run tests

hongchaodeng:
rktbot run tests

hongchaodeng:
@aaronlevy 
Can you approve this?

PR #511, opened 2017-05-15
start of e2e tests

pbx0:
cc @aaronlevy @yifan-gu

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,0 +1,16 @@
+## Bootkube E2E Testing
+
+This is the beginnings of E2E testing for the bootkube repo using the standard go testing harness. To run the tests once you have a kubeconfig to a running cluster just execute:
+`go test -v --kubeconfig=<filepath> ./e2e/`
+
+The number of nodes is required so that the setup can block on all nodes being registered.
+
+## Roadmap
+
+Implement whatever is needed to finish porting all functionality from pluton tests
+
+## Requirements
+
+Tests can't rely on network access to the cluster except via the kubernetes api. So no hitting nodes directly just because you can. This will maximize future portability with other setup tools.
+
+
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -1,0 +1,141 @@
+package e2e
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// global client for use by all tests
+var client kubernetes.Interface
+
+// non-configurable for now
+const namespace = "bootkube-e2e-testing"
+
+// TestMain handles setup before all tests
+func TestMain(m *testing.M) {
+	var (
+		kubeconfig = flag.String("kubeconfig", "../hack/quickstart/cluster/auth/kubeconfig", "absolute path to the kubeconfig file")
+	)
+	flag.Parse()
+
+	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	// create the clientset
+	client, err = kubernetes.NewForConfig(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	if err := ready(client); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	// createNamespace
+	if _, err := createNamespace(client, namespace); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	// run tests
+	exitCode := m.Run()
+
+	if err := deleteNamespace(client, namespace); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	os.Exit(exitCode)
+}
+
+func createNamespace(c kubernetes.Interface, name string) (*v1.Namespace, error) {
+	namespace, err := c.CoreV1().Namespaces().Create(&v1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create namespace with name %v %v", name, namespace)
+	}
+	return namespace, nil
+}
+
+func deleteNamespace(c kubernetes.Interface, name string) error {
+	return c.CoreV1().Namespaces().Delete(name, nil)
+}
+
+// Ready blocks until the cluster is considered available. The current
+// implementation checks that 1 schedulable node is ready.
+func ready(c kubernetes.Interface) error {
+	f := func() error {
+		list, err := c.CoreV1().Nodes().List(metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		if len(list.Items) < 1 {
+			return fmt.Errorf("cluster is not ready, waiting for 1 or more worker nodes: %v", len(list.Items))
+		}
+
+		// check for 1 or more ready nodes by ignoring nodes marked
+		// unschedulable or containing taints
+		var oneReady bool
+		for _, node := range list.Items {
+			if node.Spec.Unschedulable {
+				continue
+			}
+
+			if len(node.Spec.Taints) != 0 {
+				continue
+			}
+
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == v1.NodeReady {
+					if condition.Status == v1.ConditionTrue {
+						oneReady = true
+					}
+					break
+				}
+			}
+		}
+		if !oneReady {
+			return fmt.Errorf("waiting for one worker node to be ready")
+		}
+
+		return nil
+	}
+
+	if err := retry(50, 10*time.Second, f); err != nil {
+		return err
+	}
+	return nil
+}
+
+func retry(attempts int, delay time.Duration, f func() error) error {
+	var err error
+
+	for i := 0; i < attempts; i++ {
+		err = f()
+		if err == nil {
+			break
+		}
+
+		if i < attempts-1 {
+			time.Sleep(delay)
+		}
+	}
+
+	return err
+}
--- a/e2e/smoke_test.go
+++ b/e2e/smoke_test.go
@@ -1,0 +1,105 @@
+package e2e
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/pkg/api"
+	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
+)
+
+func TestSmoke(t *testing.T) {
+	// 1. create nginx deployment
+	di, _, err := api.Codecs.UniversalDecoder().Decode(nginxDep, nil, &v1beta1.Deployment{})
+	if err != nil {
+		t.Fatalf("unable to decode deployment manifest: %v", err)
+	}
+	d, ok := di.(*v1beta1.Deployment)
+	if !ok {
+		t.Fatalf("expected manifest to decode into *api.Service, got %T", di)
+	}
+	_, err = client.ExtensionsV1beta1().Deployments(namespace).Create(d)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer client.ExtensionsV1beta1().Deployments(namespace).Delete("nginx-deployment", &metav1.DeleteOptions{})
+
+	// 2. Get the nginx pod IP.
+	var p *v1.Pod
+	getPod := func() error {
+		l, err := client.CoreV1().Pods(namespace).List(metav1.ListOptions{LabelSelector: "app=nginx"})
+		if err != nil || len(l.Items) == 0 {
+			return fmt.Errorf("pod not yet running: %v", err)
+		}
+
+		// take the first pod
+		p = &l.Items[0]
+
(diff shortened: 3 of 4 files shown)


e2e/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+The number of nodes is required so that the setup can block on all nodes being registered.
+
+## Roadmap
+
+Implement whatever is needed to finish porting all functionality from pluton tests
+
+## Requirements
+
+Tests can't rely on network access to the cluster except via the kubernetes api. So no hitting nodes directly just because you can. This will maximize future portability with other setup tools.
```
yifan-gu: That's not true anymore right? We need ssh or an agent running on the node to do the reboot or isolation.

pbx0: I think for now we can do reboot via privileged pod/daemon-set.  Maybe this will change in future or we can find universal way to proxy out SSH access or create a more robust destructive agent.

e2e/main_test.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, fmt.Errorf("failed to create namespace with name %v %v", name, namespace)
+	}
+	return namespace, nil
+}
+
+func deleteNamespace(c kubernetes.Interface, name string) error {
+	return c.CoreV1().Namespaces().Delete(name, nil)
+}
+
+// Ready blocks until a the cluster is considered available. The current
```
yifan-gu: remove 'a'

e2e/main_test.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	}
+
+	if err := retry(12, 10*time.Second, f); err != nil {
+		return err
+	}
+	return nil
+}
+
+func retry(attempts int, delay time.Duration, f func() error) error {
```
yifan-gu: can use `"k8s.io/apimachinery/pkg/util/wait.Poll()` instead.

pbx0: Poll, doesn't bubble up errors in the retry loop unless you marked the error as fatal. This way I know what error message failed to resolve. I havn't had use for the distinction between a fatal and non-fatal error in a retry loop so far, but this function is well tested (used everywhere in mantle) and doesn't preclude anyone from using poll since we don't have a framework api yet.

e2e/main_test.go
```diff
@@ -0,0 +1,134 @@
...
+// Ready blocks until a the cluster is considered available. The current
+// implementation checks that the expected number of nodes are registered.
+func ready(c kubernetes.Interface, expectedNodes int) error {
+	f := func() error {
+		list, err := c.CoreV1().Nodes().List(metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		if len(list.Items) != expectedNodes {
```
yifan-gu: Why do we need to care the number of nodes?

pbx0: Allows us to block testing until all the nodes have registered. This way we don't begin any tests until the cluster is at least somewhat in a ready state which could cause flakes. More robust checks could be added as well.

Some nodes can take an extra minute or so to be marked ready after the first node bootstraps.

yifan-gu: @pbx0 So is this a "1 node or all nodes" situation? In that case I still don't see we need to know the exact number?

e2e/main_test.go
pbx0: done

e2e/main_test.go
yifan-gu: It's the same thing here right? The `err` returned by `f()` is ignored.

In `Poll()` you can achieve the same by ignoring the errors, then the only error can happen will be the timout.

e2e/README.md
yifan-gu: @pbx0 As SSH is available today, maybe we go that route first, and after the tests are moved to here, we can work on the framework and the agent, wdyt?

pbx0: Whatever is easiest I suppose. It will take a little work to setup the ability to SSH to each machine but it'd certainly make porting over the tests a bit easier.  

@aaronlevy What are your thoughts on the direction to take immediate destructive needs?

e2e/main_test.go
pbx0: The `retry` function here will just keep retrying the function until it returns nil. If it runs out of retries it will return the last error returned from the function. So if you were waiting on multiple conditions you'll know which one was the blocking one by writing different errors within the function that is retrying.

e2e/main_test.go
pbx0: If we just check for the presence of ready nodes then we won't know if all nodes are online. While waiting for nodes to come up they have no state at all in the api since the apiserver hasn't seen them check in at all yet. 

For example, if I setup terraform-quickstart to run 3 masters and 1 worker and I run `kubectl get nodes` right after `start-cluster.sh` succeeds. I will see that 1 node is ready and no other nodes exist. So I have to know how many nodes to expect. If a node doesn't check-in the cluster isn't in the state you expected and may not have enough nodes for certain tests.

e2e/main_test.go
yifan-gu: @pbx0 Ok, fair.

e2e/README.md
yifan-gu: @pbx0 I will go implement some SSH facilities.

e2e/main_test.go
yifan-gu: Good point.

e2e/smoke_test.go
```diff
@@ -0,0 +1,99 @@
...
+				},
+			},
+		},
+	},
+}
+
+var curlPod = &v1.Pod{
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "curl-pod",
+		Namespace: namespace,
```
yifan-gu: why this is in the `bootkubetesting` namespace, while the nginx pod is in the `default` namespace?

yifan-gu:
Other than one question, LGTM :+1:

e2e/main_test.go
aaronlevy: This still feels unnecessary to me. I feel like we should block the tests running on the apiserver being ready -- not on having to know how many nodes are going to be in the cluster. That should be left as a per-test requirement (e.g. fail etcd-scale test if masters < 3).

e2e/main_test.go
aaronlevy: I would think that we're logging every failed attempt (which would make it obvious what was failing when it times out).

aaronlevy: "Silent until failure" can actually make debugging more difficult

e2e/smoke_test.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		return nil
+	}
+	if err := retry(10, time.Second*10, getPod); err != nil {
+		t.Fatalf(fmt.Sprintf("timed out waiting for curl pod to succeed: %v", err))
+	}
+}
+
+var two int32 = 2
+var nginxDeployment = &v1beta1.Deployment{
```
aaronlevy: Won't block on this -- but to make adding assets for tests easier, it would be nice to define these as yaml then marshaling them into the object types when needed

aaronlevy: Can we add a TODO for this

e2e/smoke_test.go
```diff
@@ -0,0 +1,99 @@
...
+var curlPod = &v1.Pod{
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "curl-pod",
+		Namespace: namespace,
+	},
+	Spec: v1.PodSpec{
+		Containers: []v1.Container{
+			{
+				Name:  "curl-container",
+				Image: "byrnedo/alpine-curl", // TODO(yifan): Host this image under quay.io/coreos/
```
aaronlevy: I'd prefer that we don't use a random image (especially untagged). Short term you can use busybox and it at least has wget

e2e/main_test.go
pbx0: This can strangely effect the timeouts needed for tests that don't have a explicit dependency on number of nodes. For example with the nginx test running against 1 master and 1 worker nginx would need the worker available to schedule the pod because of the NoSchedule taint. This can sometimes take a few minutes after terraform finishes. Without the ready check the nginx timeout has to wait all the extra minutes in case node bring-up is really slow. With the ready check you can reliably have a much smaller timeout.

So If we remove this, then even on tests that don't have specific node requirements have to have timeouts set very high.

Also it serves as a sanity check that our terraform tool is doing the right thing and makes errors related to terraform show up before the tests start so the errors have better context.

e2e/smoke_test.go
pbx0: Will update, I meant for everything like this to use the testing namespace and not the default.

e2e/smoke_test.go
pbx0: I can just change this quickly. I originally landed here because I kept getting strange errors trying to decode the yaml into the right type but yifan showed me which function to use in client-go.

e2e/main_test.go
pbx0: Would recording each failed attempt and then logging it all at once if it fails work?

If we log each expected and fleeting retry failure the test output will be very noisy even on success.

aaronlevy: We would just be logging it as it happens, no need to record. I won't block on this, but it would simply be:

```
wait.Poll(interval, timeout, func() (bool, error) {
    ready, err := nodesReady(4)
    if err != nil {
        glog.Error(err)
        return false, nil
    }
    return ready, nil
})
```

aaronlevy: My contention is that when aggregating things going wrong in the test-- output every x-seconds is more helpful than no output then failure.

e2e/main_test.go
aaronlevy: I won't block on this and we can revisit -- but requiring the init to know how many nodes there will be feels a bit unnecessary. The timing issue could be resolved simply by waiting for a single node to always be ready. Then the tests themselves enforce checks on a minimum number of nodes (this could change over lifetime of cluster-- simply checking at beginning isn't always that meaningful).

But lets just move on and this can be revisited later.

e2e/main_test.go
```diff
@@ -0,0 +1,134 @@
...
+var client kubernetes.Interface
+
+// non-configurable for now
+const namespace = "bootkubetesting"
+
+// TestMain handles setup before all tests
+func TestMain(m *testing.M) {
+	var (
+		kubeconfig    = flag.String("kubeconfig", "../hack/quickstart/cluster/auth/kubeconfig", "absolute path to the kubeconfig file")
+		expectedNodes = flag.Int("nodes", 0, "the number of nodes to expect")
```
aaronlevy: Zero is never correct - why do we not just default to 1?

pbx0: I figured using default values here is always incorrect. Expected nodes should be explicit. So there is no safe default. Instead, set it to an obviously incorrect sentinel that will cause an error message before tests start.

Also setting it to 1 and not having 1 node would cause flakeyness (if two nodes check in before 1 is ready).  We could have 0 disable the check altogether, but at that point if we support that we should probably remove the check altogether adjust timeouts among tests appropriately.

aaronlevy: Hmm why would having 2 nodes make this fail? Do tests not run if there are 2 nodes in the cluster? That doesn't sound like a restriction we should adhere to.

e2e/smoke_test.go
yifan-gu: Sounds good.
```
image: busybox
command: wget $POD_IP
```
Can do the same for us.

pbx0: done

pbx0:
After some discussion we are removing the expected number of node flag for now. Might add it back later for specific test use only. Instead the `ready()` function will check that at least one worker node is considered ready before moving on to running the tests. To accomplish this I just look for nodes that are not marked unschedulable and have no taints.

Also, I changed to using busybox + wget and made the nginx deployment declared in yaml. The wget pod is still a go struct because its easier to modify dynamically with the right pod IP to hit.

e2e/smoke_test.go
pbx0: Everything should use the testing namespace. I removed this field from the pod since the namespace gets set in the API call anyway.

e2e/smoke_test.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer client.CoreV1().Pods(namespace).Delete(wgetPod.ObjectMeta.Name, &metav1.DeleteOptions{})
+
+	getPod = func() error {
+		p, err = client.CoreV1().Pods(namespace).Get("wget-pod", metav1.GetOptions{})
+		if err == nil && p.Status.Phase == v1.PodSucceeded {
+			t.Logf("waiting on wget-pod: %v", err)
+			return fmt.Errorf("pod not yet running: %v", err)
```
yifan-gu: @pbx0 Why returns an error here when it succeeded?

e2e/smoke_test.go
```diff
@@ -0,0 +1,97 @@
...
+var wgetPod = &v1.Pod{
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "wget-pod",
+		Namespace: namespace,
+	},
+	Spec: v1.PodSpec{
+		Containers: []v1.Container{
+			{
+				Name:  "wget-container",
+				Image: "busybox",
```
yifan-gu: maybe add a version, e.g. `busybox:1.26`

e2e/main_test.go
yifan-gu: Can add a TODO here to revisit.

e2e/smoke_test.go
pbx0: Oh, good catch. I completely got this logic reversed when I copy-pasted your code. I think I my mind saw an if statement and was tunnel vision about it being an error path. Fixing.

pbx0:
@yifan-gu resolved your comments PTAL

yifan-gu:
LGTM for me, any objections @aaronlevy

aaronlevy:
Feel free to merge - don't block on me

PR #512, opened 2017-05-15
templates: etcd operator should select master nodes

hongchaodeng:
All system components are sitting on master nodes.
However, etcd operator was somehow dismissed for that.

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -875,6 +875,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
       tolerations:
       - key: node-role.kubernetes.io/master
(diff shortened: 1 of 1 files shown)


xiang90:
probably we should add a test for this one.

@aaronlevy where should we put the test?

xiang90:
This causes us issue in the tectonic setup where only master nodes can talk to the on host etcd due to our security group setting.

aaronlevy:
RE tests: @xiang90 @hongchaodeng can you coordinate with @pbx0 -- he's started to move the tests into the bootkube repo (see: https://github.com/kubernetes-incubator/bootkube/pull/511 )

sym3tri:
@xiang90 in the future can you please make these changes to https://github.com/coreos/tectonic-installer/blob/master/modules/bootkube/resources/experimental/manifests/etcd-operator.yaml also?

PR #515, opened 2017-05-15
support bringing up more masters in terraform and quickstart scripts

pbx0:
This is a requirement for the etcd-scaling test to be run against a terraform-quickstart created cluster.

cc @aaronlevy @yifan-gu

### DIFF
--- a/hack/terraform-quickstart/main.tf
+++ b/hack/terraform-quickstart/main.tf
@@ -25,6 +25,17 @@ resource "aws_instance" "worker_node" {
   }
 }
 
+resource "aws_instance" "master_node" {
+  ami           = "${data.aws_ami.coreos_ami.image_id}"
+  instance_type = "m3.medium"
+  key_name      = "${var.ssh_key}"
+  count         = "${var.additional_masters}"
+
+  tags {
+    Name = "${var.instance_tags}"
+  }
+}
+
 data "aws_ami" "coreos_ami" {
   most_recent = true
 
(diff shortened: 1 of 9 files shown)


yifan-gu:
@pbx0 How can I test this?

pbx0:
@yifan-gu Create a file called terraform.tfvars with the following info:
```
access_key_id = "<YOUR_KEY_HERE>"
access_key = "<YOUR_KEY_HERE>"
instance_tags = "yifan_test_terraform"
ssh_key = "<name of your ssh key on amazon>"
additional_masters = 2
```

Then run `terraform apply`
Then run `./start_cluster.sh`

Now you should have a cluster running with 3 masters and 1 worker (the default). Check out the variables files to see all the inputs to terraform. You override any of those defaults via the terraform.tfvars file.

yifan-gu:
@pbx0 Thanks, let me try it.

hack/terraform-quickstart/main.tf
```diff
@@ -25,6 +25,17 @@ resource "aws_instance" "worker_node" {
   }
 }
 
+resource "aws_instance" "master_node" {
+  ami           = "${data.aws_ami.coreos_ami.image_id}"
+  instance_type = "m3.medium"
+  key_name      = "${var.ssh_key}"
+  count         = "${var.additional_masters}"
```
aaronlevy: Actually "additional masters" is a bit vague. Can we just use a specific number: "master_count" or some such?

pbx0: We could be more symmetrical and use `num_masters` and `num_workers` it'd just take adding a little logic to terraform to figure out that one of the `num_masters` is actually the bootstrap node and is special. So I think we'd want to validate that this number is 1 or greater or something like that in terraform.

I labeled it as `addtional_masters` to imply it doesn't include the bootstrap node because it was easier to write that way.

aaronlevy: ah I see. Can we add a comment / todo to this effect? (assuming you can have comments in tf files)

pbx0: I added this to the variable description in the variables.tf file.

pbx0:
Renamed `init-bootstrap.sh -> back to init-master.sh` so that tests will pass.

PR #516, opened 2017-05-17
e2e: Add a simple ssh client and node type

yifan-gu:
The client has the same interface as the one in pluton framework,
makes it easier for us to migrate tests.


yifan-gu:
cc @aaronlevy @diegs @pbx0

pbx0:
@yifan-gu Since this is included directly in our test package and will be called directly from tests, probably the tests need not know which private key to use and would rather just pass a node object to build the SSH client. Right now, global test state is all handled through globals so we might as well continue the trend and directly source private key info from a global set up by a flag.

In the future, if we factor more things out into a framework package/object, that might handle client creation and would have knowledge of the private keys and such from flag setup.

yifan-gu:
@pbx0 Let me modify the SSH() a little bit, make it to take a `host` as the parameter, how does that sound?

yifan-gu:
Seems vendor is broken again. Trying to fix that first.

yifan-gu:
@rktbot run tests

yifan-gu:
rktbot run tests

yifan-gu:
rktbot run tests

e2e/main_test.go
```diff
@@ -139,3 +139,17 @@ func retry(attempts int, delay time.Duration, f func() error) error {
 
 	return err
 }
+
+func RebootMachine(c *SSHClient, host string) error {
+	// stop sshd so that commonMachineChecks will only work if the machine
```
pbx0: This comment isn't relevant outside the kola framework

e2e/node.go
```diff
@@ -0,0 +1,31 @@
+package e2e
+
+import (
+	"fmt"
+
+	"golang.org/x/crypto/ssh"
+)
+
+type Node struct {
```
pbx0: Maybe consider receiving a `*v1.Node` directly instead of a host?

e2e/node.go
```diff
@@ -0,0 +1,31 @@
...
+	host string
+}
+
+func NewNode(host string) *Node {
```
pbx0: Maybe consider receiving a `*v1.Node` directly instead of a host?

PR #523, opened 2017-05-18
asset: make it easier to load and write asset out of rendering phase

xiang90:
/cc @diegs 

This makes it easier to load in recovery etcd template and output recovery etcd manifest as a static pod.

### DIFF
--- a/pkg/asset/asset.go
+++ b/pkg/asset/asset.go
@@ -185,14 +185,18 @@ func (as Assets) WriteFiles(path string) error {
 		return err
 	}
 	for _, asset := range as {
-		f := filepath.Join(path, asset.Name)
-		if err := os.MkdirAll(filepath.Dir(f), 0755); err != nil {
-			return err
-		}
-		fmt.Printf("Writing asset: %s\n", f)
-		if err := ioutil.WriteFile(f, asset.Data, 0600); err != nil {
+		if err := asset.WriteFile(path); err != nil {
 			return err
 		}
 	}
 	return nil
+}
+
+func (a Asset) WriteFile(path string) error {
+	f := filepath.Join(path, a.Name)
+	if err := os.MkdirAll(filepath.Dir(f), 0755); err != nil {
+		return err
+	}
+	fmt.Printf("Writing asset: %s\n", f)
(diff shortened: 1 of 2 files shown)


pkg/asset/asset.go
```diff
@@ -185,14 +185,21 @@ func (as Assets) WriteFiles(path string) error {
...
 	return nil
 }
+
+func (a Asset) WriteFile(path string) error {
+	f := filepath.Join(path, a.Name)
+	if err := os.MkdirAll(filepath.Dir(f), 0755); err != nil {
+		return err
+	}
+	fmt.Printf("Writing asset: %s\n", f)
+	if err := ioutil.WriteFile(f, a.Data, 0600); err != nil {
```
diegs: nit: return `ioutil.WriteFile(...)`?

pkg/asset/k8s.go
```diff
@@ -163,7 +163,7 @@ func secretFromAssets(name, namespace string, assetNames []string, assets Assets
 	})
 }
 
-func mustCreateAssetFromTemplate(name string, template []byte, data interface{}) Asset {
+func MustCreateAssetFromTemplate(name string, template []byte, data interface{}) Asset {
```
diegs: Can you elaborate on why you need to make this a public API?

Currently in bootkube recovery tool we simply return an Assets array, and then the caller invokes `Assets.WriteFiles()` to write them out to disk.

xiang90: Our templates sit inside an internal pkg. There is no easy way to ONLY load in one recovery etcd template. Alternatively, I can put a special loading logic inside Asset pkg (WriteOutRecoveryEtcd or something). But I feel it is less idea.

Also I actually want to put the recovery etcd template inside the recovery pkg.

xiang90: > bootkube recovery tool 

One difference is that we load the actual manifest bytes from etcd directly. We do not try to generate manifest bytes from a template. So the current recovery tool does not need to public `MustCreateAssetFromTemplate` func.

diegs: Ok, SGTM

xiang90:
@diegs Thanks for the quick review. The nit is fixed. I will get it merged on green.

Issue #513, opened 2017-05-15
Add Calico network policy support


aaronlevy:
This should be "policy-only" support (leaving flannel as the pod-network provider).

See: Calico policy-only documentation here: 

http://docs.projectcalico.org/v2.2/getting-started/kubernetes/installation/hosted/kubernetes-datastore/

@abhinavdahiya is going to pick this up. I'll assign once he's part of the org.

/cc @joshrosso @yifan-gu

rmenn:
its really funny since i worked with @abhinavdahiya, i have a working config for canal, there is something which works with canal but a `ExecStartPre=chmod 777 /etc/cni/net.d/`  which enables canal to write a file. 
 This line causes a problem and the chmod does not take effect 
https://github.com/projectcalico/cni-plugin/blob/master/k8s-install/scripts/install-cni.sh#L100

If needed i can get the logs from the server and post here, also this was possible on a previous version as i did have canal running on the server
Here is a snippet 
```
kube-system   canal-prbgf                               3/3       Running   0          22m       192.168.64.178   192.168.64.178
kube-system   canal-vkq8f                               3/3       Running   0          22m       192.168.64.179   192.168.64.179
```

yifan-gu:
I believe this is now implemented by #524

PR #524, opened 2017-05-18
Calico policy support

abhinavdahiya:
#513 
Notes:
- Uses Calico's Kubernetes Datastore  [`installation`](http://docs.projectcalico.org/v2.2/getting-started/kubernetes/installation/hosted/kubernetes-datastore/) to provide policy only support

Needs to be discussed:
hyperkube image already ships with cni binaries in `/opt/cni/bin` and kube-calico DaemonSet also installs its cni binaries in `/opt/cni/bin`.

So there would be chances of version mismatch as kubelet doesn't mount this directory from host. To prevent this I have mounted an empty `/opt/cni/bin` from the host to the kubelet.

### DIFF
--- a/e2e/network_test.go
+++ b/e2e/network_test.go
@@ -70,40 +70,47 @@ func TestNetwork(t *testing.T) {
 	}
 
 	t.Run("DefaultDeny", HelperDefaultDeny)
-
-	resetNetworkPolicy := func() {
-		n, err := client.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
-		if err != nil {
-			t.Fatalf("unable to retrieve namespace %v", err)
-		}
-
-		n.ObjectMeta.Annotations = map[string]string{}
-		n.ObjectMeta.Annotations["net.beta.kubernetes.io/network-policy"] = defaultAllowNetworkPolicy
-		_, err = client.CoreV1().Namespaces().Update(n)
-		if err != nil {
-			t.Fatalf("unable to reset namespace network policy%v", err)
-		}
-	}
-	defer resetNetworkPolicy()
-
 	t.Run("NetworkPolicy", HelperPolicy)
 }
 
 func HelperDefaultDeny(t *testing.T) {
 	//
 	// 3. set DefaultDeny policy
-	var n *v1.Namespace
-	n, err := client.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
+	npi, _, err := api.Codecs.UniversalDecoder().Decode(defaultDenyNetworkPolicy, nil, &v1beta1.NetworkPolicy{})
 	if err != nil {
-		t.Fatalf("unable to retrieve namespace %v", err)
+		t.Fatalf("unable to decode network policy manifest: %v", err)
 	}
 
-	n.ObjectMeta.Annotations = map[string]string{}
-	n.ObjectMeta.Annotations["net.beta.kubernetes.io/network-policy"] = defaultDenyNetworkPolicy
-	_, err = client.CoreV1().Namespaces().Update(n)
-	if err != nil {
-		t.Fatalf("unable to set namespace network policy defaultdeny%v", err)
+	np, ok := npi.(*v1beta1.NetworkPolicy)
+	if !ok {
+		t.Fatalf("expected manifest to decode into *api.networkpolicy, got %T", npi)
 	}
+
+	httpRestClient := client.ExtensionsV1beta1().RESTClient()
+	uri := fmt.Sprintf("/apis/%s/%s/namespaces/%s/%s",
+		strings.ToLower("extensions"),
+		strings.ToLower("v1beta1"),
+		strings.ToLower(namespace),
+		strings.ToLower("NetworkPolicies"))
+
+	result := httpRestClient.Post().RequestURI(uri).Body(np).Do()
+	if result.Error() != nil {
+		t.Fatal(result.Error())
+	}
+	defer func() {
+		uri = fmt.Sprintf("/apis/%s/%s/namespaces/%s/%s/%s",
+			strings.ToLower("extensions"),
+			strings.ToLower("v1beta1"),
+			strings.ToLower(namespace),
+			strings.ToLower("NetworkPolicies"),
+			strings.ToLower(np.ObjectMeta.Name))
+
+		result = httpRestClient.Delete().RequestURI(uri).Do()
+		if result.Error() != nil {
+			t.Fatal(result.Error())
+		}
+
+	}()
 
 	//
 	// 4. create a wget pod that fails to hit nginx service
@@ -276,8 +283,13 @@ spec:
       targetPort: 80
 `)
 
-var defaultDenyNetworkPolicy = `{"ingress":{"isolation":"DefaultDeny"}}`
-var defaultAllowNetworkPolicy = `{"ingress":{"isolation":""}}`
+var defaultDenyNetworkPolicy = []byte(`kind: NetworkPolicy                                                      
+apiVersion: extensions/v1beta1
+metadata:
+  name: default-deny
+spec:
+  podSelector:
+`)
 
 var netPolicy = []byte(`kind: NetworkPolicy                                                      
 apiVersion: extensions/v1beta1
--- a/e2e/network_test.go
+++ b/e2e/network_test.go
@@ -1,0 +1,295 @@
+package e2e
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/client-go/pkg/api"
+	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
+)
+
+func TestNetwork(t *testing.T) {
+	//
+	// 1. create nginx service
+	di, _, err := api.Codecs.UniversalDecoder().Decode(nginxDepNT, nil, &v1beta1.Deployment{})
+	if err != nil {
+		t.Fatalf("unable to decode deployment manifest: %v", err)
+	}
+
+	d, ok := di.(*v1beta1.Deployment)
+	if !ok {
+		t.Fatalf("expected manifest to decode into *api.deployment, got %T", di)
+	}
+	_, err = client.ExtensionsV1beta1().Deployments(namespace).Create(d)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	deleteDeployment := func() {
+		delPropPolicy := metav1.DeletePropagationForeground
+		client.ExtensionsV1beta1().Deployments(namespace).Delete("nginx-deployment-nt", &metav1.DeleteOptions{
+			PropagationPolicy: &delPropPolicy,
+		})
+	}
+	defer deleteDeployment()
+
+	if err := retry(10, time.Second*10, getNginxPod); err != nil {
+		t.Fatalf("timed out waiting for nginx pod: %v", err)
+	}
+
+	si, _, err := api.Codecs.UniversalDecoder().Decode(nginxSVCNT, nil, &v1.Service{})
+	if err != nil {
+		t.Fatalf("unable to decode service manifest: %v", err)
+	}
+	s, ok := si.(*v1.Service)
+	if !ok {
+		t.Fatalf("expected manifest to decode into *api.service, got %T", si)
+	}
+	_, err = client.CoreV1().Services(namespace).Create(s)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer client.CoreV1().Services(namespace).Delete("nginx-service-nt", &metav1.DeleteOptions{})
+
+	//
+	// 2. create a wget pod that hits the nginx service
+	testPodName := fmt.Sprintf("%s-%s", "wget-pod-nt", utilrand.String(5))
+	wgetPodNT.ObjectMeta.Name = testPodName
+	_, err = client.CoreV1().Pods(namespace).Create(wgetPodNT)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := retry(10, time.Second*10, getPod(testPodName)); err != nil {
+		t.Fatalf(fmt.Sprintf("timed out waiting for wget pod to succeed: %v", err))
+	}
+
+	t.Run("DefaultDeny", HelperDefaultDeny)
+
+	resetNetworkPolicy := func() {
+		n, err := client.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
+		if err != nil {
+			t.Fatalf("unable to retrieve namespace %v", err)
+		}
+
+		n.ObjectMeta.Annotations = map[string]string{}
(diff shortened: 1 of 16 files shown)


yifan-gu:
rktbot run tests

pkg/asset/internal/templates.go
```diff
@@ -1071,4 +1071,224 @@ spec:
...
+          image: {{ .Images.CalicoCNI }}
+          command: ["/install-cni.sh"]
+          env:
+            - name: CNI_NETWORK_CONFIG
+              valueFrom:
+                configMapKeyRef:
+                  name: kube-calico-cfg
+                  key: cni_network_config
+            - name: CNI_NET_DIR
+              value: "/etc/kubernetes/cni/net.d"
```
yifan-gu: Isn't the mount path `/host/etc/cni/net.d` ?

yifan-gu:
> So there would be chances of version mismatch as kubelet doesn't mount this directory from host. To prevent this I have mounted an empty /opt/cni/bin from the host to the kubelet.

One thought on this is to have flannel pod in the daemonset install the flannel and cni binaries to the directory, so kubelet always mounts the directory regardless calico network policy is enabled or not.
However then we need to deal with the race between calico pods vs flannel pods on who installed the binaries.
To solve that we can have calico pods to only install the calico binary, not the flannel or cni. The risk here is we need to make sure calico version X works with flannel version Y and cni version Z.

cc @aaronlevy @philips

yifan-gu:
@abhinavdahiya Next step is adding a bootkube test to exercising the network policy just like the demo does. @pbx0 and I am working on getting some basic facilities working so you can use to write tests soon.

philips:
Yes, I 100% agree we should move to this model where flannel puts down the
CNI plugin and we don't use the one in the kubelet image. This will help us
remove calico from the image too.

abhinavdahiya:
@yifan-gu I added tests for network policy yesterday. Take a look.

e2e/network_test.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := retry(10, time.Second*10, getPod); err != nil {
+		t.Fatalf(fmt.Sprintf("timed out waiting for wget pod to succeed: %v", err))
+	}
+	client.CoreV1().Pods(namespace).Delete(wgetPodNT.ObjectMeta.Name, &metav1.DeleteOptions{})
+
+	fmt.Println("Starting default deny test")
```
pbx0: These two Println statements seems like separate tests within a single test. Might be worth making use of go's subtests here. 

Regardless, its better form to log via the test logger then just printing to stdout directly.

e2e/main_test.go
```diff
@@ -19,11 +19,13 @@ var client kubernetes.Interface
 // non-configurable for now
 const namespace = "bootkube-e2e-testing"
 
+var (
+	kubeconfig = flag.String("kubeconfig", "../hack/quickstart/cluster/auth/kubeconfig", "absolute path to the kubeconfig file")
```
pbx0: nit: If you need to export kubeconfig as a global. Might be best to still initialize this in TestMain via `flag.StringVar(&kubeconfig...` Just keeps all our init functionality in once place happening in deterministic order. I think either way it will be correct though.

pbx0:
rktbot run e2e

pbx0:
Needs a rebase before new e2e tests will run properly. The "e2e-selfetcd" will actually run the test you wrote in this PR.

pbx0:
rktbot run tests

pbx0:
rktbot run e2e

abhinavdahiya:
rktbot run e2e

pbx0:
rktbot run e2e

pbx0:
The new e2e tests have been consistantly passing on master, but on every run against this PR (3 times now), it is the smoke test that fails. Since we have a shared cluster for tests, is it possible that the tests here are not leaving the cluster in the same state causing the smoke test to fail?

abhinavdahiya:
@pbx0 Yup. Network test was not deleting the network policy after completing and therefore testsmoke failed. Should be fixed now

pbx0:
@abhinavdahiya tests look good to me. Thanks for debugging that.

Just noting in general that in the future test framework work, we likely want separate namespaces for each test. Ensuring everything is cleaned up can be challenging and error prone and will affect other tests such as in this case.

yifan-gu:
@abhinavdahiya Didn't see big issues in the PR, but can we squash the commits also why there are "Merge" commits?

abhinavdahiya:
@yifan-gu squashed the commits and fixed the merge conflicts

e2e/network_test.go
```diff
@@ -0,0 +1,310 @@
...
+		n.ObjectMeta.Annotations["net.beta.kubernetes.io/network-policy"] = defaultAllowNetworkPolicy
+		_, err = client.CoreV1().Namespaces().Update(n)
+		if err != nil {
+			t.Fatalf("unable to reset namespace network policy%v", err)
+		}
+	}
+	defer resetNetworkPolicy()
+
+	t.Run("NetworkPolicy", HelperPolicy)
+
```
yifan-gu: nit: empty line

e2e/network_test.go
```diff
@@ -0,0 +1,310 @@
...
+        app: nginx
+    spec:
+      containers:
+      - name: nginx
+        image: nginx:1.8
+        ports:
+        - containerPort: 80
+`)
+
+var wgetPodNT = &v1.Pod{
```
yifan-gu: nit: why this pod is a struct while others are bytes? Might want to be consistent.

e2e/network_test.go
```diff
@@ -0,0 +1,310 @@
...
+		return fmt.Errorf("pod not yet running: %v", p.Status.Phase)
+	}
+	return nil
+}
+
+func getPod(name string) func() error {
+	return func() error {
+		p, err := client.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("pod not yet running: %v", err)
```
yifan-gu: error doesn't necessary means the pod is not running, right?

e2e/network_test.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func getFailedPod(name string) func() error {
+	return func() error {
+		p, err := client.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("pod not yet running: %v", err)
+		}
+		if p.Status.Phase != v1.PodFailed {
+			return fmt.Errorf("pod not yet running: %v", p.Status.Phase)
```
yifan-gu: "pod not failed"

yifan-gu:
@abhinavdahiya Couple nits otherwise lgtm. 
One more question: After the PR, the calico daemonsets are deployed in bootkube by default right?

e2e/network_test.go
abhinavdahiya: wgetPodNT is used to start ping pods with different names so it is easier to change it from struct. While others are used as manifest files!

yifan-gu:
The new commit LGTM.

e2e/network_test.go
yifan-gu: @abhinavdahiya ok, but why the pod needs to have different name every time? BTW I didn't see the pod gets removed after tests? I know `kubectl delete namespace foo` will clean up all resources within the namespace as it implies `--cascades=true`, but not sure if `CoreV1().Namespaces().Delete(name, nil)`(which is called after each test) will do the same.

- If `CoreV1().Namespaces().Delete(name, nil)` will clean up all resources in the namespace, then a lot of the `defer`s above are not necessary.
- If not, then we need to clean up the pods.

yifan-gu: @abhinavdahiya Alright, I manually tested, `CoreV1().Namespaces().Delete(name, nil)` cleans up all the resources within the namespace. So you don't have to bother with those `defer`s.

abhinavdahiya:
rktbot run e2e

abhinavdahiya:
rktbot run e2e

abhinavdahiya:
@yifan-gu 
* flannel daemonset install cni binaries
* the cni binaries are now installed at `/opt/cni/bin` as we discussed

yifan-gu:
@abhinavdahiya New commit LGTM. Maybe squash some of the commits?

My only concern is the race between calico's install_cni.sh and the flannel pod's install-cni image, which can cause the cni binary not to have a consistent version. Maybe add a TODO somewhere and link the issue https://github.com/projectcalico/cni-plugin/issues/339

abhinavdahiya:
rktbot run e2e

abhinavdahiya:
rktbot run e2e

yifan-gu:
rktbot run e2e

yifan-gu:
rktbot run tests

yifan-gu:
I believe this depends on https://github.com/projectcalico/cni-plugin/pull/341 
Correct me if wrong @abhinavdahiya

pkg/asset/internal/templates.go
```diff
@@ -1079,13 +1079,38 @@ spec:
         - name: flannel-cfg
           mountPath: /etc/kube-flannel/
       - name: install-cni
-        image: {{ .Images.Busybox }}
-        command: [ "/bin/sh", "-c", "set -e -x; TMP=/etc/cni/net.d/.tmp-flannel-cfg; cp /etc/kube-flannel/cni-conf.json ${TMP}; mv ${TMP} /etc/cni/net.d/10-flannel.conf; while :; do sleep 3600; done" ]
+        image: {{ .Images.Alpine }}
+        command:
```
aaronlevy: This scares me. I don't think we should be wget-ing the cni release at runtime. My initial reaction would be that we need a versioned container image that has all the files we need, and we're simply copying them to a different mount.

rimusz: @aaronlevy is right

yifan-gu:
projectcalico/cni-plugin#341 is merged. Now this PR is waiting for a new calico release.

pkg/asset/internal/templates.go
yifan-gu: @abhinavdahiya Created a repo for the image here: https://github.com/coreos/flannel-cni 

/cc @aaronlevy @rimusz

pkg/asset/internal/templates.go
```diff
@@ -1113,10 +1119,246 @@ spec:
...
+                  fieldPath: spec.nodeName
+            - name: SKIP_CNI_BINARIES
+              value: bridge,cnitool,dhcp,flannel,host-local,ipvlan,loopback,macvlan,noop,portmap,ptp,tuning
+          volumeMounts:
+            - mountPath: /host/opt/cni/bin
+              name: cni-bin-dir
+            - mountPath: /host/etc/cni/net.d
+              name: cni-net-dir
+      volumes:
+        - name: lib-modules
```
aaronlevy: less than ideal, and I don't think this is required (and it's not /lib/modules from the host anyway).

abhinavdahiya: @aaronlevy removed the host mount.

hack/multi-node/bootkube-up
```diff
@@ -20,7 +20,7 @@ fi
 
 # Render assets
 if [ ! -d "cluster" ]; then
-  ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.101:443 ${etcd_render_flags}
+  ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.101:443 ${etcd_render_flags} --experimental-calico-network-policy
```
yifan-gu: Was thinking about maybe add an env for this, which defaults to false. And add the experimeltal flag only when the env is set to true.
Something similar to the `SELF_HOST_ETCD`.

hack/quickstart/init-master.sh
```diff
@@ -57,7 +57,7 @@ function init_master_node() {
     fi
 
     # Render cluster assets
-    /home/${REMOTE_USER}/bootkube render --asset-dir=/home/${REMOTE_USER}/assets ${etcd_render_flags} \
+    /home/${REMOTE_USER}/bootkube render --asset-dir=/home/${REMOTE_USER}/assets ${etcd_render_flags} --experimental-calico-network-policy \
```
yifan-gu: ditto

e2e/network_test.go
```diff
@@ -0,0 +1,307 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/client-go/pkg/api"
+	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
+)
+
+func TestNetwork(t *testing.T) {
```
yifan-gu: Is there to control the test so that it only runs when the network policy is enabled?

hack/single-node/bootkube-up
```diff
@@ -19,7 +19,7 @@ fi
 
 # Render assets
 if [ ! -d "cluster" ]; then
-  ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.100:443 ${etcd_render_flags}
+  ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.100:443 ${etcd_render_flags} --experimental-calico-network-policy
```
yifan-gu: ditto

yifan-gu:
@abhinavdahiya I just have 2 small nits, otherwise LGTM and ready to merge on green.

hack/multi-node/bootkube-up
abhinavdahiya: Sure

e2e/network_test.go
abhinavdahiya: How do you recommend we do that. flag / env variable / detect if calico daemonset exists?

yifan-gu: @abhinavdahiya A simple flag is good enough, we have something similar for etcd scale test: https://github.com/kubernetes-incubator/bootkube/blob/master/e2e/main_test.go#L32

abhinavdahiya:
coreosbot run e2e

yifan-gu:
coreosbot e2e

abhinavdahiya:
coreosbot run e2e

yifan-gu:
Looks like we need a rebase.

yifan-gu:
coreosbot e2e

yifan-gu:
coreosbot e2e

abhinavdahiya:
coreosbot run e2e

pkg/asset/images.go
```diff
@@ -2,10 +2,12 @@ package asset
 
 // DefaultImages are the defualt images bootkube components use.
 var DefaultImages = ImageVersions{
-	Busybox:         "busybox",
 	Etcd:            "quay.io/coreos/etcd:v3.1.8",
 	EtcdOperator:    "quay.io/coreos/etcd-operator:v0.3.2",
 	Flannel:         "quay.io/coreos/flannel:v0.7.1-amd64",
+	FlannelCNI:      "quay.io/coreos/flannel-cni:0.1.0",
```
klausenbusk: Where can I find the source of the `flannel-cni` image? I need some newer binaries* and just stumbled on this.

\* for the portmap plugon to work (https://github.com/containernetworking/plugins/issues/43)

PR #525, opened 2017-05-18
WIP: Make control plane run as non-root.

diegs:
With the exception of a few prvileged services, this makes as many
control plane components as possible run as non-root.

This requires a special hyperkube build that grants CAP_NET_BIND_SERVICE
to the hyperkube binary.

In addition, this removes the use of the flock since /var/lock requires
root access. In order to prevent a checkpointer race condition a simple
'sleep infinite' container is added instead.

Finally, I got tired of fighting my editor in templates.go, so I
removed all the tabs in the file and added a directive to make it behave
more like a YAML file.


diegs:
rktbot run tests

diegs:
rktbot run tests

diegs:
rktbot run checkpointertests

diegs:
rktbot run checkpointer tests

diegs:
Superceded by #789 and eventually others.

PR #526, opened 2017-05-18
Add retry loop to removing bootstrap etcd service.

diegs:
Since the etcd pivot just occurred the apiserver may be temporarily
unavailable. This retries until successful or timeout, but still doesn't
fail the entire process on failure.

### DIFF
--- a/pkg/util/etcdutil/migrate.go
+++ b/pkg/util/etcdutil/migrate.go
@@ -211,6 +211,10 @@ func waitBootEtcdRemoved(etcdServiceIP string) error {
 func cleanupBootstrapEtcdService(kubecli kubernetes.Interface) {
 	if err := wait.Poll(pollInterval, pollTimeout, func() (bool, error) {
 		if err := kubecli.CoreV1().Services(api.NamespaceSystem).Delete(bootstrapEtcdServiceName, &v1.DeleteOptions{}); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/util/etcdutil/migrate.go
```diff
@@ -209,7 +209,13 @@ func waitBootEtcdRemoved(etcdServiceIP string) error {
 }
 
 func cleanupBootstrapEtcdService(kubecli kubernetes.Interface) {
-	if err := kubecli.CoreV1().Services(api.NamespaceSystem).Delete(bootstrapEtcdServiceName, &v1.DeleteOptions{}); err != nil {
-		glog.Errorf("failed to remove bootstrap-etcd-service: %v", err)
+	if err := wait.Poll(pollInterval, pollTimeout, func() (bool, error) {
+		if err := kubecli.CoreV1().Services(api.NamespaceSystem).Delete(bootstrapEtcdServiceName, &v1.DeleteOptions{}); err != nil {
+			glog.Errorf("failed to remove bootstrap-etcd-service: %v", err)
```
aaronlevy: only nit might be to check if this was a "does not exist" error. But not sure that's actually a situation we could realistically get into.

diegs: Done

PR #528, opened 2017-05-18
*: add backup file option to recover subcommand

xiang90:
/cc @diegs 

This is still WIP. But if you want to take a early view. Do it :P. I will keep this updated with my progress, and ping you again when it is finished and tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ bootkube start --asset-dir=my-cluster
 
 ### Recover a downed cluster
 
-In the case of a partial or total control plane outage (i.e. due to lost master nodes) an experimental `recover` command can extract and write manifests from a backup location. These manifests can then be used by the `start` command to reboot the cluster. Currently recovery from a running apiserver or external running etcd cluster are the only supported methods.
+In the case of a partial or total control plane outage (i.e. due to lost master nodes) an experimental `recover` command can extract and write manifests from a backup location. These manifests can then be used by the `start` command to reboot the cluster. Currently recovery from a running apiserver, an external running etcd cluster, or an etcd backup taken from the self hosted etcd cluster are the methods.
 
 To see available options, run:
 
@@ -78,7 +78,13 @@ Recover from a running apiserver (i.e. if the scheduler pods are all down):
 bootkube recover --asset-dir=recovered --kubeconfig=/etc/kubernetes/kubeconfig
 ```
 
-For a complete recovery example please see the [hack/multi-node/bootkube-test-recovery](hack/multi-node/bootkube-test-recovery) script.
+Recover from an etcd backup when self hosted etcd is enabled:
+
+```
+bootkube recover --asset-dir=recovered --etcd-backup-file=backup --kubeconfig=/etc/kubernetes/kubeconfig
+```
+
+For a complete recovery example please see the [hack/multi-node/bootkube-test-recovery](hack/multi-node/bootkube-test-recovery) and the [hack/multi-node/bootkube-test-recovery-self-hosted-etcd](hack/multi-node/bootkube-test-recovery-self-hosted-etcd) scripts.
 
 ## Building
 
--- a/cmd/bootkube/recover.go
+++ b/cmd/bootkube/recover.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/kubernetes-incubator/bootkube/pkg/bootkube"
 	"github.com/kubernetes-incubator/bootkube/pkg/recovery"
+	"github.com/kubernetes-incubator/bootkube/pkg/util/etcdutil"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/spf13/cobra"
@@ -73,27 +74,30 @@ func runCmdRecover(cmd *cobra.Command, args []string) error {
 	case recoverOpts.etcdBackupFile != "":
 		bootkube.UserOutput("Attempting recovery using etcd backup file at %q...\n", recoverOpts.etcdBackupFile)
 
-		err := recovery.StartRecoveryEtcdForBackup(recoverOpts.podManifestPath, recoverOpts.etcdBackupFile)
+		err = recovery.StartRecoveryEtcdForBackup(recoverOpts.podManifestPath, recoverOpts.etcdBackupFile)
 		if err != nil {
 			return err
 		}
 		defer func() {
-			err := recovery.CleanRecoveryEtcd(recoverOpts.podManifestPath)
+			err = recovery.CleanRecoveryEtcd(recoverOpts.podManifestPath)
 			if err != nil {
 				bootkube.UserOutput("Failed to cleanup recovery etcd from the podManifestPath %v\n", recoverOpts.podManifestPath)
 			}
 		}()
 
 		bootkube.UserOutput("Waiting for etcd server to start...\n")
-		// TODO: better way to detect the startup...
-		time.Sleep(60 * time.Second)
 
-		recoverOpts.etcdServers = "http://localhost:32379"
+		err = etcdutil.WaitClusterReady(recovery.RecoveryEtcdClientAddr)
+		if err != nil {
+			return err
+		}
+
+		recoverOpts.etcdServers = recovery.RecoveryEtcdClientAddr
 		etcdClient, err := createEtcdClient()
 		if err != nil {
 			return err
 		}
-		backend = recovery.NewEtcdBackend(etcdClient, recoverOpts.etcdPrefix)
+		backend = recovery.NewSelfHostedEtcdBackend(etcdClient, recoverOpts.etcdPrefix, recoverOpts.etcdBackupFile)
 
 	default:
 		bootkube.UserOutput("Attempting recovery using apiserver at %q...\n", recoverOpts.kubeConfigPath)
@@ -122,6 +126,9 @@ func validateRecoverOpts(cmd *cobra.Command, args []string) error {
 	}
 	if recoverOpts.kubeConfigPath == "" {
 		return errors.New("missing required flag: --kubeconfig")
+	}
+	if recoverOpts.etcdBackupFile != "" && recoverOpts.podManifestPath == "" {
+		return errors.New("missing required flag: --pod-manifest-path (--etcd-backup-file flag is specified)")
 	}
 	return nil
 }
--- a/cmd/bootkube/recover.go
+++ b/cmd/bootkube/recover.go
@@ -36,6 +36,8 @@ var (
 		etcdServers         string
 		etcdPrefix          string
 		kubeConfigPath      string
+		podManifestPath     string
+		etcdBackupFile      string
 	}
 )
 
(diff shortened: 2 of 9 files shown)


cmd/bootkube/recover.go
```diff
@@ -46,8 +48,10 @@ func init() {
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdCertificatePath, "etcd-certificate-path", "", "Path to an existing certificate that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-private-key-path, and must have etcd configured to use TLS with matching secrets.")
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdPrivateKeyPath, "etcd-private-key-path", "", "Path to an existing private key that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-certificate-path, and must have etcd configured to use TLS with matching secrets.")
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdServers, "etcd-servers", "", "List of etcd server URLs including host:port, comma separated.")
+	cmdRecover.Flags().StringVar(&recoverOpts.etcdServers, "etcd-backup-file", "", "Path to the etcd backup file.")
```
diegs: `&recoverOpts.etcdBackupFile`?

cmd/bootkube/recover.go
```diff
@@ -66,6 +70,22 @@ func runCmdRecover(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		backend = recovery.NewEtcdBackend(etcdClient, recoverOpts.etcdPrefix)
+	case recoverOpts.etcdBackupFile != "":
```
diegs: Maybe just move this above the switch, set the etcdServer, and then let it fall througH?

pkg/recovery/etcd.go
```diff
@@ -96,3 +99,28 @@ func (s *etcdBackend) list(ctx context.Context, key string, listObj runtime.Obje
...
+	d, f := path.Split(backupPath)
+
+	config := struct {
+		Image      string
+		BackupFile string
+		BackupDir  string
+	}{
+		// TODO: this already exists in bootkube/cmd.
+		// do not duplicate this!
+		Image:      "quay.io/coreos/etcd:v3.1.6",
```
diegs: I'd be ok with factoring out an `images` package.

cmd/bootkube/recover.go
```diff
@@ -66,6 +70,22 @@ func runCmdRecover(cmd *cobra.Command, args []string) error {
...
+		bootkube.UserOutput("Attempting recovery using etcd backup file at %q...\n", recoverOpts.etcdBackupFile)
+
+		err := recovery.StartRecoveryEtcd(recoverOpts.podManifestPath, recoverOpts.etcdBackupFile)
+		if err != nil {
+			return err
+		}
+		defer recovery.CleanRecoveryEtcd(recoverOpts.podManifestPath)
```
diegs: nit: ignoring error returned by function.

pkg/recovery/etcd.go
```diff
@@ -96,3 +99,28 @@ func (s *etcdBackend) list(ctx context.Context, key string, listObj runtime.Obje
 	}
 	return decodeList(elems, listPtr, s.decoder)
 }
+
+const assetPathRecoveryEtcd = "recovery-etcd.yaml"
+
+func StartRecoveryEtcd(p, backupPath string) error {
```
diegs: nit: would put "Backup" somewhere in the name to be clearer.

cmd/bootkube/recover.go
xiang90: thought about it. but the logging will be duplicated. it will look like

```
Attempting recovery using etcd backup file at ...
Attempting recovery using etcd cluster at ...
```

but, yea, i will try to clean up code duplication here in another way.

pkg/recovery/etcd.go
```diff
@@ -31,6 +37,16 @@ func NewEtcdBackend(client *clientv3.Client, pathPrefix string) Backend {
 	}
 }
 
+// NewSelfHostedEtcdBackend constructs a new etcdBackend for the given client and pathPrefix, and backup file.
+func NewSelfHostedEtcdBackend(client *clientv3.Client, pathPrefix, backupPath string) Backend {
+	return &etcdBackend{
```
diegs: I think this violates the SRP.

Why not create a new type, etcdBackupBackend, which inherits etcdBackend, calls etcdBackend.Read(), and then adds the bootEtcd asset to the results?

xiang90: will do.

cmd/bootkube/recover.go
diegs: It's not a big deal.

cmd/bootkube/recover.go
```diff
@@ -66,6 +70,31 @@ func runCmdRecover(cmd *cobra.Command, args []string) error {
...
+		}
+		defer func() {
+			err := recovery.CleanRecoveryEtcd(recoverOpts.podManifestPath)
+			if err != nil {
+				bootkube.UserOutput("Failed to cleanup recovery etcd from the podManifestPath %v\n", recoverOpts.podManifestPath)
+			}
+		}()
+
+		bootkube.UserOutput("Waiting for etcd server to start...\n")
+		// TODO: better way to detect the startup...
```
diegs: I have a poll loop in pkg/util/etcdutil/migrate.go#createBootstrapEtcdService that we can factor out

pkg/recovery/etcd_template.go
```diff
@@ -0,0 +1,135 @@
...
+    - -ec
+    - |
+      /usr/local/bin/etcd \
+      --listen-client-urls=http://0.0.0.0:32379 \
+      --listen-peer-urls=http://0.0.0.0:32380 \
+      --advertise-client-urls=http://localhost:32379 \
+      --data-dir=/var/etcd/data && \
+      etcdctl \
+      --endpoints=http://localhost:32379 \
+      del /registry/ThirdPartyResourceData/etcd.coreos.com/clusters/kube-system/kube-etcd
```
diegs: This is slick, but we won't know if it fails. Might have to use the client instead.

xiang90: there is no where in the code we can do this correctly. 

we have to delete the TPR after we recovered the boot etcd, but before API server starts to use it.

Maybe you can try to find a better way to do this?

diegs: My original understanding was that you would:
(1) Start temporary etcd from backup, creating a datadir and extracting the manifests
(2) The bootstrap etcd would talk to the datadir you created in (1)

Thus, I thought you'd be able to delete the TPR from the running etcd in (1). However, it looks like (2) also starts up directly from the backup? In that case then I agree that there's no place in code to remove the TPR, so I suppose this should work. I'd say let's go ahead and try it.

xiang90: > However, it looks like (2) also starts up directly from the backup?

Yes. It was not possible to reuse the same datadir since (2) needs Pod.IP, which is hard to know when we create datadir for (1). But as we start to use a well-known service IP for boot-etcd. It is possible now.

I still choose to use the backup file approach since we can save one flag for bootkube recovery. Or user needs to specify another etcd data dir which will persist after bootkube recover for the bootkube start command.

But it is up to you to decide which we should go.

diegs: Since we are recovering to a known asset-dir, it would be possible to put the etcd dat dir in there (and use the service IP).

If the approach you are using appears to work right now, then I'd say let's stick with it to get something in and add a TODO/open an issue to explore the slightly cleaner approach.

xiang90:
@diegs This works now. Can you take a look? I am trying to extend `bootkube-test-recovery` to include self hosted case.

if you want to give this a try, you can do:

1. SELF_HOST_ETCD=true ./bootkube-up

1. wait the control plane to be up

1. ssh into c1 and copy /var/etcd/kube-etcd-0000/member/snap/db to local machine

1. recreate c1

1. copy db file back to c1 as /home/core/db

1. sudo ./bootkube recover --asset-dir=/home/core/recovered --etcd-backup-file=/home/core/db --kubeconfig=/home/core/kubeconfig

1. sudo /home/core/bootkube start --asset-dir=/home/core/recovered

xiang90:
@diegs I added a test script.

cmd/bootkube/recover.go
```diff
@@ -46,8 +49,10 @@ func init() {
...
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdPrivateKeyPath, "etcd-private-key-path", "", "Path to an existing private key that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-certificate-path, and must have etcd configured to use TLS with matching secrets.")
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdServers, "etcd-servers", "", "List of etcd server URLs including host:port, comma separated.")
+	cmdRecover.Flags().StringVar(&recoverOpts.etcdBackupFile, "etcd-backup-file", "", "Path to the etcd backup file.")
 	cmdRecover.Flags().StringVar(&recoverOpts.etcdPrefix, "etcd-prefix", "/registry", "Path prefix to Kubernetes cluster data in etcd.")
 	cmdRecover.Flags().StringVar(&recoverOpts.kubeConfigPath, "kubeconfig", "", "Path to kubeconfig for communicating with the cluster.")
+	cmdRecover.Flags().StringVar(&recoverOpts.podManifestPath, "pod-manifest-path", "/etc/kubernetes/manifests", "The location where the kubelet is configured to look for static pod manifests. (Only need to be set when recovering from a etcd backup file)")
```
diegs: Please add validation that this flag is non-empty iff etcd-backup-file is non-empty (yes, even though it has a default value).

cmd/bootkube/recover.go
```diff
@@ -66,6 +71,34 @@ func runCmdRecover(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		backend = recovery.NewEtcdBackend(etcdClient, recoverOpts.etcdPrefix)
+	case recoverOpts.etcdBackupFile != "":
+		bootkube.UserOutput("Attempting recovery using etcd backup file at %q...\n", recoverOpts.etcdBackupFile)
+
+		err := recovery.StartRecoveryEtcdForBackup(recoverOpts.podManifestPath, recoverOpts.etcdBackupFile)
```
diegs: nit: scope errs inside if statements (here and 2 places below)

xiang90:
@diegs Thanks for the review. All fixed.

diegs:
@xiang90 awesome! One last thing: can you please update the README? Then I think we're good to merge.

xiang90:
@diegs 

> One last thing: can you please update the README? Then I think we're good to merge.

Sure. I will do it late today or tomorrow (need to attend coreos fest).

xiang90:
@diegs readme updated.

README.md
```diff
@@ -78,7 +78,13 @@ Recover from a running apiserver (i.e. if the scheduler pods are all down):
 bootkube recover --asset-dir=recovered --kubeconfig=/etc/kubernetes/kubeconfig
 ```
 
-For a complete recovery example please see the [hack/multi-node/bootkube-test-recovery](hack/multi-node/bootkube-test-recovery) script.
+Recover from an etcd backup when self hosted etcd is enabled:
```
diegs: Please update line 61 to add the new supported case, and add a section just like lines 72/78 with a command line example. Thanks!

xiang90: i think the command line example is already added two line below, right? or you want something different? i will update line 61.

diegs: Whoops, sorry totally misread the diff. That SG, thanks!

xiang90:
@diegs 

CI failed with:

>         run.go:147: creating cluster: error calling NewMachines: <nil> ssh journalctl failed: dial tcp 104.196.152.95:22: getsockopt: connection refused

I do not think this is relevant with my PR. So I am going to merge this PR.

diegs:
rktbot run tests

PR #529, opened 2017-05-19
wip: e2e reboot test

pbx0:
This ports over the reboot test based on the ssh code yifan wrote. Reboot all nodes. Wait until same number of nodes comes back.

Currently, having trouble getting SSH to work which is why I labeled WIP. 

Also there is a potential race condition in which extra nodes join the cluster while the reboot test is running. This means not all nodes would reboot and its possible that a new node joining the cluster could take the place of a node that died making the test pass when it should fail. One way around is to block tests from starting until all expected nodes have joined and are ready. Another way around is keeping track of node names and waiting until those particular nodes are back. With the second method its still possible we don't end up rebooting all nodes that we created. 

cc @yifan-gu @aaronlevy


e2e/main_test.go
```diff
@@ -65,7 +69,13 @@ func TestMain(m *testing.M) {
 }
 
 func createNamespace(c kubernetes.Interface, name string) (*v1.Namespace, error) {
-	namespace, err := c.CoreV1().Namespaces().Create(&v1.Namespace{
+	namespace, err := c.CoreV1().Namespaces().Get(name, metav1.GetOptions{})
```
diegs: Is this strictly necessary? You can also just create and check if the error is https://godoc.org/k8s.io/apimachinery/pkg/api/errors#IsAlreadyExists. (That approach also removes a race)

e2e/reboot_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if err := nodesReady(client, len(nodeList.Items), t); err != nil {
+		t.Fatalf("Some or all nodes did not recover from reboot: %v", err)
+	}
+
+}
+
+// block until n nodes are ready
+func nodesReady(c kubernetes.Interface, expectedNodes int, t *testing.T) error {
+	f := func() error {
```
diegs: mega-nit: any reason to not just pass the func directly as an arg to `retry()`? `gofmt` works wonders...

hack/quickstart/init-master.sh
```diff
@@ -7,6 +7,7 @@ CLUSTER_DIR=${CLUSTER_DIR:-cluster}
 IDENT=${IDENT:-${HOME}/.ssh/id_rsa}
 SSH_OPTS=${SSH_OPTS:-}
 SELF_HOST_ETCD=${SELF_HOST_ETCD:-false}
+CLOUD_PROVIDER=${CLOUD_PROVIDER:-}
```
diegs: Is not setting `CLOUD_PROVIDER` an error?

hack/quickstart/init-node.sh
```diff
@@ -7,6 +7,7 @@ REMOTE_PORT=${REMOTE_PORT:-22}
 IDENT=${IDENT:-${HOME}/.ssh/id_rsa}
 SSH_OPTS=${SSH_OPTS:-}
 TAG_MASTER=${TAG_MASTER:-false}
+CLOUD_PROVIDER=${CLOUD_PROVIDER:-}
```
diegs: same question as above.

hack/quickstart/init-master.sh
pbx0: It doesn't error, but i had assumed leaving it as the empty string would be the same as not having it at all.

The flag description however suggests that leaving it as the empty string will disable cloud-provider support rather then auto-detect by default. Now, that I've figured out the IAM permissions for AWS, I'll try testing without the flag and see if it auto-detects.

https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/options/options.go#L240

diegs: Ok, I'm fine with disabling it explicitly (as it is now) when not provided. Auto-detection doesn't feel right to me.

e2e/main_test.go
pbx0: Good call, will change

e2e/reboot_test.go
pbx0: I've always done it this way with the belief that its more readable to declare the function and then pass it to retry. I tried it inline and I kinda like it now though.

diegs: Your call :)

hack/quickstart/init-master.sh
pbx0: sgtm

e2e/reboot_test.go
pbx0: Just leaving as is for now since its done this way in a few places.

pbx0:
PTAL. Added a commit to make reboot test safe against further nodes joining. Unless we decide that the number of nodes in a cluster to expect is a fixed invariant its the best we can do.

e2e/reboot_test.go
```diff
@@ -0,0 +1,105 @@
...
+// nodesReady blocks until all nodes in list are ready based on Name. Safe
+// against new unknown nodes joining while the original set reboots.
+func nodesReady(c kubernetes.Interface, expectedNodes *v1.NodeList, t *testing.T) error {
+	f := func() error {
+		list, err := c.CoreV1().Nodes().List(metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		if len(list.Items) < len(expectedNodes.Items) {
```
diegs: nit: not strictly necessary since you're checking their existence below?

e2e/reboot_test.go
```diff
@@ -0,0 +1,105 @@
...
+			return fmt.Errorf("cluster is not ready, expected %v nodes got %v", len(expectedNodes.Items), len(list.Items))
+		}
+
+		var expectedNodeSet map[string]struct{}
```
diegs: can hoist out of the closure?

pbx0: done

e2e/reboot_test.go
pbx0: done

pbx0:
Passes new jenkins job designed to run these new tests.

PR #540, opened 2017-05-24
etcdutil: factor out wait ready

xiang90:
/cc @diegs 

we want to reuse the cluster wait logic for recovery.

### DIFF
--- a/pkg/util/etcdutil/migrate.go
+++ b/pkg/util/etcdutil/migrate.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"net/http"
 	"time"
 
 	"github.com/kubernetes-incubator/bootkube/pkg/asset"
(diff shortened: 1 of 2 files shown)


pkg/util/etcdutil/migrate.go
```diff
@@ -102,34 +101,22 @@ func waitEtcdTPRReady(restClient restclient.Interface, ns string) error {
...
+	svcb, err := ioutil.ReadFile(svcPath)
 	if err != nil {
 		return err
 	}
-	if err := kubecli.CoreV1().RESTClient().Post().RequestURI(fmt.Sprintf("/api/v1/namespaces/%s/services", api.NamespaceSystem)).SetHeader("Content-Type", "application/json").Body(svc).Do().Error(); err != nil {
+	if err := kubecli.CoreV1().RESTClient().Post().RequestURI(fmt.Sprintf("/api/v1/namespaces/%s/services", api.NamespaceSystem)).SetHeader("Content-Type", "application/json").Body(svcb).Do().Error(); err != nil {
+		return err
+	}
+
+	svc, err := kubecli.CoreV1().Services(api.NamespaceSystem).Get(bootstrapEtcdServiceName, v1.GetOptions{})
```
xiang90: @diegs 

is there a reason we also put this get func into the wait loop? if 108 returns, we should be able to get the service, right?

diegs: Doesn't really matter either way, you can pull it out if you like.

pkg/util/etcdutil/util.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/golang/glog"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// WaitReady waits the etcd server ready to serve client requests.
+func WaitReady(endpoint string) error {
```
diegs: `waitClusterReady`? As in, `etcdutil.WaitClusterReady(endpoint)`

xiang90:
ignore e2e-selfetcd  according to @pbx0.

merging. thanks for the review @diegs !

PR #541, opened 2017-05-24
Fix pod-checkpointer container name.

diegs:
By convention the primary container should match the name of the
daemonset.


aaronlevy:
rktbot run tests

pbx0:
rktbot run e2e

diegs:
rktbot run e2e

PR #546, opened 2017-05-25
e2e: bump smoke test timeout

pbx0:


pbx0:
pluton e2e is flake, e2e-selfetcd is the test actually exercising this code and passes, merging.

PR #543, opened 2017-05-24
Use a variable for the username on coreos to fix the gce quickstart guide

roberthbailey:

PR #550, opened 2017-05-25
hack/quickstart: add ssh reqs for quickstart-gce

pbx0:
Alternative PR to: https://github.com/kubernetes-incubator/bootkube/pull/543

cc @roberthbailey

### DIFF
--- a/hack/quickstart/quickstart-gce.md
+++ b/hack/quickstart/quickstart-gce.md
@@ -8,6 +8,28 @@ This can be changed to identify separate clusters.
 export CLUSTER_PREFIX=quickstart
 ```
 
+### Add your SSH public key to project under user core
+
+Quickstart scripts assume you can ssh into the `core` user. Either follow the commands below or go to the web console to add a key for the user `core`.
+
+**WARNING** The following process will clobber existing keys in your project's ssh metadata.
+
+Make a copy of your public key. It often exists at the path: `~/.ssh/id_rsa.pub`
+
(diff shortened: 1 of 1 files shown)


hack/quickstart/quickstart-gce.md
```diff
@@ -8,6 +8,28 @@ This can be changed to identify separate clusters.
 export CLUSTER_PREFIX=quickstart
 ```
 
+### Add your SSH public key to project under user core
+
+Quickstart scripts assume you can ssh into the `core` user. Either follow the commands below or go to the web console to add a key for the user `core`.
+
+**WARNING** The following process will clobber existing keys in your project's ssh metadata.
```
aaronlevy: eh? This seems a bit more heavy-handed than giving deployer option of select user (https://github.com/kubernetes-incubator/bootkube/pull/543)

pbx0: Yeah I discovered the side effect while testing. I don't have strong feelings really either way on which PR to go forth with, but it seemed simpler in my mind to just require the core user upfront when people are using gce-quickstart.

aaronlevy: The option (which defaults to "core" anyway) doesn't change default behavior -- so that seems reasonable. Telling people to stomp on their project wide keys doesn't seem ideal (i'm not really familiar with gce, so I don't know if there are ways to scope this down to per-instance launches or something.

pbx0: You can do it per-instance, but they list it as deprecated. SGTM, will close in favor of original PR.

PR #553, opened 2017-05-26
Remove apiserver flock timeout.

diegs:


diegs:
rktbot run e2e

diegs:
rktbot run tests

PR #558, opened 2017-06-01
quickstart: Compress binary when sending to nodes

kamalmarhubi:
The binary is ~80 MB. Enabling compression speeds up the quickstart for
"typical" internet connections with << 100 Mbps.


yifan-gu:
LGTM

PR #559, opened 2017-06-01
pkg/bootkube: Fix time-of-check-to-time-of-use in copyFile

kamalmarhubi:
If not overwriting, use the `O_EXCL` flag when opening the destination
file to ensure we create it.


yifan-gu:
@kamalmarhubi Thanks for the fix! LGTM.
Have you signed the cla btw?

PR #561, opened 2017-06-05
Change bootstrap etcd service port to 20.

### LINKED ISSUE #531: Bootstrap self-hosted etcd is using the 200th IP in the service range
Is there a reason to pick the 200th IP in the service range for bootstrap self-hosted etcd? The other IP offsets for apiserver, dns, and self-hosted etcd are 1, 10 and 15 respectively. Did you mean to pick 20?

https://github.com/kubernetes-incubator/bootkube/blob/master/cmd/bootkube/render.go#L24

I ask because this seems to assume the service range is a `/24` or larger. I'd expect inputing a `/25` would quietly fail with self-hosted etcd enabled.

diegs:
Fixes #531

### DIFF
--- a/cmd/bootkube/render.go
+++ b/cmd/bootkube/render.go
@@ -21,7 +21,7 @@ const (
 	apiOffset                    = 1
 	dnsOffset                    = 10
 	etcdOffset                   = 15
-	bootEtcdOffset               = 200
+	bootEtcdOffset               = 20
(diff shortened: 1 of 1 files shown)


xiang90:
lgtm

diegs:
rktbot run tests

dghubble:
:+1: Nice, can you just be sure this gets mentioned in the changelog?

diegs:
@dghubble will do

Issue #557, opened 2017-05-31
Enable logging during bootstrap phase


abhinavdahiya:
Right now the only way to access the kubelet logs and logs of the static pods in the bootstrap control plane on the bootstrap node is to ssh into the node!

It would ideal if we could deploy something like a static fluentd pod on the bootstrap node to upload all the logs to something like a s3 bucket for debugging purposes.

/cc: @aaronlevy @yifan-gu @diegs

abhinavdahiya:
fluentd is a good option for extracting logs as it have various configurable input & output options, esp output to s3 bucket.

fluentd needs a fluent.conf to configure sources & outputs

Some ways we can implement is
- static pod during bootstrap
** fluentd.conf is shipped in asset-dir
** bake fluentd.conf in a custom fluentd-image 
- non static pod after api-server is up.

yifan-gu:
At the first glance, a custom fluentd image seems cleaner than shipping config file in the asset-dir given that it hides the details and the config file should be pretty stable. The cons are that if we do want to make a change on conf file or fluentd image, we need to bake a new image.

Non-static pod might be a less ideal option as it will depend on api server to be running before it can be deployed. However, "why api server is NOT running" is the exact problem we want to debug with the logging.

diegs:
It seems like there are 2 main areas we want to address.

(1) Issues during bootstrap, e.g. the boostrap control plane doesn't come up.
(2) Issues with the self-hosted control plane and downstream operations (this is not bootkube-specific and is more relevant to tectonic-installer, various operators, and test frameworks).

For (1) I think a static fluentd pod that can hook up to something easily accessible on the bootstrap node (e.g. journal logs) would suffice and actually be preferable. Bootkube requires you to ssh to that node anyway to run it. I don't think it's within the scope of the bootkube project to add more dependencies on logging sinks, such as S3 buckets. Currently the bootkube e2e tests fetch the journal logs at the end of the test--that's a simple command to run which we could publish in docs.

For (2) I think we should just recommend that people to deploy a more permanent fluentd dameonset as part of the asset-dir/manifests directory. But again, I think this is environment-specific and doesn't belong in the bootkube project, except perhaps as an example in the hack-scripts.

So to summarize, I think doing something simple for (1) that doesn't add external infrastructure dependencies sounds good, and I think it would be sufficient to just make sure the pod logs are aggregated in a place that's easy to extract.

pbx0:
[...]

PR #564, opened 2017-06-06
Add fluentd during bootstrap

abhinavdahiya:
#557 
collects logs from bootstrap phase to
`var/log/bootkube-bootstrap/containers.*.log` &
`var/log/bootkube-bootstrap/services.*.log` using fluentd on the bootstrap node.

currently the image used is https://github.com/abhinavdahiya/bootkube-fluentd.

/cc: @yifan-gu

### DIFF
--- a/hack/multi-node/bootkube-up
+++ b/hack/multi-node/bootkube-up
@@ -26,6 +26,13 @@ if [ ! -d "cluster" ]; then
   sed -i.bak -e '/node-role.kubernetes.io\/master/d' cluster/user-data-worker
 fi
 
+# Add fluentd static pod
+ENABLE_BOOTSTRAP_LOGS=${ENABLE_BOOTSTRAP_LOGS:-false}
+if [ ${ENABLE_BOOTSTRAP_LOGS} = "true" ]; then
+    echo "Writing asset: cluster/bootstrap-manifests/bootstrap-fluentd.yaml"
+    cp bootstrap-fluentd.sample cluster/bootstrap-manifests/bootstrap-fluentd.yaml
+fi
+
 # Start the VM
 vagrant up
 vagrant ssh-config c1 > ssh_config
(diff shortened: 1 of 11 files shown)


pkg/asset/internal/templates.go
```diff
@@ -225,6 +225,48 @@ spec:
...
+  volumes:
+  - name: varlog
+    hostPath:
+      path: /var/log
+  - name: varlibdockercontainers
+    hostPath:
+      path: /var/lib/docker/containers
+  - name: libsystemddir
+    hostPath:
+      path: /usr/lib64
```
pbx0: Curious about what host libraries fluentd needs access to.

diegs: The fluentd team provides an image that doesn't have this requirement:

https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/fluentd-daemonset-elasticsearch.yaml

abhinavdahiya: https://github.com/abhinavdahiya/bootkube-fluentd/blob/master/run.sh#L6-L10

pkg/asset/images.go
```diff
@@ -12,4 +12,5 @@ var DefaultImages = ImageVersions{
 	KubeDNSMasq:     "gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64:1.14.1",
 	KubeDNSSidecar:  "gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.1",
 	PodCheckpointer: "quay.io/coreos/pod-checkpointer:4e7a7dab10bc4d895b66c21656291c6e0b017248",
+	Fluentd:         "quay.io/abhinavdahiya/bootkube-fluentd:latest",
```
yifan-gu: Needs to host it under quay.io/coreos, do you have access to that? @abhinavdahiya

yifan-gu:
After OOB discussion, we agreed it's better not introduce new commands for now. Just put the program in hack directory and enable it with an environment variable if required.

abhinavdahiya:
I have moved the static pod manifests from bootkube to hack scripts.

@yifan-gu

yifan-gu:
Based on later discussion, we decided to use the log collector #579  in e2e test instead of just bootstrap process. 
Closing for now.

Issue #563, opened 2017-06-05
Set resource requirements for self-hosted etcd


hongchaodeng:
Currently we haven't set any resource requirements for self-hosted etcd pods. This could be done via cluster spec.

For each etcd pod, we should require at least 1GB mem and 1 core cpu.

hongchaodeng:
cc @xiang90 @aaronlevy

ericchiang:
Self-hosted etcd has been removed from bootkube for the time being. Will re-open if/when we add it back in.

PR #566, opened 2017-06-06
SelfHostedEtcd: set resource requests

hongchaodeng:
fix https://github.com/kubernetes-incubator/bootkube/issues/563

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -994,6 +994,12 @@ var EtcdTPRTemplate = []byte(`{
     "size": 1,
     "version": "v3.1.6",
     "pod": {
+      "resources": {
+        "requests": {
+          "memory": "1Gi",
+          "cpu": "1"
+        }
+      },
       "nodeSelector": {
         "node-role.kubernetes.io/master": ""
       },


hongchaodeng:
cc @xiang90 @aaronlevy

pkg/asset/internal/templates.go
```diff
@@ -994,6 +994,12 @@ var EtcdTPRTemplate = []byte(`{
     "size": 1,
     "version": "v3.1.6",
     "pod": {
+      "resources": {
+        "requests": {
+          "memory": "1Gi",
+          "cpu": "1"
```
xiang90: i would even suggest this to be 2...

xiang90:
@aaronlevy is this something we can enforce here?

diegs:
You will probably want to update the hack scripts... e.g. https://github.com/kubernetes-incubator/bootkube/blob/master/hack/multi-node/Vagrantfile#L13

The cloud ones (GCE and AWS) all specify 1-cpu workers too...

hongchaodeng:
So the questions comes to how do we make this happen? Ideally, in a product cluster, we would expect etcd to occupy at least 2 cores to be stable.

Option 1: We can definitely enforce it. But anyone running a machine deployment with <= 2 cores would fail to start.

Option 2: We can expose this as configuration. That's adding complexity over configuring bootstrap tools.

Option 3: We can detect the number of cores in bootstrap tool, and set resource request appropriately. That's complex too.

diegs:
I don't think we need to do any of those options.

For bootkube I think this should be documented, and we should modify the hack scripts to use these minimum requirements (at least for the cloud examples--I don't think the VM examples need to be production-ready).

For other projects such as https://github.com/coreos/tectonic-installer, the installer should enforce these requirements.

cc @aaronlevy

aaronlevy:
2gb and 2cpu's seems a little high for the outset (but to be fair, I don't have data to back this up).

I feel like we should have a low limit (reasonably functional dev cluster) and then hopefully can link to better information about how to size your etcd cluster based on actual needs.

FWIW it looks like upstream only has a request of 200m: https://github.com/kubernetes/kubernetes/blob/master/cluster/saltbase/salt/etcd/init.sls#L68

hongchaodeng:
> FWIW it looks like upstream only has a request of 200m: 

That's a bad example. We have talked in upstream a couple of times before to change it. But it didn't happen unfortunately.

> (but to be fair, I don't have data to back this up).
> I feel like we should have a low limit (reasonably functional dev cluster) and then hopefully can link to better information about how to size your etcd cluster based on actual needs.

etcd provides some data and sizing recommendation docs: https://github.com/coreos/etcd/blob/master/Documentation/op-guide/hardware.md#hardware-recommendations

hongchaodeng:
@diegs 
Sorry I didn't catch it.
Can you tell me more detailedly and explicitly more to do?

aaronlevy:
@hongchaodeng what would be a minimum reasonable amount to set for a dev cluster? This would need to be also changed for all our hack examples / cloud examples. Also note that whatever minimum is set will need to be used for all of our test infra -- so if it's high, it's gonna get more expensive.

Then in this repo we can add to `Documentation/Production` a note about setting resource requests higher (and can link to that etcd doc). Tectonic I think is a more reasonable place to enforce more "production-like" resource limits (but with testing infra the same thing applies).

diegs:
@hongchaodeng basically what @aaronlevy said in his last comment:

- we should add the etcd resource requirements to the README somewhere
- we should update the various scripts under https://github.com/kubernetes-incubator/bootkube/blob/master/hack/ to make sure they run with enough resources for etcd.

klausenbusk:
> 2gb and 2cpu's seems a little high for the outset (but to be fair, I don't have data to back this up).

From my point of view, that also seems a bit overkill, at least for small clusters.
My [coreos-kubernetes](https://github.com/coreos/coreos-kubernetes) cluster which consists of 3 master nodes and 5 worker nodes, the 3 etcd members currently use: 237M, 192M and 168M. For small cluster 256M or 512M should be enough, 1GB or even 2GB seems way to high and add a lot of overhead for no reason (??).

xiang90:
we decided to enforce this in tectonic instead. probably we can close this in bootkube?

/cc @hongchaodeng @aaronlevy

hongchaodeng:
Let's close this

PR #578, opened 2017-06-13
WIP: attempt to use hostport plugin with 1.7

diegs:
In theory this may let us run the apiserver as non-root.

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -1109,12 +1109,38 @@ spec:
           mountPath: /etc/kube-flannel/
       - name: install-cni
         image: {{ .Images.Busybox }}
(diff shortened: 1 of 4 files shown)


pkg/asset/internal/templates.go
```diff
@@ -1034,11 +1039,23 @@ metadata:
 data:
   cni-conf.json: |
```
abhinavdahiya: @diegs 
Looks like the cni config file name should have `.conflist` suffix to load list of plugins
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/cni/cni.go#L110-L127

squeed:
Two comments:
1) Flannel isn't working with hostport forwarding because of containernetworking/plugins#23
2) Flannel needs to talk to the etcd apiserver in order to bring a container online; taking the apiserver off the host network seems operationally dangerous.

diegs:
Obsolete, closing.

Issue #582, opened 2017-06-14
Add TLS to self-hosted etcd


hongchaodeng:
Currently self-hosted etcd mode doesn't have TLS option. We want to make use of static TLS feature in etcd operator and provide that in self-hosted etcd.

Attach plan in bootkube:
- ~~Add flags to pass in etcd server's client and peer communication x509 certs~~
- [X] Auto-generate 3 (CA, key, cert) certificate suites for self-hosted etcd (https://github.com/kubernetes-incubator/bootkube/issues/585)
  - operator etcd client SAN: none
  - etcd server client port SAN:
    - etcd service IP
    - boot etcd service IP
    - 127.0.0.1
    - localhost
    - `*.kube-etcd.kube-system.svc.cluster.local`
    - `kube-etcd-client.kube-system.svc.cluster.local`
  - etcd server peer port SAN:
    - boot etcd service IP
    - `*.kube-etcd.kube-system.svc.cluster.local`
    - `kube-etcd-client.kube-system.svc.cluster.local`
- [x] Render (etcd client's x509 CA, key, cert), (etcd server's client communication x509 CA, key, cert), (etcd server's peer communication x509 CA, key, cert) into secrets (https://github.com/kubernetes-incubator/bootkube/pull/586)
- [x] Auto-detect TLS assets in self-hosted and create secure etcd client as needed. (https://github.com/kubernetes-incubator/bootkube/pull/587)
- [x] Have option to start secure self-hosted etcd cluster and enable it by default (https://github.com/kubernetes-incubator/bootkube/pull/592)
  - Add static TLS with above secrets to etcd TPR
  - Change bootstrap etcd to use TLS
  - Change APIServer to use https to talk to etcd

hongchaodeng:
@diegs @aaronlevy 
To continue discuss at https://github.com/kubernetes-incubator/bootkube/pull/583#discussion_r122050298,
I have updated the top comment to reflect what would be included in auto-generated certs.

aaronlevy:
Are localhost and 127.0.0.1 necessary?

I'm unsure about only adding san's for the full dns name (`.kube-system.svc.cluster.local`) You don't actually have to use full domain. Would we want to have san's, for example, of:

kube-etcd-client
kube-etcd-client.kube-system
kube-etcd-client.kube-system.svc
kube-etcd-client.kube-system.cluster.local
*.kube-etcd
*.kube-etcd.kube-system
*.kube-etcd.kube-system.svc
[...]

PR #583, opened 2017-06-14
render: add flags for self-hosted etcd certs

hongchaodeng:


cmd/bootkube/render.go
```diff
@@ -65,6 +71,12 @@ func init() {
...
 	cmdRender.Flags().StringVar(&renderOpts.etcdPrivateKeyPath, "etcd-private-key-path", "", "Path to an existing private key that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-certificate-path, and must have etcd configured to use TLS with matching secrets.")
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdClientCAPath, "self-hosted-etcd-client-ca", "", "Path to an existing PEM encoded CA that will be used for self-hosted etcd's client port.")
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdClientCertPath, "self-hosted-etcd-client-cert", "", "Path to an existing certificate that will be used for self-hosted etcd's client port.")
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdClientKeyPath, "self-hosted-etcd-client-key", "", "Path to an existing private key that will be used for self-hosted etcd's client port.")
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdPeerCAPath, "self-hosted-etcd-peer-ca", "", "Path to an existing PEM encoded CA that will be used for  self-hosted etcd's peer port.")
```
diegs: extra space in `for  self` (next line too)

cmd/bootkube/render.go
```diff
@@ -65,6 +71,12 @@ func init() {
...
 	cmdRender.Flags().StringVar(&renderOpts.etcdCertificatePath, "etcd-certificate-path", "", "Path to an existing certificate that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-private-key-path, and must have etcd configured to use TLS with matching secrets.")
 	cmdRender.Flags().StringVar(&renderOpts.etcdPrivateKeyPath, "etcd-private-key-path", "", "Path to an existing private key that will be used for TLS-enabled communication between the apiserver and etcd. Must be used in conjunction with --etcd-ca-path and --etcd-certificate-path, and must have etcd configured to use TLS with matching secrets.")
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdClientCAPath, "self-hosted-etcd-client-ca", "", "Path to an existing PEM encoded CA that will be used for self-hosted etcd's client port.")
```
diegs: Is there a way to combine these flags with the non-self hosted flags? Since there's already a flag that enables self-hosted etcd, the behavior of the other flags would simply change based on that mode.

cmd/bootkube/render.go
```diff
@@ -65,6 +71,12 @@ func init() {
...
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdClientCertPath, "self-hosted-etcd-client-cert", "", "Path to an existing certificate that will be used for self-hosted etcd's client port.")
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdClientKeyPath, "self-hosted-etcd-client-key", "", "Path to an existing private key that will be used for self-hosted etcd's client port.")
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdPeerCAPath, "self-hosted-etcd-peer-ca", "", "Path to an existing PEM encoded CA that will be used for  self-hosted etcd's peer port.")
+	cmdRender.Flags().StringVar(&renderOpts.selfHostedEtcdPeerCertPath, "self-hosted-etcd-peer-cert", "", "Path to an existing certificate that will be used for self-hosted etcd's peer port.")
```
diegs: Will `render` generate certs in any scenario, or do you always have to specify all 6 flags?

Please add usage information and validation to the validation method below.

cmd/bootkube/render.go
hongchaodeng: > Is there a way to combine these flags with the non-self hosted flags? 

I'm following what it does on apiserver's etcd client certs. I'm fine with whatever way you want to deal with configs. But I would like to move forward my other work for https://github.com/kubernetes-incubator/bootkube/issues/582 instead of optimizing flags right now.

cmd/bootkube/render.go
hongchaodeng: > Will render generate certs in any scenario, or do you always have to specify all 6 flags?

No. The CA could not be used to generate certs. We have this discussion previously in etcd operator on TLS story like taking only the CA and generating certs on our own. But it doesn't work. We couldn't make assumptions on this. But I'm open to any suggestions to make it more automated.

hongchaodeng:
fixed

hongchaodeng:
@diegs @aaronlevy 
Can we move forward this?

cmd/bootkube/render.go
diegs: We already generate certs for external etcd if they are not provided:

https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/tls.go#L118

The functionality should be similar for self-hosted etcd (and shouldn't require many changes).

However, I would like to get this right now, rather than optimizing flags later. It shouldn't be much work. I expect @aaronlevy would agree.

Could you please take a pass at combining flags and making the functionality match was we do for non-self-hosted? It will also help you when implementing tests and updating hack scripts.

hongchaodeng: OH. I see what you meant. That sounds good.

hongchaodeng:
Closing this PR since we are going to generate self-hosted etcd certs automatically. See https://github.com/kubernetes-incubator/bootkube/pull/583#discussion_r122050298 .

hongchaodeng:
@diegs 
Just to make sure above comment is right.

diegs:
@hongchaodeng well, up to you. For non-self-hosted etcd we support passing in existing certs, falling back to generating new certs when they are not passed in.

Up to you as to what makes sense for self-hosted etcd.

Issue #112, opened 2016-08-30
Simple scheduler & controller-manager disaster recovery


aaronlevy:
There are potential failure cases, where you have permanently lost all schedulers and/or all controller-managers, where recovery leaves you in a chicken-egg state: 

For example, assume you have lost all schedulers - but still have a functioning api-server which contains the scheduler deployment object:

You need a controller-manager to convert the deployment into unscheduled pods, and you need a scheduler to then schedule those pods to nodes (scheduler to schedule the scheduler, if you will).

While these types of situations should be mitigated by deploying across failure domains, it is still something we need to cover.

In the short term this could mean documenting, for example, how to create a temporary scheduler pod pre-assigned to a node (once a single scheduler exists, it will then schedule the rest of the scheduler pods).

Another option might be to build a tool which knows how to do this for you based on parsing an existing object:

```
kube-recover deployment kube-controller-manager --target=<node-ip>
kube-recover deployment kube-scheduler --target=<node-ip>
```

Where the `kube-recover` tool could read the object from the api-server (or from disk), parse out the podSpec and pre-assign a pod to the target node (bypassing both need for controller-manager and scheduler).

aaronlevy:
Thinking longer term, rather than a separate `kube-recover` tool, this could be a command in `kubectl` which knows how to extract a pod from higher-order object -- this could be useful in the case where we have a kubelet-pod-api (so we could use this to natively push pod to kubelet api from a deployment/daemonset object rather than needing the intermediate pod state). See https://github.com/coreos/bootkube/issues/97#issuecomment-238311771

chancez:
👍 I think an interesting idea is to play with re-running `bootkube`. I found that generally if I just re-run bootkube nothing too crazy happens and the temporary control plane brings up a scheduler, allowing everything to correct itself. I really like the UX of this approach, even if its not a real solution. 

In the static manifest case this problem isnt really a problem because the pods which run aren't scheduled, so the scheduler failing in any scenario always results in it being restarted and able to run (assuming it can grab it's lease). 

I wonder if there's a good way to do this, perhaps by check pointing the scheduler to nodes which have `role=master` would be a good start, as an alternative to your proposed `kube-recover` strategy? This would be mostly automated, the only issue is how do you run this checkpointer in a way that works when the scheduler runs into this? Well, that's probably a static pod, and that means we can't easily manage it, which is problematic if the master nodes are changing.

chancez:
[...]

Issue #311, opened 2017-02-16
Instructions for multi master


rmenn:
Greetings, 

I have been trying to ask this on irc as well as the k8s slack, but am restoring to a ticket here. I apologize. 

I wanted to know if a multi master setup is possible with bootkube, if so how do i do it, especially with the experimental etcd flag set.

Just need someone to point me in the right direction. 

Thanks

bzub:
For Master k8s components you should only need to give another node the `master=true` label.

```
kubectl label node node2.zbrbdl "master=true"
```

Then you can either create a LoadBalancer with external IP to the kubernetes service (default namespace) or point external DNS to one or all of the API server nodes for kubectl clients to use.

For self-hosted-etcd you can try the [steps](https://github.com/coreos/etcd-operator#resize-an-etcd-cluster) in the etcd-operator README.  I haven't tried that yet myself.

aaronlevy:
As @bzub points out, simply labeling the node as a master will start master components on these nodes. The main change is that you need some way of addressing your multiple api-servers from a single address. kubeconfig only supports a single api-server address, and even though you can specify multiple on the kubelet command line, only the first is really used.

So a loadbalancer which fronts all api-servers (master nodes), or DNS entry which maps to those nodes are usually good options. You would then set your api server address in the kubeconfig to point to the dns or loadbalancer.

There is also somewhat of a limitation in the internal kubernetes service where the multiple api-servers will all overwrite eachother as the only endpoint. (To see this `kubectl get service kubernetes` - if you have multiple apiservers running, the endpoint will constantly change).

This isn't the worst thing in the world, but it's not ideal (there is work to resolve this upstream). In the interim, an option is to also use your loadbalancer/dns entry for this endpoint as well, which can be done by setting the apiserver `--advertise-address=<your loadbalancer>` - so the endpoint will always point to the same location.

bzub:
I'm just starting to implement an automated HA failover system for kube-apiserver with [keepalived-vip](https://github.com/kubernetes/contrib/tree/master/keepalived-vip) and @aaronlevy your comment about the default kubernetes service was very enlightening.  I really would have overlooked that issue, as limited as it is.

[...]

PR #584, opened 2017-06-15
Add disaster recovery documentation.

### LINKED ISSUE #432: Documentation: Disaster recovery scenarios

diegs:
Fixes #432, also one way to address issues raised in #112.


diegs:
cc @coresolve

Documentation/disaster-recovery.md
```diff
@@ -0,0 +1,112 @@
...
+
+### If an etcd backup is available (self-hosted etcd)
+
+If using self-hosted etcd, recovery is supported via reading from an etcd
+backup file:
+
+```
+bootkube recover --asset-dir=recovered --etcd-backup-file=backup --kubeconfig=/etc/kubernetes/kubeconfig
+```
+
```
xiang90: we tried this with our field engs yesterday. 

there are a few things we need to make sure before run this script:

1. kubelet is running on the machine
2. no related containers are running (old etcd, old api server, etc.. this is also applied to other recovery cases i believe)
3. docker state is clean (docker ps -a does not contain old states of relevant containers). kubelet has bugs that it might incorrectly believes the static pod has dead when old state exists.
4. /var/etcd dir is clean on ALL master nodes

Documentation/disaster-recovery.md
```diff
@@ -0,0 +1,112 @@
...
+[![asciicast](https://asciinema.org/a/dsp43ziuuzwcztni94y8l25s5.png)](https://asciinema.org/a/dsp43ziuuzwcztni94y8l25s5)
+
+### If an api-server is still running
+
+If an api-server is still running but other control plane components are down,
+preventing cluster functionality (i.e. the scheduler pods are all down), the
+control plane can be extracted directly from the api-server:
+
+```
+bootkube recover --asset-dir=recovered --kubeconfig=/etc/kubernetes/kubeconfig
```
xiang90: one relevant issue: field engs suggest to rename --asset-dir to output-asset-dir. when they first tried without our help, they tried to pass in the old asset-dir in here.

xiang90:
Awesome start!

Documentation/disaster-recovery.md
diegs: Thanks for the feedback! Created #589

Documentation/disaster-recovery.md
diegs: Cool, do you want me to add this directly to the documentation?

xiang90: @diegs just fyi. we can address them later.

xiang90:
LGTM

Documentation/disaster-recovery.md
```diff
@@ -0,0 +1,112 @@
...
+failure scenarios:
+
+- Loss of all api-servers
+- Loss of all schedulers
+- Loss of all controller-managers
+- Loss of all self-hosted etcd nodes
+
+To minimize the likelihood of any of the these scenarios, production
+self-hosted clusters should always run in a [high-availability
+configuration](https://kubernetes.io/docs/admin/high-availability/).
```
aaronlevy: I'm on the fence about linking to those docs -- as they're pretty different to how self-hosted HA works (which we need docs for: https://github.com/kubernetes-incubator/bootkube/issues/311). It does touch on some important topics like leader-election, but even then we already have that and all we care about is scaling replica counts (for example).

Documentation/disaster-recovery.md
```diff
@@ -0,0 +1,112 @@
...
+  to this node.
+
+A Pod Checkpointer DaemonSet is deployed by default when using `bootkube
+render` to create cluster manifests. Using the Pod Checkpointer is highly
+recommended for all self-hosted clusters to ensure node reboot resiliency.
+
+For more information, see the [Pod Checkpointer
+README](https://github.com/kubernetes-incubator/bootkube/blob/master/cmd/checkpoint/README.md).
+
+## Bootkube Recover
```
aaronlevy: We may want to have some kind of versioning convention. I'm assuming right now it's: you should always use the latest bootkube release when running recover. This may not be a confusion point, but I wonder if users will try and use the same bootkube release that they installed with (which is probably fine in most cases, unless there are new bug fixes they should have).

Documentation/disaster-recovery.md
diegs: Ok, added a TODO linking to the issue instead for now.

Documentation/disaster-recovery.md
diegs: Added note recommending to always use the latest version.

coresolve:
We need to test this against a tectonic cluster. Not just the bootkube rendered cluster. Looks really good though.

diegs:
@coresolve sgtm, we should go through that (and especially with self-hosted etcd) next.

PR #585, opened 2017-06-15
selfHostedEtcd: autogen certs from CA cert

hongchaodeng:
ref: https://github.com/kubernetes-incubator/bootkube/issues/582

### DIFF
--- a/cmd/bootkube/render.go
+++ b/cmd/bootkube/render.go
@@ -99,9 +99,6 @@ func validateRenderOpts(cmd *cobra.Command, args []string) error {
 	if (renderOpts.etcdCAPath != "" || renderOpts.etcdCertificatePath != "" || renderOpts.etcdPrivateKeyPath != "") && (renderOpts.etcdCAPath == "" || renderOpts.etcdCertificatePath == "" || renderOpts.etcdPrivateKeyPath == "") {
 		return errors.New("You must specify either all or none of --etcd-ca-path, --etcd-certificate-path, and --etcd-private-key-path")
 	}
-	if renderOpts.etcdCertificatePath != "" && renderOpts.selfHostedEtcd {
-		return errors.New("Cannot specify --etcd-certificate-path with --experimental-self-hosted-etcd")
-	}
 	if renderOpts.assetDir == "" {
 		return errors.New("Missing required flag: --asset-dir")
 	}
--- a/cmd/bootkube/render.go
+++ b/cmd/bootkube/render.go
@@ -99,6 +99,10 @@ func validateRenderOpts(cmd *cobra.Command, args []string) error {
 	if (renderOpts.etcdCAPath != "" || renderOpts.etcdCertificatePath != "" || renderOpts.etcdPrivateKeyPath != "") && (renderOpts.etcdCAPath == "" || renderOpts.etcdCertificatePath == "" || renderOpts.etcdPrivateKeyPath == "") {
 		return errors.New("You must specify either all or none of --etcd-ca-path, --etcd-certificate-path, and --etcd-private-key-path")
 	}
+	if renderOpts.etcdCertificatePath != "" && renderOpts.selfHostedEtcd {
+		return errors.New("Cannot specify --etcd-certificate-path with --experimental-self-hosted-etcd." +
+			" Self-hosted etcd + TLS will auto-generate certs based on root CA cert.")
+	}
 	if renderOpts.assetDir == "" {
 		return errors.New("Missing required flag: --asset-dir")
 	}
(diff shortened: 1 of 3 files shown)


hongchaodeng:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

cmd/bootkube/render.go
```diff
@@ -99,9 +99,6 @@ func validateRenderOpts(cmd *cobra.Command, args []string) error {
 	if (renderOpts.etcdCAPath != "" || renderOpts.etcdCertificatePath != "" || renderOpts.etcdPrivateKeyPath != "") && (renderOpts.etcdCAPath == "" || renderOpts.etcdCertificatePath == "" || renderOpts.etcdPrivateKeyPath == "") {
 		return errors.New("You must specify either all or none of --etcd-ca-path, --etcd-certificate-path, and --etcd-private-key-path")
 	}
-	if renderOpts.etcdCertificatePath != "" && renderOpts.selfHostedEtcd {
```
diegs: Actually it seems like you want to keep this check. AFAICT you are ignoring the etcd certificate flags when using self-hosted and just generating certs from the master CA.

Maybe add some more information to the flag or this error that explains what the behavior is for self-hosted etcd + TLS.

Also need to edit line 214 to not display that warning when using self-hosted etcd.

pkg/asset/tls.go
```diff
@@ -148,14 +148,78 @@ func newEtcdTLSAssets(etcdCACert, etcdClientCert *x509.Certificate, etcdClientKe
 	return assets, nil
 }
 
+func newSelfHostedEtcdTLSAssets(etcdSvcIP, bootEtcdSvcIP string, caCert *x509.Certificate, caPrivKey *rsa.PrivateKey) (Assets, error) {
```
diegs: So, this method does not allow any user specifying of etcd certs, which is fine, but maybe just add a comment or something.

cmd/bootkube/render.go
hongchaodeng: How about add one more line to the error msg:
```
 Self-hosted etcd + TLS will auto-generate certs based on root CA cert.
```

hongchaodeng:
fixed

cmd/bootkube/render.go
```diff
@@ -100,7 +100,8 @@ func validateRenderOpts(cmd *cobra.Command, args []string) error {
 		return errors.New("You must specify either all or none of --etcd-ca-path, --etcd-certificate-path, and --etcd-private-key-path")
 	}
 	if renderOpts.etcdCertificatePath != "" && renderOpts.selfHostedEtcd {
-		return errors.New("Cannot specify --etcd-certificate-path with --experimental-self-hosted-etcd")
+		return errors.New("Cannot specify --etcd-certificate-path with --experimental-self-hosted-etcd." +
+			" Self-hosted etcd + TLS will auto-generate certs based on root CA cert.")
```
diegs: Sorry, one last thing: you need to update the section line 176:
- Pull out the `else` branch to just always run.
- Remove the `http://` override (lines 177-180)
- Remove `defaultSelfHostedEtcdServers` and refactor that check+warning to just use the CIDR and not the full URL w/ scheme (so it's compatible with both `http` and `https` -- what it is really checking is the CIDR)

pkg/asset/tls.go
```diff
@@ -148,14 +148,82 @@ func newEtcdTLSAssets(etcdCACert, etcdClientCert *x509.Certificate, etcdClientKe
 	return assets, nil
 }
 
+// newSelfHostedEtcdTLSAssets automatically generates three suites of x509 certificates (CA, key, cert)
+// for self-hosted etcd related components. Two suites are used by etcd members' client and peer ports;
```
diegs: s/suite/cert?

hongchaodeng: A suite is (CA, key, cert).
Better use suite to differentiate from cert.

cmd/bootkube/render.go
hongchaodeng: Yeah. This part is definitely something I'm gonna fix in next PR.
I want this PR to be more about autogen self-hosted certs.
I already have the entire work done. I'm separating them into smaller PRs so that it eases review work and make it good enough to pass out tests and multi-node working.
Let's take this in the following PR.

pkg/asset/tls.go
diegs: Ah, ok :)

cmd/bootkube/render.go
hongchaodeng: Can we move forward this PR?
I don't think this comment is relevant to this PR.

diegs: Ok to fix in a follow-up, though as the code stands now you cannot render an `https` etcd address with a self-hosted cluster.

hongchaodeng: Will fix it shortly.

> now you cannot render an https etcd address with a self-hosted cluster.

That's fine. As long as it didn't break the CI.

PR #589, opened 2017-06-16
Change `bootkube recover` flag to --recovery-dir.

diegs:
Field ops reports that people find the --asset-dir flag confusing; they
think that means you need to pass in the same --asset-dir that was used
to create the cluster.

cc @coresolve


xiang90:
lgtm

coresolve:
Yay

PR #555, opened 2017-05-27
Use selected API server port in manifest templates

rfratto:

PR #597, opened 2017-06-19
Use selected API server port in manifest templates

diegs:
Just patching in #555 which is a good change but seems abandoned.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: go
 go:
-  - 1.7
+  - 1.8
 script:
   - make clean release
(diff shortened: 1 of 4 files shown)


diegs:
Ah, url.Port requires Go 1.8+

We can hold off on this until Bootkube officially moves to Go 1.8

diegs:
coreosbot run e2e

diegs:
coreosbot run e2e

diegs:
coreosbot run e2e

diegs:
@aaronlevy don't think that affects bootkube anymore since we pulled the apiserver out, and the fix was cherry-picked all the way back to 1.5.

Issue #581, opened 2017-06-14
etcd-operator on aws s3 backups


sstarcher:
Setup:
* bootkube 0.4.4
* aws
* self hosted etcd

Change:
* configuring the etcd-operator to backup to an s3 bucket
```
  backup:
    backupIntervalInSecond: 300
    maxBackups: 5
    storageType: "S3"
    s3:
      s3Bucket: accessible-bucket
```

Reference: https://github.com/coreos/etcd-operator/issues/1196

In the above situation the etcd-operator attempts to create `/etc/etcd-operator` directory, but bootkube is setting the user to a non-root user who does not have access to create that directory.  

The result is the etcd-operator crashes the cluster.

sstarcher:
Location of the code https://github.com/coreos/etcd-operator/blob/859be539426b252765bc2f15b10c1836eef059e9/pkg/cluster/backupstorage/s3.go#L38

sstarcher:
The etcd-operator manifest is running the container as the non-root user so when it attempts to create /etc/etcd-operator it fails.

aaronlevy:
cc @xiang90 @hongchaodeng

sstarcher:
@aaronlevy @hongchaodeng already closed my other ticket against the etcd-operator saying the issue is with bootkube.

xiang90:
Yes. We are going to solve it here

hongchaodeng:
Since there are too many opinions in https://github.com/kubernetes-incubator/bootkube/pull/598, let's close the PR and continue the discussion here until we have a solution.

sstarcher:
This is simple to solve,  have the container create the directory so the code does not have to. The code should run as nobody. This should be fixed on the etcd operator side.

hongchaodeng:
@sstarcher 
Yeah. That sounds good to me.

xiang90:
@hongchaodeng 

@diegs mentioned to me, we also try to access ./etc/.aws. probably also need to figure that out.

xiang90:
@sstarcher 

> The code should run as nobody.

nobody can only access /tmp, no? We cannot use /tmp to store tmp snapshot files as they can be quite large.

@aaronlevy we probably can change the code in the operator to use $HOME. But still, we need to pass in a uid rather than nobody in bootkube. Is it the direction you would like to see?

aaronlevy:
Maybe `/var/tmp`?

[...]

PR #598, opened 2017-06-20
templates: EtcdOperator shouldn't run as nobody

hongchaodeng:
fix https://github.com/kubernetes-incubator/bootkube/issues/581

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -916,9 +916,6 @@ spec:
               fieldPath: metadata.name
       nodeSelector:
         node-role.kubernetes.io/master: ""
-      securityContext:
-        runAsNonRoot: true
-        runAsUser: 65534
       tolerations:
       - key: node-role.kubernetes.io/master
         operator: Exists


xiang90:
lgtm.

aaronlevy:
@hongchaodeng Does etcd operator need to run as root?

hongchaodeng:
@aaronlevy 
Yes.
ref: https://github.com/coreos/etcd-operator/issues/806
Is it a problem? We can try prioritize if there is any issues

aaronlevy:
Ideally nothing is running as the root user. We're trying to convert all existing objects to not require this either. What host-level privileges does etcd-operator need?

hongchaodeng:
@aaronlevy 

No host-level priv is needed. It needs some container level priv to create some dirs, like:
```
mkdir /etc/etcd-operator: permission denied
```

aaronlevy:
So we should probably pick a user to give to etcd-operator (nobody shouldn't own files). But giving operator root sounds unnecessary.

hongchaodeng:
@aaronlevy 
I'm not sure but who else can create dir in `/etc` ?
And why is it a problem? It never escapes the container.

aaronlevy:
/etc on the host or in the container fs?

hongchaodeng:
container fs

Quentin-M:
We also have it run as `nobody` in Tectonic: https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/etcd/tectonic-etcd-operator.yaml#L32-L34. If this is actually broken and there is no way around it, please advise before code freeze.

The image/app should probably be fixed so that it can run as `nobody`. We have been doing that for all the other images (i.e. adding setcap on our ingress controller, and various chown/chmod).

hongchaodeng:
@Quentin-M 

> We also have it run as nobody in Tectonic ...

Why do you want to set specific user? Any concern to use default user which is root?

hongchaodeng:
Note that even if etcd operator image is set to something else, the etcd pod & backup sidecar is still running as root. There needs broader investigation on everything if there is real concern.

PR #604, opened 2017-06-21
hack/terraform-quickstart: remove dependancy on default network

pbx0:
This PR avoids terraform depending on a sane default network existing to
expose the nodes. This includes creating a VPC, subnet, route table,
internet gateway, open security group, and a network ACL.  This will be
useful once tools to auto-wipe a region are used so that it doesn't take
out all our tests that depend on default networks lying around.

Besides changing the variable `instance_tags` to `resource_owner` to tag our resources this should ultimately be creating nodes that are no different from before. As an aside, we should probably conform to the new tagging standard, but since most uses of this terraform are transient anyway its not urgent.

Also, its important we get this in tomorrow or CI will likely break.

cc @aaronlevy

### DIFF
--- a/hack/terraform-quickstart/main.tf
+++ b/hack/terraform-quickstart/main.tf
@@ -10,8 +10,13 @@ resource "aws_instance" "bootstrap_node" {
   key_name             = "${var.ssh_key}"
   iam_instance_profile = "${aws_iam_instance_profile.bk_profile.id}"
 
+  vpc_security_group_ids      = ["${aws_security_group.allow_all.id}"]
+  subnet_id                   = "${aws_subnet.main.id}"
+  associate_public_ip_address = true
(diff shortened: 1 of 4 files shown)


pbx0:
coreosbot run e2e

pbx0:
coreosbot run e2e

pbx0:
coreosbot run conformance

pbx0:
coreosbot run e2e

pbx0:
Took a bit of voodoo to get it all working, but I'm gaining confidence now that it passes conformance. e2e tests aren't going so great, but I'm fairly sure its unrelated and not making the flakes any worse at this point. I've seen the tests pass at least once, just not at the same time. Digging around a little more locally, but i'll likely merge this soon.

PR #606, opened 2017-06-22
e2e: add ability to drop out of test and into SSH shell

pbx0:
The ergonomics could be improved a bit (not all tests have ready access
to the list of hostnames), but that is more in-scope for other
framework-ish cleanups.

Posting this up while I wait for flakes to reproduce locally.

cc @diegs @yifan-gu


yifan-gu:
coreosbot run e2e

PR #607, opened 2017-06-22
templates: EtcdOperator shouldn't run as nobody

hongchaodeng:
fix https://github.com/kubernetes-incubator/bootkube/issues/581

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -916,9 +916,6 @@ spec:
               fieldPath: metadata.name
       nodeSelector:
         node-role.kubernetes.io/master: ""
-      securityContext:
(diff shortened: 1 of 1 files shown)


hongchaodeng:
Actually we need to add a user to etcd operator image anyway, we will probably delete the user setting in bootkube.

hongchaodeng:
We should run etcd operator as root whose security boundary will be limited inside the container. Use case see https://github.com/kubernetes-incubator/bootkube/issues/581#issuecomment-310517175 .

hongchaodeng:
Closing this based on https://github.com/kubernetes-incubator/bootkube/issues/581#issuecomment-310522134

Issue #495, opened 2017-05-05
Figure out how we can re-enable healthchecks on the kube-apiserver


aaronlevy:
We disable healthchecks on the api-server because paired with bootstrapping and checkpointing, they can make the pivot process somewhat flaky due to the apiserver using an flock (and "running", but not actually serving requests).

For example during bootstrap:

- bootstrap api-server starts (acquires lock)
- self-hosted api-server starts (waits on lock)
- kubelet tries to health-check self-hosted api-server, but fails (not listening yet)
- kubelet keeps restarting apiserver over and over
- eventually bootstrap api-server stops (and gives up lock)
- As long as the kubelet doesn't die in this window, it will eventually restart the self-hosted apiserver (but this could be a long window due to backoff).

Or during checkpointing:

- api-server is down
- Start a checkpointed copy of apiserver (and acquire lock)
- eventually self-hosted api-server comes back up (waits on lock)
- kubelet tries health-check on self-hosted apiserver and fails (because it's not listening yet)
- kubelet keeps restarting apiserver container over and over
- The problem here is that the checkpointer must use the /runningcontainers endpoint to determine if there is a local apiserver running (https://github.com/kubernetes-incubator/bootkube/blob/0a9b736d7e99a9156a2e6d7ce084213f9a19032a/cmd/checkpoint/main.go#L34-L36), because the kubelet's /pod endpoint could have stale data. However, the /runningcontainers endpoint is only looking at the container runtime state - which means that it might think that no apiserver is running (if it looks while kubelet is restarting the apiserver - there is no container). It then thinks it needs to restart the checkpointed copy of the api-server, then starts process over again. This leads to it sometimes eventually settling down based on timing, but it is quite flaky. 

We could revisit how we determine local running state. Or maybe revist the lock process (upstream something like the kubelet lock, where you can specify "exit-on-lock-contention" so we can coordinate control between two components). Or we might continue to work around it maybe by doing something like using an `exec` healthcheck that is aware of the lock state (via @chancez).

jamiehannaford:
[...]

PR #616, opened 2017-06-29
Remove apiserver flock.

diegs:
The pivot and checkpointer both appear to work fine without it.

If this works, possibly a partial fix for #495

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -164,8 +164,6 @@ spec:
       - name: kube-apiserver
         image: {{ .Images.Hyperkube }}
         command:
-        - /usr/bin/flock
-        - /var/lock/api-server.lock
         - /hyperkube
         - apiserver
         - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds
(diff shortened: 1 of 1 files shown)


diegs:
@aaronlevy  PTAL, I added a liveness probe to the apiserver as well.

diegs:
cc @ericchiang

In order to make this work we need to enable anonymous auth on the apiserver, because the kubelet does not use authentication to probe (https://github.com/kubernetes/kubernetes/issues/9866)

The default permissions for unauthenticated users seem innocuous, but they are also not small:

https://github.com/kubernetes/kubernetes/blob/v1.7.0/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/testdata/cluster-roles.yaml#L373

ericchiang:
Re: anonymous auth. I think a release notes and plans to turn on audit logging would suffice. We could also remove the default system:user-basic binding.

rphillips:
@diegs can you rebase this PR to master to kick the build bot once more? I tried this change locally on master and everything booted correctly.

diegs:
@rphillips done.

redbaron:
what is the status of it? do we need flock at the end?

diegs:
Let me give this another try now that the checkpointer has a grace period when swapping pods.

diegs:
Going to run this locally a few times to make sure the behavior sane, and also rerun e2e.

diegs:
coreosbot run e2e

diegs:
coreosbot run e2e

PR #618, opened 2017-06-30
Design principles doc

aaronlevy:
/cc @sym3tri @diegs @yifan-gu

### DIFF
--- a/design-principles.md
+++ b/design-principles.md
@@ -5,41 +5,41 @@ There are exceptions to these principles, but these are general guidelines the p
 ## General
 
 - Bootkube should be a single-use tool, which only runs on the first node in a cluster.
-    - An exception is the `recovery` subcommand, discussed below.
-- Bootkube should not be required to run to add new nodes to a cluster.
+    - An exception is the `recover` subcommand, discussed below.
+- Bootkube should not be required to add new nodes to a cluster.
     - For example, adding nodes should not require a `bootkube join` command.
+    - Ideally all that should be required of adding a node is starting the kubelet and providing a valid kubeconfig. All other configuration should be sourced from API objects.
 - Should not require flag or configuration coordination between the `render` and `start` steps.
     - Required flag coordination means certain `render` assets will only work with certain `start` flags, and this is something we should avoid.
     - For example, `bootkube render --self-hosted-etcd` requires no changes when ultimately running `bootkube start`.
 - Avoid adding feature flags as much as possible. This makes testing & stability very difficult to maintain.
-    - Users should be able to modify assets after `render` and before `start`. This allows flexibility for any use-case without bootkube owning the complexity in the codebase.
+    - Users customize their cluster by modifying the output of bootkube render to fit their configuration. Users can generate their own assets for use with bootkube start, subject to following a small number of conventions.
     - Complex rendering needs can be handled by custom rendering tools.
         - For example, the [CoreOS Tectonic Installer](https://github.com/coreos/tectonic-installer) performs its own rendering step, but utilizes `bootkube start` to launch the cluster.
-- Launching compute resources is out of scope. Bootkube merely provides quickstart examples, but should not be perscriptive.
+- Launching compute resources is out of scope. Bootkube merely provides quickstart examples, but should not be prescriptive.
 
 ## Bootkube Render
 
 - Bootkube is not meant to be a fully-featured rendering engine. There are much better tools for this - we shouldn't write yet another.
-- Bootkube render should be considerd a useful starting point, which generates assets utilizing latest version and upstream best-practices.
+- Bootkube render should be considered a useful starting point, which generates assets utilizing latest versions and best-practices.
 - Should render assets for the most recent upstream release. If another version is desired, this can be left to the user to modify in their rendered templates.
 - Should avoid using upstream alpha features, and instead allow a user to enable these on a case-by-case basis.
-- Adding new configuration flags should be avoided!
+- Adding new configuration flags should be avoided. Instead users can customize the output and/or use an external rendering tool.
 
 ## Bootkube Start
 
-- Should be able to launch a cluster with only specifying an `--asset-dir`.
+- Should be able to launch a cluster by only specifying an `--asset-dir`.
 - Should be agnostic to the version of kubernetes cluster that is being launched.
 - Should strive toward idempotent bootstrap operation.
     - Although, this is not currently the case when bootstrapping with self-hosted etcd.
 
 ## Bootkube Recovery
 
-- When recovering with a functional API server, only a `kubeconfig` should be required.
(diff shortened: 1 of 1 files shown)


design-principles.md
```diff
@@ -0,0 +1,45 @@
+# Design Principles
+
+There are exceptions to these principles, but these are general guidelines the project strives to adhere to.
+
+## General
+
+- Bootkube should be a single-use tool, which only runs on the first node in a cluster.
+    - An exception is the `recovery` subcommand, discussed below.
```
diegs: nit: `recover`

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+- Bootkube should not be required to run to add new nodes to a cluster.
```
diegs: nit: s/to run//

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+    - For example, adding nodes should not require a `bootkube join` command.
```
diegs: Perhaps add a short sentence on how to do it instead and/or why there is no `join` command?

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+- Should not require flag or configuration coordination between the `render` and `start` steps.
+    - Required flag coordination means certain `render` assets will only work with certain `start` flags, and this is something we should avoid.
+    - For example, `bootkube render --self-hosted-etcd` requires no changes when ultimately running `bootkube start`.
+- Avoid adding feature flags as much as possible. This makes testing & stability very difficult to maintain.
+    - Users should be able to modify assets after `render` and before `start`. This allows flexibility for any use-case without bootkube owning the complexity in the codebase.
```
diegs: Maybe expand on this and make it more of an active voice? e.g. "Users customize their cluster by modifying the output of `bootkube render` to fit their configuration. Users can generate their own assets for use with `bootkube start`, subject to following a small number of conventions."

I think this is one of the strongest advantages of bootkube. I like the "complex rendering needs" part next, but as written right now it sounds a little scary.

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+    - Complex rendering needs can be handled by custom rendering tools.
+        - For example, the [CoreOS Tectonic Installer](https://github.com/coreos/tectonic-installer) performs its own rendering step, but utilizes `bootkube start` to launch the cluster.
+- Launching compute resources is out of scope. Bootkube merely provides quickstart examples, but should not be perscriptive.
```
diegs: nit: prescriptive

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Bootkube Render
+
+- Bootkube is not meant to be a fully-featured rendering engine. There are much better tools for this - we shouldn't write yet another.
+- Bootkube render should be considerd a useful starting point, which generates assets utilizing latest version and upstream best-practices.
```
diegs: nit: considered
nit: latest upstream versions and best practices?

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+- Should render assets for the most recent upstream release. If another version is desired, this can be left to the user to modify in their rendered templates.
+- Should avoid using upstream alpha features, and instead allow a user to enable these on a case-by-case basis.
+- Adding new configuration flags should be avoided!
```
diegs: Maybe add rationale (users can customize the output and/or write their own renderer)

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Bootkube Start
+
+- Should be able to launch a cluster with only specifying an `--asset-dir`.
```
diegs: nit: s/with/by

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+- Should be agnostic to the version of kubernetes cluster that is being launched.
+- Should strive toward idempotent bootstrap operation.
+    - Although, this is not currently the case when bootstrapping with self-hosted etcd.
+
+## Bootkube Recovery
+
+- When recovering with a functional API server, only a `kubeconfig` should be required.
+- When recoverying from an etcd backup, no external assets should be required (all necessary data should be extracted from backup).
```
diegs: This is not currently true; recovering from self-hosted and external etcd both require a `kubeconfig` since `bootkube start` needs one (is there a way to extract that from etcd?).

design-principles.md
```diff
@@ -0,0 +1,45 @@
...
+- The latest version of the recovery tool should strive to be able to recover all previously installed versions (no coupling between recovery process, and time of installation).
+
+## Hack Directory
+
+- Provides simple options for local development on bootkube
+- Provide simple / non-production examples for some providers.
```
diegs: nit: cloud providers

design-principles.md
aaronlevy: Used your wording verbatim

design-principles.md
aaronlevy: Added a little blurb to that effect

design-principles.md
aaronlevy: Removed that wording (and updated the line above to be more generic). We could extract the info we needed to generate our own kubeconfig. It could use one of the service accounts (contains CA+signed bearer token) or maybe we could extract another components tls info. But the service account seems reasonable for a temp kubeconfig.

Opened an issue to track the possibility: https://github.com/kubernetes-incubator/bootkube/issues/623

design-principles.md
aaronlevy: Added a bit more wording here (ptal)

aaronlevy:
@diegs thanks for review (and spellcheck). straight-forward stuff I just fixed. But then added some additional wording in places if you don't mind taking another pass.

design-principles.md
```diff
@@ -0,0 +1,46 @@
...
+    - An exception is the `recover` subcommand, discussed below.
+- Bootkube should not be required to add new nodes to an existing cluster.
+    - For example, adding nodes should not require a `bootkube join` command.
+    - Ideally all that should be required to add a node is starting the kubelet and providing a valid kubeconfig.
+    - Configuration beyond the initial kubeconfig should be sourced from API objects. For example, the pod network is configured via CNI daemonset.
+- Should not require flag or configuration coordination between the `render` and `start` steps.
+    - Required flag coordination means certain `render` assets will only work with certain `start` flags, and this is something we should avoid.
+    - For example, `bootkube render --self-hosted-etcd` requires no changes when ultimately running `bootkube start`.
+- Avoid adding feature flags as much as possible. This makes testing & stability very difficult to maintain.
+    - Users customize their cluster by modifying the output of bootkube render to fit their configuration. Users can generate their own assets for use with bootkube start, subject to following a small number of conventions.
```
diegs: nit: `bootkube start` (formatting)

Issue #599, opened 2017-06-21
self hosted etcd has a circular dependency on service ip.


coresolve:
With self hosted etcd we see that significant memory pressure on the masters will cause the cluster to become unstable.

This will mean that the kubelet will eventually mark all the etcd pods unhealthy and thus the kube-proxy will rejecting packets headed for the `kube-system/etcd-service:client`

At this point the apiserver is unable to start as it is configured to connect to etcd cluster with the service ip address of the self hosted cluster. The service continues to have no health endpoints because the kubelet can't talk to the apiserver and tell it that the service has healthy endpoints. 

So now we have a healthy etcd cluster and no way to update the record in etcd via the apiserver.

chancez:
https://github.com/coreos/etcd-operator/issues/622 seems related.

hongchaodeng:
Tolerate unreadiness is not enough. We also need to take into account restarts.

hongchaodeng:
From this comment: https://github.com/kubernetes/kubernetes/issues/25283#issuecomment-264099193

Seems the tolerate-unready feature might be able to tolerate restart (rolling upgrade). We need to test out though

coresolve:
Further when you restart a node in this condition we aren't checkpointing the sh etcd cluster. So there is no way to get the cluster back in a reboot.

hongchaodeng:
ref: https://github.com/kubernetes/kubernetes/issues/47880

hongchaodeng:
We are going to enable alpha annotation `TolerateUnreadyEndpointsAnnotation` in etcd service. Will submit a PR shortly.

PR #626, opened 2017-07-05
etcd_svc: tolerate unready endpoints

hongchaodeng:
fix https://github.com/kubernetes-incubator/bootkube/issues/599

### DIFF
--- a/Documentation/alpha-features.md
+++ b/Documentation/alpha-features.md
@@ -1,0 +1,19 @@
+# Alpha Features
+
+Tracking document for alpha features that bootkube installed clusters may make use of.
+
+We track these alpha features as their behavior may change or be deprecated between Kubernetes versions.
+Therefore clusters that use these features need to keep track of any potential changes in upstream releases.
+See the upstream [api versioning documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#alpha-beta-and-stable-versions) for more information.
+
+
+### TolerateUnreadyEndpointsAnnotation
+
+Used by the etcd service object when self-hosted etcd cluster is enabled.
+
+This alpha annotation will retain the endpoints even if the etcd pod isn't ready.
+This feature is always enabled in endpoint controller in k8s even it is alpha.
+
+References:
+- https://github.com/kubernetes-incubator/bootkube/issues/599
+- https://github.com/kubernetes-incubator/bootkube/pull/626#issuecomment-313187659
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -930,6 +930,10 @@ kind: Service
 metadata:
   name: {{ .EtcdServiceName }}
   namespace: kube-system
+  # This alpha annotation will retain the endpoints even if the etcd pod isn't ready.
+  # This feature is always enabled in endpoint controller in k8s even it is alpha.
+  annotations:
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
 spec:
   selector:
     app: etcd


hongchaodeng:
@xiang90 @aaronlevy @diegs

xiang90:
looks good to me.

@aaronlevy where is the best place to track the alpha feature we use?

diegs:
IIUC We should not be using alpha features by default in bootkube.

fxposter:
@diegs But this annotation doesn't break anything, isn't it? It is not possible to achieve this behavior for previous versions of kubernetes, so you lose nothing.

hongchaodeng:
@diegs 
So this has some offline discussion with @aaronlevy and @xiang90 . Let me share it on github too.

First of all, we have seen many problems with unready (or not "Running") endpoints in service:
- https://github.com/kubernetes-incubator/bootkube/issues/599
- https://github.com/coreos/tectonic-installer/issues/1266

We would need this feature from k8s to make the whole thing more robust. Otherwise, it's very fragile and only work under **good assumptions**.

Second, this is not something new in k8s world. There has been long-discussed issues (https://github.com/kubernetes/kubernetes/issues/39363, https://github.com/kubernetes/kubernetes/issues/25283) for how to handle stateful workloads on k8s. This is something that upstream is moving towards (even slowly) for stateful workloads. Many statefulset applications are using it.

Third, as mentioned above, this doesn't break anything. Further more, self-hosted-etcd is also alpha feature.

diegs:
Thanks for the additional context. It also looks like despite being an alpha feature it's enabled all the time (https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/endpoint/endpoints_controller.go) so it seems ok.

Can you add a brief YAML comment to the template saying a little bit about why we need this alpha feature (and also maybe mention that it has been around a long time & is always enabled)? That way people who are deriving their own configs from the templates will understand why it's helpful.

diegs:
Though it looks like the migration to self-hosted etcd failed in the e2e test:

```
Migrating to self-hosted etcd cluster...
I0705 17:59:32.627115     997 log.go:19] warning: ignoring ServerName for user-provided CA for backwards compatibility is deprecated
I0705 17:59:42.630782     997 migrate.go:65] created etcd cluster TPR
Error: services "etcd-service" not found
Tearing down temporary bootstrap control plane...
Error: services "etcd-service" not found
Error: services "etcd-service" not found
services "etcd-service" not found
```

aaronlevy:
@xiang90 maybe start a `Documentation/alpha-features.md` document that tracks what alpha features are being used, and why.

Ideally we aren't using alpha features -- but we had discussed the risk in enabling this and it seemed low enough to at least try and exercise it with self-hosted etcd. 

- the feature is already enabled by default in the controller-manager (just opt-in with annotation)
- it's used already by / relied upon by statefulsets
- we still consider self-hosted etcd to be in alpha itself.

My main concern would be people upgrading their clusters in the future and this feature essentially being graduated to an actual field, which means the annotation would just silently stop working unless they knew to update to the new fields. Generally covered by release notes, but just something to consider (and why I'd like to have an alpha feature tracking list).

aaronlevy:
Also agree with @diegs that a brief note directly in the YAML would be helpful too

xiang90:
> @xiang90 maybe start a Documentation/alpha-features.md document that tracks what alpha features are being used, and why.

sounds good.

hongchaodeng:
comments added

hongchaodeng:
> maybe start a Documentation/alpha-features.md document

We will write this thing in etcd-operator repo.
@aaronlevy Can you write that in bootkube repo?

Documentation/alpha-features.md
```diff
@@ -0,0 +1,6 @@
+# Alpha Features
+
```
aaronlevy: Add a quick description of this document. Something like:

> Tracking document for alpha features that bootkube installed clusters may make use of.
>
> We track these alpha features as their behavior may change or be deprecated between Kubernetes versions. Therefore clusters that use these features need to keep track of any potential changes in upstream releases. See the upstream [api versioning documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#alpha-beta-and-stable-versions) for more information.

Documentation/alpha-features.md
```diff
@@ -0,0 +1,6 @@
...
+### TolerateUnreadyEndpointsAnnotation
+
```
aaronlevy: Add a note about what component uses this. Something like:

> Used by the etcd service object when self-hosted etcd cluster is enabled.

Documentation/alpha-features.md
```diff
@@ -0,0 +1,6 @@
...
+This alpha annotation will retain the endpoints even if the etcd pod isn't ready.
```
aaronlevy: Can you also add a link to issue that discusses why this is needed?

hongchaodeng:
@aaronlevy 
fixed

hongchaodeng:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

xiang90:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

xiang90:
coreosbot run e2e

xiang90:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

hongchaodeng:
coreosbot run e2e

diegs:
@hongchaodeng rebasing may help with the tests.

hongchaodeng:
pass after rebase

PR #630, opened 2017-07-06
e2e: flake / debug improvements

aaronlevy:
Previously if the wget-pod, for example, was never running all we saw was the phase "Failed" -- but no reason why it failed. We should output a bit more info to help debugging flaky PR tests.

### DIFF
--- a/e2e/network_test.go
+++ b/e2e/network_test.go
@@ -264,7 +264,7 @@ var wgetPodNT = &v1.Pod{
 			{
 				Name:    "wget-container",
 				Image:   "busybox:1.26",
-				Command: []string{"wget", "--timeout", "5", "--retry-connrefused", "--waitretry=1", "--tries", "5", "nginx-service-nt"},
+				Command: []string{"wget", "--timeout", "5", "--retry-connrefused", "--waitretry", "1", "--tries", "5", "nginx-service-nt"},
 			},
 		},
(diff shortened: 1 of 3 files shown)


e2e/network_test.go
```diff
@@ -263,7 +264,7 @@ var wgetPodNT = &v1.Pod{
 			{
 				Name:    "wget-container",
 				Image:   "busybox:1.26",
-				Command: []string{"wget", "--timeout", "5", "nginx-service-nt"},
+				Command: []string{"wget", "--timeout", "5", "--retry-connrefused", "--waitretry", "1", "--tries", "5", "nginx-service-nt"},
```
abhinavdahiya: @aaronlevy looks like busybox:1.26 wget doesn't support `--retry-connrefused flag`
https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e/238/artifact/e2e/logs/container.fluentd-worker-31t1q.wget-pod-nt-jbb5s_bootkube-e2e-testing_wget-container.20170706.log

aaronlevy: Unfortunate. We need to re-think this logic somehow -- retrying on checking the pod status doesn't make a lot of sense (if it fails once it's not going to change the status).

abhinavdahiya: Let me think if I can add some features to ktestutil to handle this nginx deployment/pings

PR #635, opened 2017-07-07
e2e: use m4.large instances.

diegs:


xiang90:
why m4.large is required?

diegs:
@xiang90 I was told that Tectonic tests use m4.large. cc @abhinavdahiya

I'm seeing if this helps alleviate unhealthy self-hosted etcd clusters & failed scale tests. See: https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-selfetcd/366/console

Ctrl-f for `etcdscale_test`, it failed to scale up to 3. I looked at pod logs and it looks like only 2 came up & first goes unhealthy, etcd-operator doesn't show anything useful in logs.

Does etcd-team have any thoughts on this?

e2e/deleteapi_test.go
```diff
@@ -34,10 +34,9 @@ func TestDeleteAPI(t *testing.T) {
 	// apiserver to go down the next step will return sucess before the
 	// apiserver is ever destroyed.
 	waitDestroy := func() error {
-		// only checking api being down , specific function not important
+		// only checking api being down, specific function not important
 		_, err := client.Discovery().ServerVersion()
-
-		if err == nil {
+		if err != nil {
```
pbx0: `err == nil` is what we want here since, we are specifically looping until we can no longer contact the apiserver as evidence that it was successfully destroyed. Maybe there is a more clear way to write it so it doesn't look like a typo though.

diegs: Thanks @pbx0, I noticed that later while re-reading the test.

Issue #605, opened 2017-06-22
CI matrix is larger now that we have Calico option


pbx0:
Currently our CI runs a set of e2e tests with 2 cluster configurations: with self-hosted etcd and without. We have a new Calico flag that has been added. So technically we should have 4 CI jobs kick off on each PR for each cluster configuration. I propose that maybe we just do 3 since maybe its not so important that calico is tested both with self-hosted etcd and without. So we could have:
```
SelfHostEtcd:true Calico: false
SelfHostEtcd:true Calico: true
SelfHostEtcd: false Calico: false
```
The Calico e2e test is going to have to make it it skips when calico is not used. One reason in favor of not doing the entire matrix is that if you have any probability of flakiness at all, running more of the same tests increases your likely hood of hitting the flake.

pbx0:
I believe it got set to on by default to make the new network policy tests pass. I should have made this issue a stopper for reviewing/merging looking back. So at minimum we need the network policy tests to self-skip in absence of Calico and then turn calico back to false by default or set explicitly in CI. Then we could add a third test later to keep those tests getting exercised. Maybe gate it under `coreosbot run network tests`

aaronlevy:
Closing based on https://github.com/kubernetes-incubator/bootkube/pull/639 - which disables test if calico is not running

PR #636, opened 2017-07-07
skip network tests when kube-calico ds doesn't exist

abhinavdahiya:
https://github.com/kubernetes-incubator/bootkube/issues/605#issuecomment-310480674

### DIFF
--- a/e2e/network_test.go
+++ b/e2e/network_test.go
@@ -18,8 +18,11 @@ import (
 func TestNetwork(t *testing.T) {
 	// check if kube-calico daemonset exists
 	// if absent skip this test
-	if _, err := client.ExtensionsV1beta1().DaemonSets("kube-system").Get("kube-calico", metav1.GetOptions{}); err != nil && apierrors.IsNotFound(err) {
-		t.Skip("skipping as kube-calico daemonset is not installed")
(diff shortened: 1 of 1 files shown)


e2e/network_test.go
```diff
@@ -15,6 +16,12 @@ import (
 )
 
 func TestNetwork(t *testing.T) {
+	// check if kube-calico daemonset exists
+	// if absent skip this test
+	if _, err := client.ExtensionsV1beta1().DaemonSets("kube-system").Get("kube-calico", metav1.GetOptions{}); err != nil && apierrors.IsNotFound(err) {
```
aaronlevy: I'm okay with this if we think of the name being stable. If not, is there a label query we could rely on instead?

aaronlevy: I guess name is up to us, so probably a non-issue

aaronlevy:
Post-lgtm nit: seems a little weird to ignore the error we return (or should add comment or don't assign it at all).

aaronlevy:
post-lgtm-lgtm

abhinavdahiya:
Fixed in https://github.com/kubernetes-incubator/bootkube/pull/639

PR #640, opened 2017-07-11
hyperkube: bump to v1.6.7_coreos.0

diegs:


diegs:
coreosbot run e2e

diegs:
Bumping as part of #642 instead.

PR #644, opened 2017-07-12
Readme cleanup

aaronlevy:
Some minor cleanup then also moving the development docs into their own section


Documentation/development.md
```diff
@@ -0,0 +1,54 @@
...
+make clean all
+```
+
+## Local Development Environments
+
+To easily launch local vagrant development clusters:
+
+```
+# Launch a single-node cluster
+make run-single`
```
yifan-gu: Does this also clean up the `/cluster` dir?

aaronlevy: Yes

PR #652, opened 2017-07-17
pod-checkpointer: remove isRunning check.

diegs:
This could lead to a race where the pod is not running and therefore
the checkpoint isn't created, but the lack of checkpoint leaves us
vulnerable to a service disruption.

Theoretically the apiserver reconciliation loop should handle cleaning
the checkpoint up.


diegs:

diegs:

diegs:

diegs:

PR #655, opened 2017-07-18
templates: enable etcd-quorum-read in apiserver

hongchaodeng:
For HA setup, enabling quorum-read is necessary.

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -174,6 +174,7 @@ spec:
         - --etcd-certfile=/etc/kubernetes/secrets/etcd-client.crt
         - --etcd-keyfile=/etc/kubernetes/secrets/etcd-client.key
 {{- end }}
+        - --etcd-quorum-read=true
         - --etcd-servers={{ range $i, $e := .EtcdServers }}{{ if $i }},{{end}}{{ $e }}{{end}}
         - --insecure-port=0
         - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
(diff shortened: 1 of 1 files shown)


aaronlevy:
@hongchaodeng can you also set it on the bootstrap apiserver?

aaronlevy:
Then as a follow-up this should be done in tectonic-installer too: https://github.com/coreos/tectonic-installer/blob/master/modules/bootkube/resources/bootstrap-manifests/bootstrap-apiserver.yaml

hongchaodeng:
added to bootstrap apiserver

pkg/asset/internal/templates.go
```diff
@@ -175,6 +175,7 @@ spec:
         - --etcd-keyfile=/etc/kubernetes/secrets/etcd-client.key
 {{- end }}
         - --etcd-servers={{ range $i, $e := .EtcdServers }}{{ if $i }},{{end}}{{ $e }}{{end}}
+        - --etcd-quorum-read=true
```
diegs: nit: quorum comes before servers

pkg/asset/internal/templates.go
```diff
@@ -251,6 +252,7 @@ spec:
     - --etcd-keyfile=/etc/kubernetes/secrets/etcd-client.key
 {{- end }}
     - --etcd-servers={{ range $i, $e := .EtcdServers }}{{ if $i }},{{end}}{{ $e }}{{end}}{{ if .SelfHostedEtcd }},https://127.0.0.1:12379{{end}}
+    - --etcd-quorum-read=true
```
diegs: and here.

hongchaodeng:
fixed

diegs:
Merging because the tests passed last time and this failure is unrelated.

PR #657, opened 2017-07-20
kenc: bump to 0.0.2

diegs:
cc @xiang90


dghubble:
`v` vs no `v`, is this being read by semver somewhere and leaving it off is required?

diegs:
That's a good question for @hasbro17, who made the image.

hasbro17:
@dghubble no there's no requirement. I can easily update the image tag with the `v` prefix if that's the required convention here. Let me know if that's preferred.

Since I wasn't too familiar with the "to v or not to v" question, I (kind of arbitrarily) kept the image tag the same as the release tag on GitHub https://github.com/coreos/kenc/releases/tag/0.0.2

PR #663, opened 2017-07-27
Add TLS auto-approval for kubelet.

diegs:


pkg/asset/internal/templates.go
```diff
@@ -11,14 +29,42 @@ clusters:
...
+
+var CSRApproverRoleBindingTemplate = []byte(`kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: approve-node-client-csr
+rules:
+- apiGroups: ["certificates.k8s.io"]
+  resources: ["certificatesigningrequests/nodeclient"]
+  verbs: ["create"]
+- apiGroups: ["certificates.k8s.io"]
```
ericchiang: note for later:

There's actually a `system:node-bootstrapper` role we can just bind to https://github.com/kubernetes/kubernetes/blob/3a0d8f8fea75140e4a38f5165a75e93bcd28c60c/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/testdata/cluster-roles.yaml#L724-L742

pkg/tlsutil/tlsutil.go
```diff
@@ -127,3 +128,21 @@ func NewSignedCertificate(cfg CertConfig, key *rsa.PrivateKey, caCert *x509.Cert
 	}
 	return x509.ParseCertificate(certDERBytes)
 }
+
+// NewRandomToken generates an n-byte random hexadecimal token. The output will have length n*2.
+func NewRandomToken(n int) ([]byte, error) {
```
ericchiang: nit: there's actually a defined format for another authenticator. We might just want to copy that for now: https://kubernetes.io/docs/admin/bootstrap-tokens/#token-format

pkg/asset/tls.go
```diff
@@ -91,20 +91,19 @@ func newAPIKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey, altNa
...
 }
 
-func newKubeletKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey) (*rsa.PrivateKey, *x509.Certificate, error) {
+func newAdminKeyAndCert(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey) (*rsa.PrivateKey, *x509.Certificate, error) {
 	// TLS organizations map to Kubernetes groups, and "system:masters"
 	// is a well-known Kubernetes group that gives a user admin power.
 	//
-	// For now, put the kubelets in this group. Later we can restrict
-	// their credentials, likely with the help of TLS bootstrapping.
+	// For now, put the admins in this group.
```
ericchiang: We can remove this comment. We're not going to change this group ever.

pkg/asset/internal/templates.go
diegs: Done

pkg/asset/tls.go
diegs: Done

pkg/tlsutil/tlsutil.go
diegs: Done

ericchiang:
Seeing some issues with only using the RotateKubeletClientCertificate to perform the first CSR (and not the bootstrap flag). Reported issue is that it requires a kubelet restart after doing the CSR to start authenticating as a node.

I think the kubelet is opening a connection to the API server with the bootstrapping credentials to perform its CSR, then keeping that connection alive even though it's got a client cert. The kubelet never closes its existing connection to the API server, so it never re-performs a TLS handshake to authenticate using its client cert.

Only workaround I see is to have systemd restart the kubelet once it detects the kubelet's written its first certificate to disk. Blah.

jcbsmpsn:
@ericchiang Yes, I can see how that would happen. The underlying connection is only re-established if it is disrupted for some reason. The rotated cert is set immediately, and will be used when the connection is re-established.

liggitt:
> I think the kubelet is opening a connection to the API server with the bootstrapping credentials to perform its CSR, then keeping that connection alive even though it's got a client cert.

That's concerning... the client and transport built from the bootstrapping credentials shouldn't be reused beyond the bootstrap phase. Once it gets the first client cert, the input to building the API client should be different, so it should get a different tls config, with a new set of keepalive connections.

liggitt:
oh, by bootstrapping, you don't actually mean `--bootstrap`... you mean the initial content of `--kubeconfig`

ericchiang:
@jcbsmpsn because the API server authenticates x509 certs at the time of the request, and not on the handshake[1] I think we'll hit the same issue when the kubelet cert expires. E.g. once the kubelet's cert expires, it'll keep the connection open, but the requests will start getting forbidden errors.

I think the kubelet's rotation code should shut down its connections once it successfully rotates a cert. That would let it immediately start using the new certs and prevent these kind of issues with long lived connections and old credentials.

[1] https://github.com/kubernetes/kubernetes/blob/v1.7.2/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go

jcbsmpsn:
@ericchiang That would be good. I do not think that will be easy.

jcbsmpsn:
https://github.com/kubernetes/kubernetes/pull/49899 addresses closing existing connections when the certificate changes.

diegs:
@jcbsmpsn yes, I added a temporary commit that switches the kubelet to use a hyperkube with that PR to test.

ericchiang:
@diegs this PR has been updated with 1.8 features. It also now enables node authorizer and admission controller.

https://kubernetes.io/docs/admin/authorization/node/
https://kubernetes.io/docs/admin/admission-controllers/#noderestriction

Everything works now except the checkpointer. I think we do want to run the checkpointer with a service account and ideally it's own RBAC role. Since it's running as a static pod, and wont have access to `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT`, we probably want to do something similar the `kube-proxy-kubeconfig` in this PR.

cc @dghubble @colemickens --- in case either of you want to take a look.

pkg/asset/internal/templates.go
```diff
@@ -56,7 +112,7 @@ spec:
       - name: kubelet
         image: {{ .Images.Hyperkube }}
         command:
-        - ./hyperkube
+        - /hyperkube
```
ericchiang: woops, need to remove this. think its a rebase artifact.

hack/multi-node/user-data.sample
```diff
@@ -19,15 +19,18 @@ coreos:
           --mount volume=opt-cni-bin,target=/opt/cni/bin \
           --insecure-options=image"
         ExecStartPre=/bin/mkdir -p /opt/cni/bin
-        ExecStartPre=/bin/mkdir -p /etc/kubernetes/manifests
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/cni/net.d
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/checkpoint-secrets
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/inactive-manifests
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes/kubelet/tls
```
rothgar: Default directory for kubeadm is /etc/kubernetes/pki. Does bootkube care to be the same? Just might help people who use both systems but not a real concern for functionality.

diegs: @rothgar I think that's a good suggestion, but `tls` is used in a few places in bootkube. Mind opening an issue to suggest aligning with kubeadm? We can address in a separate PR.

hack/quickstart/kubelet.master
```diff
@@ -32,7 +34,9 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --network-plugin=cni \
   --node-labels=node-role.kubernetes.io/master \
   --pod-manifest-path=/etc/kubernetes/manifests \
-  --register-with-taints=node-role.kubernetes.io/master=:NoSchedule
+  --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
+  --require-kubeconfig \
```
aaronlevy: I think this is deprecated

colemickens: yup https://github.com/kubernetes/kubernetes/commit/2189314895a28e37f30781c721bddd9228001c29

ericchiang: Thanks, will remove when I rebase for https://github.com/kubernetes-incubator/bootkube/pull/767

ericchiang:
coreosbot run e2e

ericchiang:
coreosbot run e2e

diegs:
@ericchiang you're going to need to release a new and bump the version in images.go to pick it up. @dghubble or @aaronlevy might be able to help you find the Jenkins job.

redbaron:
Checkpointer is running in a dedicated service account not, is there anything left [I can help with], which stops this PR from being merged?

diegs:
@redbaron I am wrapping up #784 which makes the tests better and changes the checkpointer to only look at its own namespace, which relaxes the required roll bindings. Hoping to finish that up in the next day or two.

ericchiang:
> Checkpointer is running in a dedicated service account now, is there anything left [I can help with], which stops this PR from being merged?

This is actually already done, even without @diegs changes. The changes were manifest only and didn't require any changes to the checkpointer itself.

I think we still need to add an approve all rule for rotation[1 (different from bootstrap approval) as well as docs. Will update this when I get the chance.

[1] https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/#kube-controller-manager-configuration

hack/single-node/user-data.sample
```diff
@@ -38,7 +42,8 @@ coreos:
           --lock-file=/var/run/lock/kubelet.lock \
           --network-plugin=cni \
           --node-labels=node-role.kubernetes.io/master \
-          --pod-manifest-path=/etc/kubernetes/manifests
+          --pod-manifest-path=/etc/kubernetes/manifests \
+          --rotate-certificates
```
redbaron: shouldn't it be `--bootstrap-kubeconfig` and `--require-kubeconfig` as per https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/#kubelet-configuration ?

ericchiang: This is actually intentional. We want to start static pods even if the rotation hasn't completed. `--bootstrap-kubeconfig` will block, but `--rotate-certificates` wont, even though they both have the same effect.

ericchiang:
Still waiting for e2e. If this passes then I think the code bits are done.

We still need to do some rotation testing (e.g. if we run a long lived cluster the rotation works) and add some docs.

aaronlevy:
@ericchiang any updates on this?

ericchiang:
No I haven't had time to do the short rotation time period testing.

ericchiang:
Found a bug in the rotation code. It's fixed now.

Tested rotation with the following diff. We probably want to do something similar during the e2e tests. Maybe as a follow up?

```diff
--- a/pkg/asset/internal/templates.go               
+++ b/pkg/asset/internal/templates.go               
@@ -503,6 +503,7 @@ spec:                           
         - --leader-elect=true                      
         - --root-ca-file=/etc/kubernetes/secrets/ca.crt                                                 
         - --service-account-private-key-file=/etc/kubernetes/secrets/service-account.key                
+        - --experimental-cluster-signing-duration=15m                                                   
         livenessProbe:                             
           httpGet:                                 
             path: /healthz                         
@@ -536,7 +537,6 @@ spec:                           
       dnsPolicy: Default # Don't use cluster DNS.  
 `)                                                 
                                                    
-                                                   
 var ControllerManagerServiceAccount = []byte(`apiVersion: v1                                            
 kind: ServiceAccount                               
 metadata:                                          
@@ -581,6 +581,7 @@ spec:                           
     - --leader-elect=true                          
     - --root-ca-file=/etc/kubernetes/secrets/ca.crt                                                     
     - --service-account-private-key-file=/etc/kubernetes/secrets/service-account.key                    
+    - --experimental-cluster-signing-duration=15m  
     volumeMounts:                                  
     - name: secrets                                
       mountPath: /etc/kubernetes/secrets
```

diegs:
@ericchiang sure, that sounds good, except does that require waiting 15m for the certs to rotate?

ericchiang:
@diegs yes. Maybe during conformance instead?

diegs:
@ericchiang SG. also code LGTM but I can't LGTM "my own" PR.

ericchiang:
coreosbot run e2e

diegs:
Needs another approval from a member, cc @rphillips @aaronlevy

pkg/asset/k8s.go
```diff
@@ -105,40 +109,67 @@ func newDynamicAssets(conf Config) Assets {
...
+// newBootstrapToken constructs a bootstrap token in conformance with the following format:
+// https://kubernetes.io/docs/admin/bootstrap-tokens/#token-format
+func newBootstrapToken() (id string, secret string, err error) {
+	token := make([]byte, 6+16)
+	if _, err := rand.Read(token); err != nil {
+		return "", "", err
+	}
+
+	for i, b := range token {
+		token[i] = validBootstrapTokenChars[int(uint8(b))%len(validBootstrapTokenChars)]
```
aaronlevy: nit: don't think the uint() necessary

pkg/asset/k8s.go
```diff
@@ -105,40 +109,67 @@ func newDynamicAssets(conf Config) Assets {
 	return assets
 }
 
+const validBootstrapTokenChars = "0123456789abcdefghijklmnopqrstuvwxyz"
+
+// newBootstrapToken constructs a bootstrap token in conformance with the following format:
+// https://kubernetes.io/docs/admin/bootstrap-tokens/#token-format
+func newBootstrapToken() (id string, secret string, err error) {
+	token := make([]byte, 6+16)
```
aaronlevy: nit: not immediately clear what 6+16 is (id+secret).

diegs: Added comment.

pkg/asset/k8s.go
diegs: Removed.

PR #668, opened 2017-08-02
etcd-operator: 0.5.0

diegs:


klausenbusk:
Just a question, how would I upgrade etcd-operator in a existing cluster with selfhosted etcd? Reading the [upgrade doc](https://github.com/coreos/etcd-operator/blob/master/doc/user/upgrade/upgrade_guide.md#v04x---v05x), I get the feeling that it can't be done without restarting the cluster?

PR #672, opened 2017-08-03
checkpointer: handle kubelet API denials and make kubeconfig configurable

### LINKED ISSUE #625: Allow pod checkpointer to use different kubeconfig paths
kubeadm doesn't save a standard `/etc/kubernetes/kubeconfig` file, but rather `/etc/kubernetes/admin.conf` and a series of specific component files. The problem is that the pod checkpointer expects the standard and already mounts `/etc/kubernetes` so you can't alias any of these files.

Also, will the checkpointer code ever live in its own repo?

### LINKED ISSUE #648: Enable kubelet authorization
https://kubernetes.io/docs/admin/kubelet-authentication-authorization/

Kubelet authorization lets us identify which components in the system are allowed to talk to the kubelet API. This allows more fine grained access to the kubelet API beyond white listing client certs signed by a specific CA. It would, for instance, let prometheus talk to specific kubelet API paths with only its service account token.

Authorization policies are managed by RBAC on the APi server, with the kubelet performing authorization checks using SubjectAccessReviews. Because of this, _if the API server is down all requests to the kubelet API are denied._

As discussed with @diegs, the checkpointer will need to be updated to deal with this case. Our proposal is that if the checkpointer can't contact the kubelet, it should assume that no pods are running, and try to start them via static manifests. This can be paired with other features, such as limits on how long after a pod has last run that the checkpointer should try to run it.

ericchiang:
This is a requirement for kubelet authorization.

We still force the checkpointer to not check the kubelet's serving cert no matter what's in the kubeconfig, since in our deployments it's self signed.

Updates #648
Closes #625

cc @diegs

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -2,12 +2,9 @@ package main
 
 import (
 	"bytes"
-	"crypto/tls"
-	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
-	"net/http"
 	"os"
 	"path"
 	"path/filepath"
@@ -20,9 +17,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/pkg/api"
 	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
@@ -31,17 +32,10 @@ const (
 	podNameEnv      = "POD_NAME"
 	podNamespaceEnv = "POD_NAMESPACE"
 
-	// We must use both the :10255/pods and :10250/runningpods/ endpoints, because /pods endpoint could have stale data.
-	// The /pods endpoint will only show the last cached status which has successfully been written to an apiserver.
-	// However, if there is no apiserver, we may get stale state (e.g. saying pod is running, when it really is not).
-	kubeletAPIPodsURL        = "http://127.0.0.1:10255/pods"
-	kubeletAPIRunningPodsURL = "https://127.0.0.1:10250/runningpods/"
-
 	activeCheckpointPath    = "/etc/kubernetes/manifests"
 	inactiveCheckpointPath  = "/etc/kubernetes/inactive-manifests"
 	checkpointSecretPath    = "/etc/kubernetes/checkpoint-secrets"
 	checkpointConfigMapPath = "/etc/kubernetes/checkpoint-configmaps"
-	kubeconfigPath          = "/etc/kubernetes/kubeconfig"
 
 	shouldCheckpointAnnotation = "checkpointer.alpha.coreos.com/checkpoint"    // = "true"
 	checkpointParentAnnotation = "checkpointer.alpha.coreos.com/checkpoint-of" // = "podName"
@@ -55,13 +49,29 @@ var (
 	lockfilePath string
 
 	// TODO(yifan): Put these into a struct when necessary.
-	nodeName     string
-	podName      string
-	podNamespace string
+	nodeName       string
+	podName        string
+	podNamespace   string
+	kubeconfigPath string
+
+	// podSerializer is an encoder for writing checkpointed pods.
+	//
+	// Perfer this instead of json.Marshal because it corrects metadata before
+	// serializing. For example it automatically fills in the "apiVersion" field.
+	podSerializer = scheme.Codecs.EncoderForVersion(
+		json.NewSerializer(
+			json.DefaultMetaFactory,
+			scheme.Scheme, // client-go's default scheme.
+			scheme.Scheme,
+			false, // don't pretty print.
+		),
+		v1.SchemeGroupVersion,
+	)
 )
 
 func init() {
 	flag.StringVar(&lockfilePath, "lock-file", "/var/run/lock/pod-checkpointer.lock", "The path to lock file for checkpointer to use")
+	flag.StringVar(&kubeconfigPath, "kubeconfig", "/etc/kubernetes/kubeconfig", "Path to a kubeconfig file containing credentials used to talk to the kubelet.")
 	flag.Set("logtostderr", "true")
 }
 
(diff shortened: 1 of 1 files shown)


cmd/checkpoint/main.go
```diff
@@ -486,47 +460,49 @@ func getAPIParentPods(client kubernetes.Interface, nodeName string) map[string]*
...
-	return podListToParentPods(&podList), nil
-}
-
-// localRunningPods uses the /runningpods/ kubelet api to retrieve the local container runtime pod state
-func (c *kubeletClient) localRunningPods() (map[string]*v1.Pod, error) {
-	resp, err := c.client.Get(kubeletAPIRunningPodsURL)
-	if err != nil {
-		return nil, fmt.Errorf("failed to contact kubelet pod api: %v", err)
+	if client.secureClient, err = rest.RESTClientFor(&secureConfig); err != nil {
+		return nil, fmt.Errorf("failed creating kubelet client: %v", err)
```
rphillips: secure client

ericchiang: I don't think the user actually cares about the difference between these two. Let me improve the docstrings on the kubeletClient struct explaining why we have an insecure client and a secure client.

pkg/asset/images.go
```diff
@@ -13,5 +13,5 @@ var DefaultImages = ImageVersions{
 	KubeDNS:         "gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.14.4",
 	KubeDNSMasq:     "gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64:1.14.4",
 	KubeDNSSidecar:  "gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.4",
-	PodCheckpointer: "quay.io/coreos/pod-checkpointer:4e7a7dab10bc4d895b66c21656291c6e0b017248",
+	PodCheckpointer: "quay.io/coreos/pod-checkpointer-dev:ea7cd433a74626ec303030bfac6acac3c0bf8773",
```
diegs: Please remember to revert.

cmd/checkpoint/main.go
```diff
@@ -486,47 +460,49 @@ func getAPIParentPods(client kubernetes.Interface, nodeName string) map[string]*
...
 
-	var podList v1.PodList
-	if err := json.Unmarshal(b, &podList); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal podlist: %v", err)
+// localRunningPods uses the /runningpods/ kubelet api to retrieve the local container runtime pod state
+func (c *kubeletClient) localRunningPods() map[string]*v1.Pod {
+	podList := new(v1.PodList)
+	if err := c.secureClient.Get().AbsPath("/runningpods/").Do().Into(podList); err != nil {
+		// Assume there are no local running pods.
+		glog.Errorf("failed to list local running pods: %v", err)
```
diegs: and here.

cmd/checkpoint/main.go
```diff
@@ -486,47 +460,49 @@ func getAPIParentPods(client kubernetes.Interface, nodeName string) map[string]*
...
-	defer resp.Body.Close()
-	b, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read response: %v", err)
+// localParentPods will retrieve all pods from kubelet api that are parents & should be checkpointed
+func (c *kubeletClient) localParentPods() map[string]*v1.Pod {
+	podList := new(v1.PodList)
+	if err := c.insecureClient.Get().AbsPath("/pods/").Do().Into(podList); err != nil {
+		// Assume there are no local parent pods.
+		glog.Errorf("failed to list local parent pods: %v", err)
```
diegs: would also note that will assume no pods are running, helpful for debugging.

pkg/asset/internal/templates.go
```diff
@@ -374,6 +374,7 @@ spec:
         image: {{ .Images.PodCheckpointer }}
         command:
         - /checkpoint
+        - --in-cluster=true
```
diegs: Maybe I missed something, but how will the statically running version flip over to --in-cluster=false?

ericchiang: I was under the impression it would checkpoint its own service account secret.

aaronlevy:

ericchiang:
As a reminder, kubelet authorization is a blocker for:
* Prometheus monitoring - being able to let a service account hit the secure API.
* TLS bootstrapping - so every client cert can't do arbitrary things against a kubelet.

This PR updates the checkpointer to
* Be okay if it can't contact the kubelet API. e.g. if the API server is down and kubelet authorization fails.
* Be able to take a configurable kubeconfig so it can stop using the kubelet's kubeconfig and use its own service account creds to talk to the kubelet in the future. 

Tested this locally with the following diff:

```diff
$ git diff .
diff --git a/hack/single-node/user-data.sample b/hack/single-node/user-data.sample
index a2f7411a..36543035 100644
--- a/hack/single-node/user-data.sample
+++ b/hack/single-node/user-data.sample
@@ -27,6 +27,8 @@ coreos:
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --allow-privileged \
           --anonymous-auth=false \
+          --authorization-mode=Webhook \
+          --authentication-token-webhook=true \
           --client-ca-file=/etc/kubernetes/ca.crt \
           --cluster_dns=10.3.0.10 \
           --cluster_domain=cluster.local \
diff --git a/pkg/asset/images.go b/pkg/asset/images.go
index d22450b6..319a8b23 100644
--- a/pkg/asset/images.go
+++ b/pkg/asset/images.go
@@ -13,5 +13,5 @@ var DefaultImages = ImageVersions{
        KubeDNS:         "gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.14.4",
        KubeDNSMasq:     "gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64:1.14.4",
        KubeDNSSidecar:  "gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.4",
-       PodCheckpointer: "quay.io/coreos/pod-checkpointer:4e7a7dab10bc4d895b66c21656291c6e0b017248",
+       PodCheckpointer: "quay.io/ericchiang/pod-checkpointer:81296cf2833f7b840910dce7e5a770a8b2705146",
 }
```

Was able to get `kubectl exec` working with the following role bindings:

```yaml
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: kubelet-admin
rules:
- apiGroups: [""]
  verbs: ["*"]
  resources: ["nodes/stats", "nodes/metrics", "nodes/log", "nodes/spec", "nodes/proxy"]
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: kubelet-admin
subjects:
- kind: User
  name: kube-apiserver
roleRef:
  kind: ClusterRole
  name: kubelet-admin
  apiGroup: rbac.authorization.k8s.io
```

cmd/checkpoint/main.go
```diff
@@ -362,16 +339,16 @@ func createCheckpointsForValidParents(client kubernetes.Interface, pods map[stri
 
 // writeCheckpointManifest will save the pod to the inactive checkpoint location if it doesn't already exist.
 func writeCheckpointManifest(pod *v1.Pod) error {
-	b, err := json.Marshal(pod)
-	if err != nil {
+	buff := new(bytes.Buffer)
```
diegs: optional: more idiomatic to write `&bytes.Buffer{}`

ericchiang: done

ericchiang:
(btw, I'm assigned but don't have merge access to this repo)

diegs:

PR #675, opened 2017-08-08
hack/quickstart: Mask locksmithd instead of update-engine

klausenbusk:
update-engine was initially masked to prevent the servers from
rebooting, when testing.
This caused locksmithd to restart forever and spam the system log,
as it depend on the update-engine.
With this change, the system would update but won't reboot
automatic.

### DIFF
--- a/hack/quickstart/init-master.sh
+++ b/hack/quickstart/init-master.sh
@@ -50,7 +50,7 @@ EOF
 # Initialize a Master node
 function init_master_node() {
     systemctl daemon-reload
-    systemctl stop update-engine; systemctl mask update-engine
+    systemctl stop locksmithd; systemctl mask locksmithd
 
     if [ "$SELF_HOST_ETCD" = true ] ; then
         echo "WARNING: THIS IS NOT YET FULLY WORKING - merely here to make ongoing testing easier"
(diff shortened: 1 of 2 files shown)


klausenbusk:
The bot seems a bit "spamish". 5 times `Can one of the admins verify this patch?`.

aaronlevy:
I'm on the fence if we should just enable the auto updates. Technically people should turn them on. I had them off in the quickstarts because during testing it can be annoying to have a node reboot on your while you're developing.

dghubble:
You probably want to leave update-engine alone and mask locksmithd. OS updates will be downloaded and installed on the next boot, but without locksmith coordinating reboots it shouldn't interrupt your development.

This setup also matches the [requirements](https://github.com/coreos/container-linux-update-operator#requirements) for deploying the container-linux-update-operator should you wish to test that on a vagrant-based setup.

dghubble:
@coreosbot ok to test

aaronlevy:
@klausenbusk would you mind updating the PR to just mask locksmithd and not update-engine (as @dghubble suggested)? That seems like a pretty reasonable default (download updates, but don't apply until reboot). And as mentioned - we might consider using the container-linux-update-operator at some point to coordinate updates across the cluster.

klausenbusk:
> @klausenbusk would you mind updating the PR to just mask locksmithd and not update-engine (as @dghubble suggested)?

Yes, of course. Done :)

PR #685, opened 2017-08-13
hack/quickstart: Make it easy to use COREOS_PUBLIC_IPV4 as hostname-o…

klausenbusk:
…verride

With PUBLIC_IP=true set, --hostname-override will be set to
${COREOS_PUBLIC_IPV4} instead of ${COREOS_PRIVATE_IPV4}.
This make most sense when using the quickstart scripts with a
cloud provider like DigitalOcean and you don't want to limit yourself
to a single location.
Maybe a bit out of scope for the quickstart script, but they are
very simple and get the job done.

### DIFF
--- a/hack/quickstart/init-master.sh
+++ b/hack/quickstart/init-master.sh
@@ -10,6 +10,7 @@ SSH_OPTS=${SSH_OPTS:-}
 SELF_HOST_ETCD=${SELF_HOST_ETCD:-false}
 CALICO_NETWORK_POLICY=${CALICO_NETWORK_POLICY:-false}
 CLOUD_PROVIDER=${CLOUD_PROVIDER:-}
+PUBLIC_IP=${PUBLIC_IP:-false}
 
 function usage() {
     echo "USAGE:"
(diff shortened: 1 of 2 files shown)


aaronlevy:
I'm not overly enthusiastic about adding new options to the quickstart scripts. They really are meant as references, where a user would fork/edit/commit their own kubelet.service files to suit their actual needs.

The risk of adding new options is that they essentially become untested configuration skew, and makes the debugging/support/maintenance much more difficult (e.g. user says something isn't working, might track down to firewall rules based on this flag).

That being said we have added similar options for things like cloud provider/ network policy / etc. So I'm not 100% opposed.

@dghubble @diegs any thoughts here?

diegs:
I think that `--hostname-override` is a bit niche to add flags for.

Some of the other things (cloud provider, network policy) are higher-level, mainstream concepts that are important for getting good self-hosting test coverage.

For smaller options like this I think people are better off customizing the rendered manifests before they invoke bootkube. We can clarify the README if that's not obvious to users for whatever reason.

klausenbusk:
Closing, I will strip-down and customize the `hack/quickstart` for my use-case.

PR #694, opened 2017-08-25
pkg/assets: Update Calico from v2.4.0 to v2.5.0

dghubble:
Changes since v2.4.0:
* Kubernetes datastore now uses CustomResourceDefinitions (CRDs) instead of TPRs
* Update RBAC [policies for Calico v2.5.0](https://github.com/projectcalico/calico/blob/v2.5.0/master/getting-started/kubernetes/installation/hosted/rbac-kdd.yaml) and CRDs
* Bug fixes in v2.4.1 https://github.com/projectcalico/calico/releases/tag/v2.4.1
* Fix naming to match upstream (easier to diff, no need for us to prefix everything with `kube-`)

Note:

* Anyone attempting to perform an in-place migration MUST follow https://github.com/projectcalico/calico/blob/master/upgrade/v2.5/README.md
* This should be a prereq for Calico leaving experimental, since we don't want to later write a TPR to CRD migration if it can be addressed upfront.

cc @squeed


dghubble:
Needs further changes.

dghubble:
Needs fixes, WIP

squeed:
Seems mostly reasonable, though the Calico deployment file seems out of date.Have you gotten the latest from https://docs.projectcalico.org/v2.5/getting-started/kubernetes/installation/hosted/kubernetes-datastore/ ?

squeed:
cc @lxpollitt

dghubble:
Applying these manifests to an existing cluster produced some weird results I'm still investigating.

dghubble:
@squat All the environment variable arguments match upstream now.

There are still a few unexplained differences from upstream. Upstream sets cniVersion to `0.1.0` (we have 0.3.0), sets the MTU explicitly, and mounts `/lib/modules` in `calico/node`.

aaronlevy:
`/lib/modules` shouldn't be needed if we're just using calico for network policy. Not sure about the rest

fasaxc:
@aaronlevy We mount /lib/modules into the calico/node container to allow the iptables and ipsets binaries to load their corresponding kernel modules (and so we can insert the IPIP kernel module if that feature is turned on).  (We use iptables/ipsets to implement Calico policy.)

If you don't mount /lib/modules then you need to make sure that all the kernel modules that we need are compiled in or pre-loaded before starting calico/node.  Since that part of the kernel is broken into dozens of modules, it's pretty fiddly to get that right.

dghubble:
Reviewing calico/node [requirements](https://github.com/projectcalico/calico/blob/master/master/reference/requirements.md), Container Linux (stable 1465.6.0) has all listed kernel modules loaded except `ip6_tables` and `ipt_rpfilter`,  (@euan points out `ip_tables` in builtin with the kernel) which are expected for Calico policy.

dghubble:
```
sudo -E ./calicoctl node checksystem
Checking kernel version...                          
                4.12.7-coreos                                           OK                               
Checking kernel modules...                          
                xt_set                                                  OK                               
                ip6_tables                                              OK                               
System meets minimum system requirements to run Calico!
```

fasaxc:
>  perhaps its only at runtime and they haven't been needed yet.

Yes, that's the case.  Calico doesn't load them itself.  To manipulate iptables/ipsets, we use the iptables/ipsets command line tools and they detect compiled in modules or load modules on-demand.  iptables has one module per match type so if Calico executes `iptables -A foo -m set` then the `xt_set` kernel module will get loaded.  We use, at least the `set`, `conntrack`, `addrtype`, `rpfilter`, `multiport`, `icmp` and `icmp6` match types so their corresponding modules are needed.

We also expect `nf_conntrack` and friends to be loaded.

I suspect our checksystem command needs to be beefed up if it's not checking for those additional modules.

squeed:
@fasaxc fun fact: the kernel will load modules like `xt_set` automatically, and will do so in the root mount namespace. It does this by (shudder) shelling out to modprobe directly.

Yes, that's right, the kernel shells out to modprobe directly. What a world.

Fortunately, it does so in the root namespace, so we don't need to worry about that. Just the initial module loads (which are a good enough reason to just mount in `/lib/modules`).

dghubble:
I manually checked the loaded modules against the requirements. Except for those I mentioned, they should already be loaded.

@squeed added `/lib/modules` mount to match upstream and so they have a shot at loading any missing  (or future) modules.

dghubble:
I don't have further changes. Looks good to everybody?

aaronlevy:
I'm not a huge fan of requiring `/lib/modules` be mounted into this pod - is it actually needed, or will the required modules be automatically loaded (and if they're not, why?)

pkg/asset/internal/templates.go
```diff
@@ -1225,6 +1225,8 @@ spec:
               value: "info"
             - name: CALICO_NETWORKING_BACKEND
               value: "none"
+            - name: CLUSTER_TYPE
+              value: "k8s"
```
tmjd: I think this should be "bootkube", this will indicate the specific integration and also that it is a K8s deployment.

pkg/asset/internal/templates.go
```diff
@@ -1263,6 +1265,8 @@ spec:
               port: 9099
             periodSeconds: 10
           volumeMounts:
+            - mountPath: /lib/modules
+              name: lib-modules
```
tmjd: You should add `readOnly: true` to the lib-modules mount. The Calico manifests at docs.projectcalico.org have it set so it should also be here.

pkg/asset/internal/templates.go
tmjd: If flannel is the only option for networking when using Calico then actually I think it should be "bootkube,canal".

dghubble: Ok, changed to that - bootkube always uses flannel for networking and calico only for policy so canal seems like a good name.

I'm experimenting with calico for networking in [matchbox](https://github.com/coreos/matchbox/pull/645) and [typhoon](https://github.com/poseidon/typhoon), so maybe that becomes an option here in future.

pkg/asset/internal/templates.go
dghubble: added

dghubble:
I'm on vacation next week. Talked to @pbx0 about fixups (if any) to merge this.

dghubble:
Rebased

PR #695, opened 2017-08-25
pkg/asset: Disable etcd-operator analytics

dghubble:
An incubator project probably shouldn't have this enabled by default.


dghubble:
ok to test

PR #704, opened 2017-09-12
pkg/asset: kube-{flannel,proxy}, calico-node: Tolerate all taints

### LINKED ISSUE #699: Worker nodes with taints are never ready
I'm trying to add nodes with pre-registered taints via the kubelet flag `--register-with-taints=foo=bar:NoSchedule`. However the node is never ready because certain pods like `kube-flannel` and `kube-proxy` will never be scheduled there even though they should be running on every node.

According to the documentation on taints https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/, perhaps some of these pods should specify tolerations to tolerate all taints. 

Ex.
```yaml
tolerations:
  - operator: "Exists"
```

klausenbusk:
kube-{flannel,proxy} need to run on every node, no matter what
the taints is. Without kube-{flannel,proxy} the node is never
ready.
I'm not sure if the node will work without calico-node, but it should
run on every node anyway.

Fix #699

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -684,9 +684,10 @@ spec:
       hostNetwork: true
       serviceAccountName: kube-proxy
       tolerations:
-      - key: node-role.kubernetes.io/master
+      - effect: NoSchedule
         operator: Exists
-        effect: NoSchedule
+      - effect: NoExecute
+        operator: Exists
       volumes:
       - name: lib-modules
         hostPath:
@@ -1143,9 +1144,10 @@ spec:
           mountPath: /host/opt/cni/bin/
       hostNetwork: true
       tolerations:
-      - key: node-role.kubernetes.io/master
+      - effect: NoSchedule
         operator: Exists
-        effect: NoSchedule
+      - effect: NoExecute
+        operator: Exists
       volumes:
         - name: run
           hostPath:
@@ -1224,9 +1226,10 @@ spec:
       hostNetwork: true
       serviceAccountName: calico-node
       tolerations:
-        # Allow the pod to run on master nodes
-        - key: node-role.kubernetes.io/master
-          effect: NoSchedule
+        - effect: NoSchedule
+          operator: Exists
+        - effect: NoExecute
+          operator: Exists
       containers:
         - name: calico-node
           image: {{ .Images.Calico }}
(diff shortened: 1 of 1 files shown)


squat:
coreosbot test this please

klausenbusk:
I just pushed this to production and it is working as expected.

weikinhuang:
Does this need to be added to the calico daemonset too?

klausenbusk:
> Does this need to be added to the calico daemonset too?

Definitely..

Done and rebased.

klausenbusk:
Are we waiting on anything, or can we merge this?

dghubble:
You could also add a toleration for your custom taint to the flannel or calico DaemonSet in the usual way. It feels a tad presumptuous to tolerate all possible taints, we've been wrong about broad assertions like that in the past. On the other hand I don't have a case in mind, since a NotReady node isn't so useful (maybe in debugging an issue, but its a stretch).

weikinhuang:
To be fair these are system level networking nodes. Is there a reason they shouldn't be running on a node?

dghubble:
Like I said, I don't have a concrete one in mind for flannel and calico. Some folks do run without kube-proxy for whatever reason (I'm not one of them).

The question we need to answer is not "is there a reason not to run them", its "is it ok to always run them" (onus lies on the broader assertion) because if it turns out there is some case, a user has to go find all the taints in use and edit components to re-add the explicit tolerations.

klausenbusk:
> You could also add a toleration for your custom taint to the flannel or calico DaemonSet in the usual way.



> because if it turns out there is some case, a user has to go find all the taints in use and edit components to re-add the explicit tolerations.

I think that most people, still want to run `kube-{flannel,proxy}, calico-node` after adding a taint. Not running `kube-{flannel,proxy}, calico-node` seems a bit like a corner case, and if people do that they is probably smart enough to tweak the tolerations them-self.

klausenbusk:
Any update on this?

klausenbusk:
Hmm.. Could someone decide if this is go or no-go? :)

dghubble:
I'm not opposed, but I'd feel better if upstream Calico manifests switched or indicated this was pending. They're ultimately the main upstream for these manifests.

dghubble:
coreosbot run e2e

klausenbusk:
Rebased..

klausenbusk:
Upstream PR: https://github.com/projectcalico/calico/pull/1443

klausenbusk:
New upstream PR: https://github.com/projectcalico/calico/pull/1634

klausenbusk:
@dghubble The upstream PR (https://github.com/projectcalico/calico/pull/1634) was just merged. I have rebased the PR and implemented it as upstream.

Can we merge this now?

dghubble:
ok to test

dghubble:
Please update the description since the conclusion was to update the tolerations, not to allow all.

PR #705, opened 2017-09-12
pkg/asset: Update flannel-cni from v0.2.0 to v0.3.0

dghubble:
Bump flannel-cni to [v0.3.0](https://github.com/coreos/flannel-cni/releases/tag/v0.3.0), but retain the original flannel CNI config name. This should make no functional change in clusters and should pass tests since flannel-cni:v0.3.0 is [compatible](https://github.com/coreos/flannel-cni/pull/5#issuecomment-328663841) with prior versions.

Then, we can either merge this and followup with #697 or skip and go directly to an updated #697 which changes the CNI config to add portmap.

### Testing

I've tested a bare-metal cluster with this change.


dghubble:
Hm, that failed e2e. Let's try once more in case its a fluke. Otherwise this isn't ok.

dghubble:
Ok, seems fine.

PR #707, opened 2017-09-13
Add optional template files

realfake:
This PR adds the ability to render assets from a folder with templates,
it allows users to customize the behaviour of bootkube without recompiling/(releasing) it.

The rendering from template files will be invoked through the `--asset-template-dir` flag.


aaronlevy:
Hi @realfake sorry for the delayed response.

This is an interesting approach. I guess my concern would be that there is still a high-likelihood of breakage with bootkube releases if there are mis-matches between the external templates and compiled-in parameters.

There pretty consistently has been a want for more flexible templating - but it is a big risk (to testability and maintainability) to try and have bootkube be the component that is responsible for trying to solve everyone's use-case. See the [design principles doc](https://github.com/kubernetes-incubator/bootkube/blob/master/Documentation/design-principles.md) for a bit more discussion on this.

I like that this is a more generic approach - but I still have concerns that now the internal config objects would be tied to external template assumptions (and it's a new backwards compatibility matrix to track).

Just as a point of discussion, what kind of customized behaviors are you looking to solve?

One thing I wanted (or maybe someone else wanted) to look into was the possibility of using the patch/overlay model discussed as part of the [declarative application management proposal](https://docs.google.com/document/d/1cLPGweVEYrVqQvBLJg6sxV-TrE5Rm2MNOBA_cxZP2WU) (towards the end).

Super vague, but what this might look like:

You have the `bootkube render` output to a normal manifest dir
```
/manifest-dir
/manifest-dir/kube-apiserver.yaml
```
In this case `kube-apiserver.yaml` is the full rendered manifest

Then you have your overlay objects which will be layered essentially using the same logic as `kubectl apply`:

```
/overlays
/overlays/my-kube-apiserver-overlay.yaml
```

Where `my-kube-apiserver-overlay.yaml` would just be a subset of the manifest for the things you wish to change (say maybe add new volume mounts).

Then even as `bootkube render` is updated, as long as your overlays can be cleanly applied to the underlying object (e.g. no naming conflicts for the volumes), bootkube has no concept of your customizations - and it is a purely generic, repeatable, and in most cases backwards compatible process.

The downside of this approach is that it doesn't work well for all customizations (and why I was asking about your use-case). For example, customization of flags (https://github.com/kubernetes-incubator/bootkube/issues/712) is not immediately supported (would have to write our own merge logic to handle that) -- but it's not insurmountable.

dghubble:
There was also discussion in #708

aaronlevy:
Just read through https://github.com/kubernetes-incubator/bootkube/pull/708

So it seems like the bootkube render parameters are what are actually wanted? Might be an option that we dump all those values into an object that could then be more easily parsed (compared to regex) by subsequent (and external) render steps.

dghubble:
I'm for closing this. If anything, we're working to narrow the scope for bootkube, with `start` being the important piece and `render` being an example of one plausible way to render manifests. cc @ericchiang

ericchiang:
@dghubble agreed but I also think that that opinion isn't clear for people who are new to the repo.

Let's close this, but make a better effort to document what `render` is, how it should be used, and how you should modify the result of.

PR #708, opened 2017-09-13
Render optional templates

realfake:
This PR is build upon #707.

This PR allows bootkube to render and deploy additional assets after it has been bootstrapped.

Through the `--asset-values-file` flag values can be invoked, this can also help to get rid of hardcoded versions.


dghubble:
In my opinion, bootkube should stay focused on bootstrapping a Kubernetes control plane. Once that's up, new manifests can be deployed using the generated credentials and any of the many different tools that target Kubernetes (`kubectl` manifests, terraform, puppet, helm, etc).

I'm ok with there being a special directory of plain manifests that get created (similar to kube addon-manager and kinda possible now), but taking on responsibility for templating additional manifests is an increase in project scope.

realfake:
Yes, this feature is for rendering add-ons and other basic templates that can be part of the bootstrap process, such as https://github.com/kube-node/kube-machine. This enables the cluster to add nodes during the bootstrap phase. To do that we need access to the values from the rendering process (But they could also be obtained through some regex afterwards). This PR should not be a second helm but however ease the creation and deployment of bootstrapping related templates.

dghubble:
I've seen this templating issue come up repeatedly across projects - whichever component controls the templating, becomes responsible for an unexpected amount of scope. Here, `bootkube render` does some templating, so naturally you're forced to extend that to utilize it for your own extended templating needs.

I think we should strictly consider `bootkube render` a reference or example tool. Many (most?) cluster creation tools based on bootkube render the manifests as well as other components using some existing system designed for that purpose (e.g. Tectonic uses Terraform to render bootkube and other manifests). If bootkube supports general templating, someone will ask for partials, template functions, tools to export variables, etc. The scope grows, when there are systems designed for this purpose.

rimusz:
I totally agree with @dghubble - bootkube should stay focused on bootstrapping a Kubernetes control plane only, otherwise it will lose it's scope

realfake:
I understand, putting this into the renderer opens the door to too many things bootkube is not meant for. 
Do you think deploying optional assets is part of bootkube or should it be handled by another tool?

rimusz:
deploying optional assets could be a part bootkube

dghubble:
`bootkube start` mainly manifests (which may contain custom manifests) into the pod-manifest directory since there is just a contract where "some render step" prepares manifests in the right structure and `bootkube start` performs the prescribed bootstrapping and creation.

In practice, I'd say we should be careful not to abuse that mechanism (since most manifests are not essential for the control plane and can be created later), but its a nice mechanism for flexibility.

PR #709, opened 2017-09-13
WIP: bump k8s to 1.8.0-beta.1

rphillips:
Issues:

- [ ] modify cluster/manifests/kube-system-rbac-role-binding.yaml to use `rbac.authorization.k8s.io/v1beta1` instead of v1alpha1
  - v1beta1 changes the selectors to be immutable

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -19,7 +19,7 @@ contexts:
     user: kubelet
 `)
 
-var KubeSystemSARoleBindingTemplate = []byte(`apiVersion: rbac.authorization.k8s.io/v1alpha1
+var KubeSystemSARoleBindingTemplate = []byte(`apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
   name: system:default-sa
--- a/hack/multi-node/user-data.sample
+++ b/hack/multi-node/user-data.sample
@@ -9,7 +9,7 @@ coreos:
         [Service]
         EnvironmentFile=/etc/environment
         Environment=KUBELET_IMAGE_URL=quay.io/coreos/hyperkube
-        Environment=KUBELET_IMAGE_TAG=v1.7.5_coreos.0
+        Environment=KUBELET_IMAGE_TAG=v1.8.0-beta.1_coreos.0
         Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \
           --volume var-lib-cni,kind=host,source=/var/lib/cni \
           --volume opt-cni-bin,kind=host,source=/opt/cni/bin \
--- a/hack/multi-node/user-data.sample
+++ b/hack/multi-node/user-data.sample
@@ -9,7 +9,7 @@ coreos:
         [Service]
         EnvironmentFile=/etc/environment
         Environment=KUBELET_IMAGE_URL=quay.io/coreos/hyperkube
-        Environment=KUBELET_IMAGE_TAG=v1.8.0-beta.1_coreos.0
+        Environment=KUBELET_IMAGE_TAG=v1.8.0_coreos.0
         Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \
           --volume var-lib-cni,kind=host,source=/var/lib/cni \
           --volume opt-cni-bin,kind=host,source=/opt/cni/bin \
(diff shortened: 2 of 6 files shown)


squat:
Awesome

klausenbusk:
The `apiVersion` of DaemonSet/Deployment should also be updated.

https://github.com/kubernetes/features/blob/master/release-1.8/release_notes_draft.md#major-themes

pkg/asset/internal/templates.go
```diff
@@ -19,7 +19,7 @@ contexts:
     user: kubelet
 `)
 
-var KubeSystemSARoleBindingTemplate = []byte(`apiVersion: rbac.authorization.k8s.io/v1alpha1
+var KubeSystemSARoleBindingTemplate = []byte(`apiVersion: rbac.authorization.k8s.io/v1beta1
```
klausenbusk: Should be `v1` per the release notes draft: https://github.com/kubernetes/features/blob/master/release-1.8/release_notes_draft.md#auth-1

ericchiang:
Seeing this error when trying to use this PR

```
Ignoring json-2.0.3 because its extensions are not built.  Try: gem pristine json --version 2.0.3
Ignoring nokogiri-1.6.8.1 because its extensions are not built.  Try: gem pristine nokogiri --version 1.6.8.1
Starting temporary bootstrap control plane...
Waiting for api-server...
Creating self-hosted assets...
	created          kube-apiserver secret
	created kube-controller-manager poddisruptionbudget
	created kube-controller-manager secret
	created kube-controller-manager deployment
	created                kube-dns deployment
	created                kube-dns service
	created        kube-flannel-cfg configmap
	created          kube-scheduler poddisruptionbudget
	created          kube-scheduler deployment
	created       system:default-sa clusterrolebinding
Error creating assets: [error when creating "/home/core/cluster/manifests/kube-apiserver.yaml": DaemonSet.apps "kube-apiserver" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"k8s-app":"kube-apiserver", "tier":"control-plane"}: `selector` does not match template `labels`, error when creating "/home/core/cluster/manifests/kube-flannel.yaml": DaemonSet.apps "kube-flannel" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"k8s-app":"flannel", "tier":"node"}: `selector` does not match template `labels`, error when creating "/home/core/cluster/manifests/kube-proxy.yaml": DaemonSet.apps "kube-proxy" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"k8s-app":"kube-proxy", "tier":"node"}: `selector` does not match template `labels`, error when creating "/home/core/cluster/manifests/pod-checkpointer.yaml": DaemonSet.apps "pod-checkpointer" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"tier":"control-plane", "k8s-app":"pod-checkpointer"}: `selector` does not match template `labels`]

NOTE: Bootkube failed to create some cluster assets. It is important that manifest errors are resolved and resubmitted to the apiserver.
For example, after resolving issues: kubectl create -f <failed-manifest>

	Pod Status: kube-controller-manager	Pending
	Pod Status:        pod-checkpointer	DoesNotExist
	Pod Status:          kube-apiserver	DoesNotExist
	Pod Status:          kube-scheduler	Pending
```

pkg/asset/internal/templates.go
ericchiang: Either works. v1alpha1 doesn't anymore due to https://github.com/kubernetes/kubernetes/pull/51839

rphillips:
Reproduced... It has something to do with defaulting labels now. Looking into it.

[Reference](https://github.com/kubernetes/kubernetes/pull/50719)

rphillips:
@ericchiang fixed... updated the selectors on the daemonsets.

pkg/asset/internal/templates.go
```diff
@@ -1193,7 +1205,7 @@ data:
     }
 `)
 
-var KubeCalicoTemplate = []byte(`apiVersion: extensions/v1beta1
```
dghubble: Look for CalicoNodeTemplate, this was renamed on master

ericchiang:
@rphillips can we update this PR now that 1.8 is out? (cc @squat)

squat:
1.8.0 hyperkube is about to get cut. Waiting for the necessary LGTM and it will be available. xref https://github.com/coreos/kubernetes/pull/167

rphillips:
yep.. then I'll tweak this PR to remove the API bumps, and do a migration of the APIs in a separate PR

rphillips:
closing in favor of https://github.com/kubernetes-incubator/bootkube/pull/729

Issue #698, opened 2017-08-30
Calico network policy daemonset cannot be removed


dghubble:
Found a reproducible way to cause what I experienced in https://github.com/kubernetes-incubator/bootkube/pull/694#issuecomment-325754554

We're not deploying Calico Network Policy in a way where it can be removed cleanly. It drops its own CNI configs on hosts which actually seem to take precedence over the flannel CNI config. Removing the calico daemonset after this point causes new pods to fail to create network namespaces.

UPDATE: Removing Calico network policy requires manually removing a file on each host.

## Reproduce

First, deploy a cluster with calico network policy enabled. You can do this at cluster creation time or by applying the manifests to an existing cluster. Later, delete the calico daemonset (perhaps you don't want network policy for some reason). All *looks* normal. Launch some new pods and observe that they're stuck in ContainerCreating.

```
1m            1m              1       kubelet, sm-x8dtu2.us-west1-m.dghubble.io                       Normal          SuccessfulMountVolume   MountVolume.SetUp succeeded for volume "default-token-7db68" 
  1m            16s             8       kubelet, sm-x8dtu2.us-west1-m.dghubble.io                       Normal          SandboxChanged          Pod sandbox changed, it will be killed and re-created.
  1m            15s             9       kubelet, sm-x8dtu2.us-west1-m.dghubble.io                       Warning         FailedSync              Error syncing pod
```

The Kubelet finds CNI configs in `--cni-config-dir` is `/etc/kubernetes/cni/net.d`. After Calico is deployed on a cluster, the `calico` plugin is installed and so is a Calico CNI config (/etc/kubernetes/cni/net.d/10-calico.yaml). Checking logs, we see Kubelet seems to be calling calico to setup the network.

```
Aug 30 21:17:49 sm-x8dtu2.us-west1-m.dghubble.io kubelet-wrapper[6115]: time="2017-08-30T21:17:49Z" level=debug msg="Performing 'Get' for WorkloadEndpoint(node=sm-x8dtu2.us-west1-m.dghubble.io, orchestrator=k8s, workload=ingress.ingress-controller-public-3120200946-skdd2, name=eth0)"
Aug 30 21:17:49 sm-x8dtu2.us-west1-m.dghubble.io kubelet-wrapper[6115]: E0830 21:17:49.640629    6115 cni.go:312] Error deleting network: connection is unauthorized
[...]

PR #711, opened 2017-09-15
pkg/asset: calico: Enable hostPort for calico networking & update cniVersion to 0.3.1

weikinhuang:


weikinhuang:
I have tested on my local cluster bootstrapped with bootkube and the `--experimental-calico-network-policy` flag. Per the comments in #698 and #697 this configuration won't be loaded by default on an existing cluster running calico because `10-calico.conf` will be loaded ahead of `10-calico.conflist`.

This should work out of the box for new clusters. For existing clusters `10-calico.conf` must be deleted on every node and the daemonset must be restarted.

weikinhuang:
I signed the CLA

pkg/asset/internal/templates.go
```diff
@@ -1174,24 +1174,34 @@ data:
...
     {
-        "name": "k8s-pod-network",
-        "cniVersion": "0.3.0",
-        "type": "calico",
-        "log_level": "debug",
-        "datastore_type": "kubernetes",
-        "nodename": "__KUBERNETES_NODE_NAME__",
-        "ipam": {
+      "name": "k8s-pod-network",
+      "cniVersion": "0.3.0",
```
squeed: This should be 0.3.1 - we fixed a small bug.

squeed:
FWIW, the changes in here will also work with Flannel. I can file a PR to update that.

pkg/asset/internal/templates.go
weikinhuang: Sure thing

dghubble:
@squeed this change is made to flannel in https://github.com/kubernetes-incubator/bootkube/pull/697, but paired with changes in the flannel-cni image to cleanup the old CNI config in a way that is backwards compatible whether folks want to fix hostPort or not.

weikinhuang:
Updated `cniVersion`

dghubble:
bootkube won't properly support Calico networking until #714. Installing policy-only Calico on top of cluster using flannel connectivity, this may work because the new CNI config is ordered first. But its a hack on top of a hack.

Let's test this following #714, in a setup that does not use flannel at all and runs Calico networking alone.

dghubble:
Rebase please

weikinhuang:
done

dghubble:
coreosbot test e2e

dghubble:
coreosbot run e2e

dghubble:
coreosbot run e2e

dghubble:
Seems to have failed to bootstrap the cluster.

```
I1010 06:29:43.213792    1089 log.go:19] daemonset "pod-checkpointer" created
	Pod Status:        pod-checkpointer	Running
	Pod Status:          kube-apiserver	Running
	Pod Status:          kube-scheduler	Pending
	Pod Status: kube-controller-manager	Pending
Error: error while checking pod status: timed out waiting for the condition
Tearing down temporary bootstrap control plane...
Error: error while checking pod status: timed out waiting for the condition
Error: error while checking pod status: timed out waiting for the condition
error while checking pod status: timed out waiting for the condition
```

dghubble:
Rebase again as well? Thanks

weikinhuang:
rebased again

dghubble:
coreosbot run e2e

weikinhuang:
@dghubble I missed that there's 2 `calico-node` templates now, I updated the other one now too. should fix the issue.

dghubble:
coreosbot run e2e

dghubble:
It [doesn't look like](https://github.com/projectcalico/cni-plugin/blob/a66b1a209123e9c51e930d744eb92209eff81215/k8s-install/scripts/install-cni.sh#L133) Calico's install-cni script will remove an existing old CNI config file, like flannel's install-cni script [does](https://github.com/coreos/flannel-cni/blob/master/install-cni.sh#L21). Looks like if a user does an in-place upgrade, they'll need to remove the old config on their own.

weikinhuang:
Unfortunately that is correct. It would be a no-op on upgrades without manual intervention.

squeed:
Oh, it's much more fun than that: the kubelet re-scans for CNI configuration files every 5 seconds. It also prefers "conflists" over conf. So it _should_ work. Emphasis on the should.

weikinhuang:
I'll open a PR to calico's cni-plugin mirroring the flannel change. https://github.com/projectcalico/cni-plugin/pull/392

weikinhuang:
I'll update this PR if the calico changes get accepted and a version bump happens

dghubble:
Looks like we're just waiting on a new calico cni-plugin.

pkg/asset/internal/templates.go
```diff
@@ -1183,24 +1183,34 @@ data:
...
     {
-        "name": "k8s-pod-network",
-        "cniVersion": "0.3.0",
-        "type": "calico",
-        "log_level": "info",
-        "datastore_type": "kubernetes",
-        "nodename": "__KUBERNETES_NODE_NAME__",
-        "ipam": {
+      "name": "k8s-pod-network",
+      "cniVersion": "0.3.1",
```
dghubble: Call out this update in the commit message and PR

weikinhuang:
Looks like the next version of calico is a major version bump to `v3.x.x`

alexef:
We're also interested in this, are we waiting for the 3.x version release?

pkg/asset/internal/templates.go
```diff
@@ -1219,24 +1219,34 @@ data:
...
+          },
+          "kubernetes": {
             "k8s_api_root": "https://__KUBERNETES_SERVICE_HOST__:__KUBERNETES_SERVICE_PORT__",
             "kubeconfig": "__KUBECONFIG_FILEPATH__"
+          }
+        },
+        {
+          "type": "portmap",
+          "capabilities": {
+            "portMappings": true
```
redbaron: adding 
```
"externalSetMarkChain": "KUBE-MARK-MASQ"
```

will reuse existing iptables chains

redbaron:
Successfully bootstrapped clusters with bootkube 0.9, this change  and calico

pkg/asset/internal/templates.go
squeed: Not yet  - we have to do a plugins release first :-)

redbaron: latest bootkube uses https://github.com/projectcalico/cni-plugin/releases/tag/v1.11.1 which already includes portmap plugin. Isn't it enough?

weikinhuang: Pending support for `CNI_OLD_CONF_NAME`

redbaron: Do we need to wait for it? 
  1. is bootkube concerned with compatibility with existing clusters?
  2. Should you name it 05-calico.conflist it will be loaded ahead of existing 10-calico.conf as per https://github.com/kubernetes/kubernetes/blob/v1.8.4/pkg/kubelet/network/cni/cni.go#L107

dghubble:
> We're also interested in this, are we waiting for the 3.x version release?
 
Yes.

> Do we need to wait for it?

We waited for flannel-cni to provide a mechanism to rotate between CNI configs (and cleanup the old manifest) to make migrations easier for downstreams. Its true you can name a new config to be alphabetically first and cleanup the old config manually or later (not great). Depends how long we'd like to wait.

redbaron:
> We waited for flannel-cni
TBH I fail to see what value flannel adds comparing to calico mode where it does ipip tunneling. vxlan gives L2 overlay network, but 99.999% of users don't need L2 in kube

squeed:
Flannel doesn't need to rotate configs, only Kubelet does - and it will always use the "first" CNI configuration file, for which it scans every 5 seconds. The challenge, then, is ensuring that a network created with config A can be deleted with config B.

In this case, it will work, since CNI delete is idempotent, and the only change is adding the Portmap plugin.

redbaron:
@squeed , OMG, you say, that kubelet doesn't "memoize' cni config for each network it created so that it can reuse it at tear down time? That is unfortunate oversight.

dghubble:
Looks like config cleanup is now in https://github.com/projectcalico/calico/releases/tag/v2.6.4

weikinhuang:
@dghubble should I update the version number, or do you want to do that in a separate PR, and I'll rebase this on top after a merge?

dghubble:
Let's keep the version bump separate from revertability. Kicked off e2e tests in #818

dghubble:
Should be good to rebase :+1:

dghubble:
Gah, the separate sidecar image didn't get bumped to v1.11.1

weikinhuang:
haha, I'll rebase after #819 is merged

weikinhuang:
This has been rebased against #819

weikinhuang:
ping @dghubble

dghubble:
Looks fine, but I'm on vacation til Monday and haven't validated the
in-place conf edit/migration yet. We also should be sure to have a
changelog note, similar to the one for flannel a few cycles back.

Ping @diegs

diegs:
ok to test

pkg/asset/internal/templates.go
```diff
@@ -1302,6 +1312,10 @@ spec:
           image: {{ .Images.CalicoCNI }} 
           command: ["/install-cni.sh"]
           env:
+            - name: CNI_CONF_NAME
+              value: 10-calico.conflist
+            - name: CNI_OLD_CONF_NAME
+              value: 10-calico.conf
```
dghubble: I see this is the default in the sidecar so it could be left off.

dghubble:
Release notes blurb:

* Change the default Calico CNI config to enable portmap to fix hostPort (@weikinhuang #711 )
  * Sidecar will automatically cleanup any old CNI `10-calico.conf`
  * To retain the old behavior, set `CNI_CONF_NAME` to `10-calico.conf` and continue using the old calico-config ConfigMap (i.e. lacking the portmap plugin configuration)

PR #713, opened 2017-09-19
Moved checkpointer away from using the /runningpods endpoint

thourfor:
Instead use the CRI shim directly to obtain the running pods list


thourfor:
.

diegs:
Can you please rebase the vendor changes into a separate commit? Makes it much easier to read the actual diff.

thourfor:
sounds good

aaronlevy:
cc @ericchiang 

Also - just as a heads-up in case adopted in kubeadm docs: @luxas @timothysc

cmd/checkpoint/main.go
```diff
@@ -505,14 +545,85 @@ func (c *kubeletClient) localParentPods() map[string]*v1.Pod {
...
+			p := &v1.Pod{}
+			p.UID = types.UID(s.Metadata.Uid)
+			p.Name = s.Metadata.Name
+			p.Namespace = s.Metadata.Namespace
+
+			pods[podName] = p
+		}
+	}
+
+	containers, err := c.getKubeletContainers(false)
```
aaronlevy: I like the idea of using the sandboxes as a means of tracking if the pod is expected to be running. Are there situations where we need to check both sandbox and containers? e.g. I wouldn't expect a container to ever be running if the sandbox was not (networking wouldn't work) - at least in the case of docker (no idea for other runtimes).

aaronlevy: Overall I don't expect that this would cause a problem - just curious if this is in fact a situation we might hit. cc @yifan-gu - any thoughts here?

cmd/checkpoint/main.go
```diff
@@ -505,14 +545,85 @@ func (c *kubeletClient) localParentPods() map[string]*v1.Pod {
...
+	if !all {
+		filter.State = &runtimeapi.ContainerStateValue{
+			State: runtimeapi.ContainerState_CONTAINER_RUNNING,
+		}
+	}
+
+	return c.criClient.listContainers(filter)
+}
+
+func (c *kubeletClient) getKubeletSandboxes(all bool) ([]*runtimeapi.PodSandbox, error) {
```
aaronlevy: nit: `all` isn't used

cmd/checkpoint/main.go
```diff
@@ -505,14 +545,85 @@ func (c *kubeletClient) localParentPods() map[string]*v1.Pod {
...
+			p.Namespace = c.Labels[kubelettypes.KubernetesPodNamespaceLabel]
+
+			pods[podName] = p
+		}
+	}
+
+	return pods
+}
+
+func (c *kubeletClient) getKubeletContainers(all bool) ([]*runtimeapi.Container, error) {
```
aaronlevy: nit: `all` isn't used

aaronlevy:

cmd/checkpoint/main.go
```diff
@@ -43,16 +49,21 @@ const (
 
 	shouldCheckpoint = "true"
 	podSourceFile    = "file"
+
+	defaultRuntimeEndpoint       = "unix:///var/run/dockershim.sock"
```
diegs: nit: these constants are only used once, inline?

cmd/checkpoint/main.go
thourfor: It's used 4 lines down

`if !all {`

cmd/checkpoint/main.go
thourfor: `if !all`

cmd/checkpoint/main.go
```diff
@@ -505,14 +545,85 @@ func (c *kubeletClient) localParentPods() map[string]*v1.Pod {
...
+	sandboxes, err := c.getKubeletSandboxes(false)
+	if err != nil {
+		glog.Errorf("failed to list running sandboxes: %v", err)
+		return nil
+	}
+
+	// Add all pods that sandboxes are apart of
+	for _, s := range sandboxes {
+		if s.Metadata == nil {
+			glog.V(4).Infof("Sanbox does not have metadata: %+v", s)
```
diegs: nit: Sandbox

cmd/checkpoint/main.go
```diff
@@ -505,14 +545,85 @@ func (c *kubeletClient) localParentPods() map[string]*v1.Pod {
...
+
+	pods := make(map[string]*v1.Pod)
+
+	sandboxes, err := c.getKubeletSandboxes(false)
+	if err != nil {
+		glog.Errorf("failed to list running sandboxes: %v", err)
+		return nil
+	}
+
+	// Add all pods that sandboxes are apart of
```
diegs: nit: do you mean "Add pods from all sandboxes"?

cmd/checkpoint/main.go
thourfor: This behavior of checking both sandboxes and containers is how the kubelet performs the GetPods() action that was the implementation of the /runningpods endpoint. So I kept to to prevent changes to behavior

cmd/checkpoint/main.go
```diff
@@ -505,14 +545,85 @@ func (c *kubeletClient) localParentPods() map[string]*v1.Pod {
 	return podListToParentPods(podList)
 }
 
-// localRunningPods uses the /runningpods/ kubelet api to retrieve the local container runtime pod state
+// localRunningPods uses the CRI shim to retrieve the local container runtime pod state
```
diegs: Can you add a brief comment explaining why you use the union of pods from both sandboxes and containers?

cmd/checkpoint/main.go
aaronlevy: From in-person discussion: possibly redundant - but this is safer regardless (and what the kubelet does - so may as well do the same).

cmd/checkpoint/main.go
yifan-gu: I think aaron means we always call with `false` so the boolean is redundant?
I am fine with leave it as is tho.

cmd/checkpoint/main.go
thourfor: added comment above the calls to obtain running sandboxes

cmd/checkpoint/main.go
yifan-gu: yeah, checking both here is more accurate as there can be a moment that the "sandbox" is down, but the containers are still running. They will be killed by kubelet shortly after kubelet finds the sandbox is down. Then kubelet will restart the whole pod (sandbox + containers).
And I think that's not the case we want to activate the checkpoint since kubelet will try to restart it. So I agree with keeping as is here.

thourfor:

diegs:
ok to test

aaronlevy:
coreosbot run e2e

aaronlevy:

aaronlevy:
The e2e that is failing is unrelated. Merging

Issue #405, opened 2017-03-24
Use CNI pod to deploy CNI plugins to host filesystem


heschlie:
Raising this here at @aaronlevy request

We would like to push forward on using the host file system for the CNI binaries instead of the hyperkube file system.  This lets 3rd parties deploy the desired version of their CNI bin instead of relying on the version that is shipped with hyperkube.

This also creates some friction when deploying a CNI provider like Calico as the current standard operation is to change the `RKT_OPTS` in the kubelet to mount `/opt/cni/bin` in hyperkube to the host filesystem.  This seems sub-optimal as it wipes out the existing CNI binaries which then need to be redeployed.

aaronlevy:
The general plan would be that the kubelet.service file would add a host volume mount of `/opt/cni/bin` which would be unpopulated at initialization. It would then be the responsibility of each network plugin to supply the required CNI plugins.

For each plugin, the podspec would need to include a host volume mount also for `/opt/cni/bin`. Then (maybe as a side-car container?) the pod would be responsible for copying the CNI plugin binary from the container filesystem into that hostmount, so that the binary is now available to the kubelet.

Some questions to be answered:

- How is the flannel CNI plugin distributed? (added to the flannel container fs, or kept separately?)
- How are "general" CNI plugins like IPAM/Bridge distributed? Are there container distributions of these?

I'm not immediately sure this is the route to take - but it does seem fairly reasonable.

tomdee:
The flannel plugin is included in "general" plugins distribution so that might just answer that first question. The "official" CNI releases include a tarball and an ACI which could potentially be used as a side car container. If we can come up with a reasonable design I can help get it into upstream CNI if required.

joshrosso:
This (mounting `/opt/cni/bin` in the kubelet) is the route I've taken for multiple Tectonic installs to achieve Canal.

Please let me know if I can be of any help in how it's been bootstrapped previously.

mikebryant:
Some stuff we found out while figuring this out in our setup:
We currently have this for the kubelet:
```
          --volume=opt-cni,kind=host,source=/opt/cni,readOnly=true \
          --mount volume=opt-cni,target=/opt/weave-net \
```
and weave has this:
```
      volumes:
        - name: cni-bin
          hostPath:
            path: /opt
```

[...]

PR #714, opened 2017-09-19
Add network provider option and Calico support

### LINKED ISSUE #407: Documentation: using alternative CNI plugins
Calico / Canal etc. They're all deployed as CNI plugins so they hopefully should be fairly interchangeable. As this is tested - we should add the process to the documentation.

dghubble:
* Support choosing Flannel or Calico networking (`--network-provider`)
* On Jenkins, I've added a new test to provision a cluster with calico and perform the usual e2e tests.

This is based on [poseidon/bootkube-terraform](https://github.com/poseidon/bootkube-terraform), which uses Calico or flannel as the network provider on [AWS, GCE, and bare-metal clusters](https://github.com/poseidon/typhoon) (Digital Ocean w caveats). I figured it would be nice to have in the upstream project in case others want to use it.

rel #405 
Closes #407

### DIFF
--- a/cmd/bootkube/render.go
+++ b/cmd/bootkube/render.go
@@ -113,8 +113,8 @@ func validateRenderOpts(cmd *cobra.Command, args []string) error {
 	if renderOpts.apiServers == "" {
 		return errors.New("Missing requried flag: --api-servers")
 	}
-	if renderOpts.networkProvider != asset.NetworkFlannel && renderOpts.networkProvider != asset.NetworkCanal {
-		return errors.New("Must specify --network-provider flannel or experimental-canal")
+	if renderOpts.networkProvider != asset.NetworkFlannel && renderOpts.networkProvider != asset.NetworkCalico && renderOpts.networkProvider != asset.NetworkCanal {
+		return errors.New("Must specify --network-provider flannel or experimental-calico or experimental-canal")
 	}
 	return nil
 }
--- a/hack/quickstart/init-master.sh
+++ b/hack/quickstart/init-master.sh
@@ -61,6 +61,8 @@ function init_master_node() {
 
     if [ "$NETWORK_PROVIDER" = "canal" ]; then
         network_provider_flags="--network-provider=experimental-canal"
+    elif [ "$NETWORK_PROVIDER" = "calico" ]; then
+        network_provider_flags="--network-provider=experimental-calico"
     else
         network_provider_flags="--network-provider=flannel"
     fi
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -1225,6 +1225,130 @@ spec:
       hostNetwork: true
       serviceAccountName: calico-node
       tolerations:
+        # Allow the pod to run on master nodes
+        - key: node-role.kubernetes.io/master
+          effect: NoSchedule
+        # Mark the pod as a critical add-on for rescheduling
+        - key: "CriticalAddonsOnly"
+          operator: "Exists"
+      containers:
+        - name: calico-node
+          image: {{ .Images.Calico }}
+          env:
+            - name: DATASTORE_TYPE
+              value: "kubernetes"
+            - name: FELIX_LOGSEVERITYSCREEN
+              value: "info"
+            - name: CLUSTER_TYPE
+              value: "k8s,bgp"
+            - name: CALICO_DISABLE_FILE_LOGGING
+              value: "true"
+            - name: FELIX_DEFAULTENDPOINTTOHOSTACTION
+              value: "ACCEPT"
+            - name: FELIX_IPV6SUPPORT
+              value: "false"
+            - name: WAIT_FOR_DATASTORE
+              value: "true"
+            - name: CALICO_IPV4POOL_CIDR
+              value: "{{ .PodCIDR }}"
+            - name: CALICO_IPV4POOL_IPIP
+              value: "always"
+            - name: FELIX_IPINIPENABLED
+              value: "true"
+            - name: NODENAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            - name: IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+            - name: FELIX_HEALTHENABLED
+              value: "true"
+          securityContext:
+            privileged: true
+          resources:
+            requests:
+              cpu: 250m
+          livenessProbe:
+            httpGet:
+              path: /liveness
+              port: 9099
+            periodSeconds: 10
+            initialDelaySeconds: 10
+            failureThreshold: 6
+          readinessProbe:
+            httpGet:
+              path: /readiness
+              port: 9099
+            periodSeconds: 10
+          volumeMounts:
+            - mountPath: /lib/modules
+              name: lib-modules
+              readOnly: true
+            - mountPath: /var/run/calico
+              name: var-run-calico
+              readOnly: false
+        - name: install-cni
+          image: quay.io/calico/cni:v1.10.0
+          command: ["/install-cni.sh"]
+          env:
+            - name: CNI_NETWORK_CONFIG
+              valueFrom:
+                configMapKeyRef:
+                  name: calico-config
+                  key: cni_network_config
+            - name: CNI_NET_DIR
+              value: "/etc/kubernetes/cni/net.d"
+            - name: KUBERNETES_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - mountPath: /host/opt/cni/bin
+              name: cni-bin-dir
+            - mountPath: /host/etc/cni/net.d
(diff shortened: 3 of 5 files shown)


cmd/bootkube/render.go
```diff
@@ -50,6 +50,8 @@ var (
 		serviceCIDR         string
 		selfHostKubelet     bool
 		cloudProvider       string
+		networkProvider     string
+		networkMTU          int
```
aaronlevy: Can we get away with a sane default here rather than a flag? I'd really like to avoid this

cmd/bootkube/render.go
```diff
@@ -113,6 +117,12 @@ func validateRenderOpts(cmd *cobra.Command, args []string) error {
 	if renderOpts.apiServers == "" {
 		return errors.New("Missing requried flag: --api-servers")
 	}
+	if renderOpts.networkProvider != asset.NetworkFlannel && renderOpts.networkProvider != asset.NetworkCalico {
+		return errors.New("Must specify --network-provider flannel or calico")
+	}
+	if renderOpts.networkProvider == asset.NetworkCalico && renderOpts.calicoNetworkPolicy {
+		return errors.New("Cannot specify --experimental-calico-network-policy with --network-provider=calico, Calico supports network policy automatically.")
```
aaronlevy: Maybe we just drop this altogether and put this under network provider instead?

`--network-provider={flannel,experimental-calico-network-policy, experimiental-calico-bgp}`?

I'm leaning toward the experimental because if we don't have tests - we could break it at any time.

dghubble: Experimental prefixes work for me. I prefer we say "calico" rather than "calico-bgp", since bgp is redundant. I'd also prefer to keep "canal-style" policy-only projects separate from CNI providers, long term it doesn't really belong here since its not (supposed to be) providing networking or really essential to bringing up a minimal control plane. Suppose some other network policy app comes along. Its an addon.

cmd/bootkube/render.go
dghubble: Agreed, I was on the fence about this. An appropriate default is 1440, but folks really do need to pick the right value depending on the cloud provider or network performance suffers. Since `bootkube render` just produces valid example manifests (not ideal ones), perhaps we leave it to projects to generate assets that use the correct MTU. In my cases, I generate the right assets for the platform separately, so I'm not opinionated which way `bootkube render` goes.

dghubble:
ok to test

klausenbusk:
Should we add a brief pros/cons (/use-case) of flannel vs calico? For a newcomer I think it can be a bit hard to know what is "best".

dghubble:
coreosbot run e2e calico

dghubble:
coreosbot run e2e

dghubble:
The yellow "jenkins/job" is the old name for my new job "e2e-calico".

dghubble:
@klausenbusk to a degree, though maybe we just link to the relevant projects. So kubernetes-incubator projects stay clear of providing recommendations and leave it to users or to higher level tools to choose what's right for them.

Practically speaking, this Calico support in bootkube is experimental so flannel is still the supported option.

cmd/bootkube/render.go
dghubble: Yeah?

cmd/bootkube/render.go
dghubble: `--network-provider={flannel,experimental-calico}`

dghubble:
Re-enabling broken tests (#716) finds some issues in this PR :)

cmd/bootkube/render.go
dghubble: Removed. flannel and Calico use different defaults, so setting it in one place breaks tests anyway. This means `bootkube render` will create clusters that work using the default settings, but aren't necessarily getting optimal throughput. I think that's ok given that higher level tools do handle these cases.

dghubble:
Squashed and rebased

cmd/bootkube/render.go
```diff
@@ -72,6 +73,7 @@ func init() {
 	cmdRender.Flags().StringVar(&renderOpts.serviceCIDR, "service-cidr", "10.3.0.0/24", "The CIDR range of cluster services.")
 	cmdRender.Flags().BoolVar(&renderOpts.selfHostKubelet, "experimental-self-hosted-kubelet", false, "(Experimental) Create a self-hosted kubelet daemonset.")
 	cmdRender.Flags().StringVar(&renderOpts.cloudProvider, "cloud-provider", "", "The provider for cloud services.  Empty string for no provider")
+	cmdRender.Flags().StringVar(&renderOpts.networkProvider, "network-provider", "flannel", "CNI network provider (flannel or experimental-calico).")
 	cmdRender.Flags().BoolVar(&renderOpts.selfHostedEtcd, "experimental-self-hosted-etcd", false, "(Experimental) Create self-hosted etcd assets.")
 	cmdRender.Flags().BoolVar(&renderOpts.calicoNetworkPolicy, "experimental-calico-network-policy", false, "(Experimental) Add network policy support by calico.")
```
aaronlevy: We should drop this right?

aaronlevy: Or really this should be moved under `networkProvider={flannel, experimental-calico, experimental-canal}`?

dghubble: We could do this and change the testing setup/scripts to work that way (Jenkins jobs pass enable_network_policy to terraform to scripts to bootkube render so the jobs need changes). It may be worth a separate PR.

Concerning whether we *should*, imo flannel with policy-only Calico (i.e. "canal") probably shouldn't  be conflated with a top level CNI network provider. Policy-only Calico manifests can be deployed on top of a flannel cluster afterward, as an addon, no reason it needs to be part of control plane bootstrapping or a bootkube concern. I'd prefer we eventually remove `--experimental-calico-network-policy` and just point to the upstream plain-old-manifests that work fine. Its the same logic that governs why bootkube clusters don't come with Ingress or CLUO or other optional addons.

aaronlevy: I'd never recommend installing network policy after the fact (even if you could probably make it work). Essentially you'd end up relying on the name ordering of CNI configuration, (and that already makes me nervous), as well as deployed CNI plugin versions / compatibility. Basically -- not worth the risk (nor would we try and support that here). Networking IMO should only be considered an install-time choice.

To that end - I would much prefer that we have a single flag for the network provider rather than separating those concepts. 

What about a separate PR (that comes before this) that changes the flag from:

`--experimental-calico-network-policy=true`

to

`--network-provider=experimental-canal`

Then this as a follow-up is merely adding a new provider?

Might also consider a minor version bump for this given the flag change. But given that it's experimental (and breaking changes should be expected), I could go either way.

dghubble: > I'd never recommend installing network policy after the fact... not worth the risk

Calico policy-only manifests are jumbled up with control plane manifests and get created in random order. Their install-cni sidecar races with flannel's sidecar. Creating these manifests early shouldn't provide any sense of safety. In my view, *any* of these setups that involve flannel with policy-only calico are racy.

> nor would we try and support that here

If anything, I want the opposite. Less support responsibility for bootkube. I'd prefer we point to upstream to (try to) handle policy-only mode since their docs cover adding it to a flannel cluster. Elevating it to a top level network provider in bootkube means we continue supporting it. That's fine if we wanna keep trying, its just not a mode I'd encourage users to choose myself.

dghubble:
Basically just gonna re-do this

dghubble:
Force override old PR.

dghubble:
coreosbot run e2e

hack/quickstart/init-master.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -euox pipefail
```
aaronlevy: Do we want to leave the `-x` option, or was that during debugging? Or if it's generally useful should we turn it on for `init-node.sh` also?

dghubble: It was for debugging, I'll remove, though its a bit useful

dghubble:
Failed was related to the Jenkins testing log collection setup, which was different on the e2e-calico test.

dghubble:
coreosbot run e2e

dghubble:
coreosbot run e2e

dghubble:
coreosbot run e2e calico

pkg/asset/internal/templates.go
```diff
@@ -1225,6 +1225,130 @@ spec:
...
+            periodSeconds: 10
+          volumeMounts:
+            - mountPath: /lib/modules
+              name: lib-modules
+              readOnly: true
+            - mountPath: /var/run/calico
+              name: var-run-calico
+              readOnly: false
+        - name: install-cni
+          image: quay.io/calico/cni:v1.10.0
```
ozdanborne: I think this should this be `{{ .Images.CalicoCNI }}`

dghubble: Should have been addressed by https://github.com/kubernetes-incubator/bootkube/pull/726

PR #720, opened 2017-09-26
Documentation: Refresh the cluster upgrade guide

### LINKED ISSUE #673: Upgrade instructions need to be updated
The [upgrade instructions](https://github.com/kubernetes-incubator/bootkube/blob/master/Documentation/upgrading.md) are a bit dated at this point.

They still cover the self-hosted kubelet which is experimental, and basically unsupported at this point. As well as describes updating daemonsets before they supported rolling updates natively.

We should update the instructions to cover current state of daemosnet updates, and how to update the kubelet when it is not self-hosted (e.g. modify the kubelet.service file and point to a new version).

Could get some inspiration from matchbox's updated instructions: https://github.com/coreos/matchbox/blob/master/Documentation/bootkube-upgrades.md

dghubble:
Upgrade docs taken from https://github.com/coreos/matchbox/blob/master/Documentation/bootkube-upgrades.md, but polished a bit for kubernetes-incubator/bootkube.

Closes #673

### DIFF
--- a/Documentation/upgrading.md
+++ b/Documentation/upgrading.md
@@ -1,173 +1,163 @@
+# Upgrading self-hosted Kubernetes
 
-# Upgrading Self-hosted Kubernetes
+"Self-hosted" Kubernetes clusters run the apiserver, scheduler, controller-manager, kube-dns, kube-proxy, and flannel or calico as pods, like ordinary applications. This allows upgrades to be performed in-place using (mostly) `kubectl`, as an alternative to re-provisioning.
 
-Self-hosted Kubernetes clusters schedule Kubernetes components such as the apiserver, kubelet, scheduler, and controller-manager as pods like other applications (except with node selectors). This allows Kubernetes level operations to be performed to upgrade clusters in place, rather than by re-provisioning.
+Let's upgrade a Kubernetes v1.6.6 cluster to v1.6.7 as an example.
 
-Let's upgrade a self-hosted Kubernetes v1.4.1 cluster to v1.4.3 as an example.
+## Status
+
+The process of in-place upgrading a self-hosted Kubernetes cluster can be straight-forward, but there may be complex underlying changes that affect your cluster. In most cases, patch version upgrades (e.g. v1.6.6 to v1.6.7) are safe.
+
+Before beginning an upgrade, it is recommended you evaluate if an in-place upgrade is appropriate for your Kubernetes cluster and risk tolerance. You may wish to test the in-place upgrade on a development cluster, failover important workloads to another cluster, or be prepared to handle unforseen issues.
+
+## Prepare
+
+Find the diff between bootkube assets generated for the existing cluster version and the desired version. This depends on the tool used to generate assets:
+
+* Github [compare](https://github.com/kubernetes-incubator/bootkube/compare/v0.5.0...v0.5.1) changes between the existing and desired versions and infer the appropriate cluster changes.
+* [bootkube render](https://github.com/kubernetes-incubator/bootkube) - Install the `bootkube` binaries for the existing and desired versions. Render assets to different locations with each binary and diff the assets.
+* [External Tools](users-integrations.md) - Check the docs for the external tool and compare assets generated for each version.
+
+In simple cases, you may only need to bump the version of a few images. In more complex cases, there may be entirely new components, configuration, or flags.
 
 ## Inspect
 
-Show the control plane daemonsets and deployments which will need to be updated.
-
-    $ kubectl get daemonsets -n=kube-system
-    NAME             DESIRED   CURRENT   NODE-SELECTOR                    AGE
-    kube-apiserver   1         1         node-role.kubernetes.io/master   5m
-    kube-proxy       3         3         <none>                           5m
-    kubelet          3         3         <none>                           5m
-
-    $ kubectl get deployments -n=kube-system
-    NAME                      DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
-    kube-controller-manager   1         1         1            1           5m
-    kube-dns-v20              1         1         1            1           5m
-    kube-scheduler            1         1         1            1           5m
-
 Check the current Kubernetes version.
 
-    $ kubectl version
-    Client Version: version.Info{Major:"1", Minor:"4", GitVersion:"v1.4.0", GitCommit:"a16c0a7f71a6f93c7e0f222d961f4675cd97a46b", GitTreeState:"clean", BuildDate:"2016-09-26T18:16:57Z", GoVersion:"go1.6.3", Compiler:"gc", Platform:"linux/amd64"}
-    Server Version: version.Info{Major:"1", Minor:"4", GitVersion:"v1.4.1+coreos.0", GitCommit:"b7a02f46b972c5211e5c04fdb1d5b86ac16c00eb", GitTreeState:"clean", BuildDate:"2016-10-11T20:13:55Z", GoVersion:"go1.6.3", Compiler:"gc", Platform:"linux/amd64"}
+```sh
+$ kubectl version
+Client Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.2", GitCommit:"477efc3cbe6a7effca06bd1452fa356e2201e1ee", GitTreeState:"clean", BuildDate:"2017-04-19T20:33:11Z", GoVersion:"go1.7.5", Compiler:"gc", Platform:"linux/amd64"}
+Server Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.6+coreos.1", GitCommit:"42a5c8b99c994a51d9ceaed5d0254f177e97d419", GitTreeState:"clean", BuildDate:"2017-06-21T01:10:07Z", GoVersion:"go1.7.6", Compiler:"gc", Platform:"linux/amd64"}
+```
 
-In this case, Kubernetes is `v1.4.1+coreos.0` and our goal is to upgrade to `v1.4.3+coreos.0`. First, update the control plane pods. Then the kubelets and proxies on all nodes.
+```sh
+$ kubectl get nodes
+NAME                               STATUS    AGE       VERSION
+node1.example.com                  Ready     21d       v1.6.6+coreos.1
+node2.example.com                  Ready     21d       v1.6.6+coreos.1
+node3.example.com                  Ready     21d       v1.6.6+coreos.1
+node4.example.com                  Ready     21d       v1.6.6+coreos.1
+```
 
 ## Control Plane
 
+Show the control plane DaemonSets and Deployments that will need to be updated.
+
+```sh
+$ kubectl get daemonsets -n=kube-system
+NAME                             DESIRED   CURRENT   READY     UP-TO-DATE   AVAILABLE   NODE-SELECTOR                     AGE
+kube-apiserver                   1         1         1         1            1           node-role.kubernetes.io/master=   21d
+kube-flannel                     4         4         4         4            4           <none>                            21d
+kube-proxy                       4         4         4         4            4           <none>                            21d
+pod-checkpointer                 1         1         1         1            1           node-role.kubernetes.io/master=   21d
+
+$ kubectl get deployments -n=kube-system
+kube-controller-manager           2         2         2            2           21d
+kube-dns                          1         1         1            1           21d
+kube-scheduler                    2         2         2            2           21d
+```
+
 ### kube-apiserver
 
-Edit the kube-apiserver daemonset. Change the container image name to `quay.io/coreos/hyperkube:v1.4.3_coreos.0`.
+If only the container image version has changed, update the image with a single command.
 
-    $ kubectl edit daemonset kube-apiserver -n=kube-system
+```
+kubectl set image daemonset kube-apiserver kube-apiserver=quay.io/coreos/hyperkube:v1.6.7_coreos.0
+```
 
-Since daemonsets don't yet support rolling, manually delete each apiserver one by one and wait for each to be re-scheduled.
+You can edit the daemonset directly if other changes are needed.
 
-    $ kubectl get pods -n=kube-system
-    # WARNING: Self-hosted Kubernetes is still new and this may fail
-    $ kubectl delete pod kube-apiserver-s62kb -n=kube-system
+```sh
+$ kubectl edit daemonset kube-apiserver -n=kube-system
+```
 
-If you only have one, your cluster will be temporarily unavailable. Remember the Hyperkube image is quite large and this can take a minute.
-
-    kubectl get pods -n=kube-system
-    NAME                                       READY     STATUS    RESTARTS   AGE
-    kube-apiserver-vyg3t                       2/2       Running   0          2m
-    kube-controller-manager-1510822774-qebia   1/1       Running   2          12m
-    kube-dns-v20-3531996453-0tlv9              3/3       Running   0          12m
-    kube-proxy-8jthl                           1/1       Running   0          12m
-    kube-proxy-bnvgy                           1/1       Running   0          12m
-    kube-proxy-gkyx8                           1/1       Running   0          12m
-    kube-scheduler-2099299605-67ezp            1/1       Running   2          12m
-    kubelet-exe5k                              1/1       Running   0          12m
-    kubelet-p3g98                              1/1       Running   0          12m
-    kubelet-quhhg                              1/1       Running   0          12m
-    pod-checkpointer-node1.example.com         1/1       Running   0          12m
+With only one apiserver, the cluster may be momentarily unavailable.
 
 ### kube-scheduler
 
-Edit the scheduler deployment to rolling update the scheduler. Change the container image name for the hyperkube.
+Again, if only the container image version has changed, update the image with a single command.
 
-    kubectl edit deployments kube-scheduler -n=kube-system
+```
+kubectl set image deployment kube-scheduler kube-scheduler=quay.io/coreos/hyperkube:v1.6.7_coreos.0
+```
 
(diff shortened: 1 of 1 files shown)


dghubble:
coreosbot test e2e

dghubble:
coreosbot run e2e

Documentation/upgrading.md
```diff
@@ -1,173 +1,143 @@
...
+"Self-hosted" Kubernetes clusters run the apiserver, scheduler, controller-manager, kube-dns, kube-proxy, and flannel or calico as pods, like ordinary applications (except with taint tolerations). This allows upgrades to be performed in-place using (mostly) `kubectl` as an alternative to re-provisioning.
 
-Self-hosted Kubernetes clusters schedule Kubernetes components such as the apiserver, kubelet, scheduler, and controller-manager as pods like other applications (except with node selectors). This allows Kubernetes level operations to be performed to upgrade clusters in place, rather than by re-provisioning.
+Let's upgrade a Kubernetes v1.6.6 cluster to v1.6.7 as an example.
 
-Let's upgrade a self-hosted Kubernetes v1.4.1 cluster to v1.4.3 as an example.
+## Status
 
-## Inspect
+bootkube does not provide in-place patch diffs or migrations between releases. In-place upgrades are inherently more dangerous than a blue/green cluster deployments. Evaluate whether an in-place upgrade is appropriate for your Kubernetes cluster and risk tolerance before beginning. Failover important workloads to another cluster in advance.
```
aaronlevy: I feel like this might be a little unnecessarily scary. In most cases, even if the control plane is down the workloads are still running fine. Rather than this, I feel like we should have a section on rollbacks incase of failure scenarios (difficult to assume, enumerate).

Documentation/upgrading.md
```diff
@@ -1,173 +1,143 @@
...
 
-Edit the kube-apiserver daemonset. Change the container image name to `quay.io/coreos/hyperkube:v1.4.3_coreos.0`.
+Show the control plane daemonsets and deployments which will need to be updated.
 
-    $ kubectl edit daemonset kube-apiserver -n=kube-system
+```sh
+$ kubectl get daemonsets -n=kube-system
+NAME                             DESIRED   CURRENT   READY     UP-TO-DATE   AVAILABLE   NODE-SELECTOR                     AGE
+kube-apiserver                   1         1         1         1            1           node-role.kubernetes.io/master=   21d
+kube-etcd-network-checkpointer   1         1         1         1            1           node-role.kubernetes.io/master=   21d
```
aaronlevy: Can we drop this in the example, self-hosted etcd is alpha/experimental - we don't need to necessarily document upgrades for it.

Documentation/upgrading.md
```diff
@@ -1,173 +1,143 @@
...
 
-    $ kubectl get pods -n=kube-system
-    # WARNING: Self-hosted Kubernetes is still new and this may fail
-    $ kubectl delete pod kube-apiserver-s62kb -n=kube-system
+Edit the `kube-apiserver` daemonset to rolling update the apiserver.
 
-If you only have one, your cluster will be temporarily unavailable. Remember the Hyperkube image is quite large and this can take a minute.
+```sh
+$ kubectl edit daemonset kube-apiserver -n=kube-system
+```
```
aaronlevy: Might be useful to document just bumping the image if no other changes are needed:

```
kubectl set image daemonset/kube-apiserver kube-apiserver=quay.io/coreos/hyperkube:v1.6.7_coreos.0
```

Documentation/upgrading.md
```diff
@@ -1,173 +1,143 @@
...
-The on-host kubelet is started by systemd, and is used to launch, and be replaced by, the self-hosted kubelet. While the on-host kubelet is short-lived, it should be kept up to date with the self hosted components.
+## kubelet
 
-On each host running a kubelet, modify the kubelet systemd unit to reference a new `KUBELET_IMAGE_TAG` (an example snippet is below)
+SSH to each node and update `/etc/kubernetes/kubelet.env`. Restart the `kubelet.service`.
 
-`/etc/systemd/system/kubelet.service`
+```sh
+ssh core@node1.example.com
+sudo vim /etc/kubernetes/kubelet.env
```
aaronlevy: This doesn't exist in our hack examples. Instead we directly set the env vars in the service: https://github.com/kubernetes-incubator/bootkube/blob/master/hack/quickstart/kubelet.master#L2-L3

dghubble: SSH to each node and update the `KUBELET_IMAGE_TAG` in `kubelet.service` or `/etc/kubernetes/kubelet.env`, depending on the provisioning tool used. Restart the `kubelet.service`.

Documentation/upgrading.md
dghubble: Added

Documentation/upgrading.md
dghubble: Removed

Documentation/upgrading.md
dghubble: Any specific wording changes you'd suggest? I agree with the desire to have more robust docs about how to recover from each possible sticky situation. I don't have those docs.

I really want folks to weight whether the risks of in-place upgrades are tolerable for their cluster, since bootkube itself isn't providing guarantees if something goes wrong.

To an admin responsible for keeping production services running, creating a new cluster, deploying the same workloads, verifying services work properly, migrating traffic, and later (once you're sure) tearing down the old cluster is the responsible way to go. At least from the "don't get fired" perspective. 

For other situations, editing the control plane may be perfectly tolerable, especially if its a simple diff. Still, it is more dangerous. Its easy to typo something and end up with an inaccessible control plane. Even if "workloads are running fine", the cluster is not properly tolerating node failures, shifting pods, adding new workloads, etc.

Documentation/upgrading.md
```diff
@@ -1,173 +1,149 @@
+# Upgrading self-hosted Kubernetes
 
-# Upgrading Self-hosted Kubernetes
+"Self-hosted" Kubernetes clusters run the apiserver, scheduler, controller-manager, kube-dns, kube-proxy, and flannel or calico as pods, like ordinary applications (except with taint tolerations). This allows upgrades to be performed in-place using (mostly) `kubectl` as an alternative to re-provisioning.
```
aaronlevy: nit: maybe drop "(except with taint tolerations)" -- I would still consider that part of an ordinary application, just another scheduling mechanic.

aaronlevy: Oh just saw we had similar wording in the deleted section. Still think it's extraneous :)

Documentation/upgrading.md
```diff
@@ -1,173 +1,149 @@
...
-    $ kubectl version
-    Client Version: version.Info{Major:"1", Minor:"4", GitVersion:"v1.4.0", GitCommit:"a16c0a7f71a6f93c7e0f222d961f4675cd97a46b", GitTreeState:"clean", BuildDate:"2016-09-26T18:16:57Z", GoVersion:"go1.6.3", Compiler:"gc", Platform:"linux/amd64"}
-    Server Version: version.Info{Major:"1", Minor:"4", GitVersion:"v1.4.1+coreos.0", GitCommit:"b7a02f46b972c5211e5c04fdb1d5b86ac16c00eb", GitTreeState:"clean", BuildDate:"2016-10-11T20:13:55Z", GoVersion:"go1.6.3", Compiler:"gc", Platform:"linux/amd64"}
-
-In this case, Kubernetes is `v1.4.1+coreos.0` and our goal is to upgrade to `v1.4.3+coreos.0`. First, update the control plane pods. Then the kubelets and proxies on all nodes.
-
-## Control Plane
-
-### kube-apiserver
+bootkube does not provide in-place patch diffs or migrations between releases. In-place upgrades are inherently more dangerous than a blue/green cluster deployments. Evaluate whether an in-place upgrade is appropriate for your Kubernetes cluster and risk tolerance before beginning. Failover important workloads to another cluster in advance.
```
aaronlevy: Maybe something like:

> While the process of upgrading a self-hosted cluster can be relatively straight-forward, there may be underlying functionality changes that could affect your cluster. 

> In most cases patch version upgrades (e.g. `v1.6.6` to `v1.6.7`) are safe to apply in-place. 

>For minor version upgrades (e.g. `v1.6.7` to `v1.7.0`) it is important to evaluate if an in-place upgrade is appropriate for your Kubernetes cluster. A recommendation would be to first test the in-place upgrade on a development cluster that uses the same configurations.

Documentation/upgrading.md
```diff
@@ -1,173 +1,149 @@
...
-Edit the kube-apiserver daemonset. Change the container image name to `quay.io/coreos/hyperkube:v1.4.3_coreos.0`.
+## Prepare
 
-    $ kubectl edit daemonset kube-apiserver -n=kube-system
+Find the diff between bootkube assets between the existing cluster version and the desired cluster version. This depends on the tool used to generate assets:
 
-Since daemonsets don't yet support rolling, manually delete each apiserver one by one and wait for each to be re-scheduled.
+* Manual - Check Github [compare](https://github.com/kubernetes-incubator/bootkube/compare/v0.6.0...v0.6.1) between existing and diesred versions and infer the appropriate cluster changes
+* [bootkube render](https://github.com/kubernetes-incubator/bootkube) - install the `bootkube` binaries for the existing and desired versions. Render assets to different locations with each binary and diff the assets.
+* [poseidon/bootkube-terraform](https://github.com/poseidon/bootkube-terraform) - checkout the tag for the existing and desired versions. Run `terraform apply` at each tag to render assets to different locations and diff the assets.
```
aaronlevy: We probably shouldn't single out particular external tool here. What if we had an entry for:

* [External Integrations](https://github.com/kubernetes-incubator/bootkube/blob/master/Documentation/users-integrations.md) - If you used an external tool to provision your cluster, please see their respective documentation for upgrade instructions.

Documentation/upgrading.md
```diff
@@ -1,173 +1,149 @@
...
+bootkube does not provide in-place patch diffs or migrations between releases. In-place upgrades are inherently more dangerous than a blue/green cluster deployments. Evaluate whether an in-place upgrade is appropriate for your Kubernetes cluster and risk tolerance before beginning. Failover important workloads to another cluster in advance.
 
-Edit the kube-apiserver daemonset. Change the container image name to `quay.io/coreos/hyperkube:v1.4.3_coreos.0`.
+## Prepare
 
-    $ kubectl edit daemonset kube-apiserver -n=kube-system
+Find the diff between bootkube assets between the existing cluster version and the desired cluster version. This depends on the tool used to generate assets:
 
-Since daemonsets don't yet support rolling, manually delete each apiserver one by one and wait for each to be re-scheduled.
+* Manual - Check Github [compare](https://github.com/kubernetes-incubator/bootkube/compare/v0.6.0...v0.6.1) between existing and diesred versions and infer the appropriate cluster changes
```
aaronlevy: Can we use the diff in the example (v1.6.6, v1.6.7)

Documentation/upgrading.md
```diff
@@ -1,173 +1,149 @@
...
+node2.example.com                  Ready     21d       v1.6.6+coreos.1
+node3.example.com                  Ready     21d       v1.6.6+coreos.1
+node4.example.com                  Ready     21d       v1.6.6+coreos.1
+```
 
-Edit the scheduler deployment to rolling update the scheduler. Change the container image name for the hyperkube.
+## Control Plane
 
-    kubectl edit deployments kube-scheduler -n=kube-system
+Show the control plane daemonsets and deployments which will need to be updated.
```
aaronlevy: nit: s/which/that/

Documentation/upgrading.md
```diff
@@ -1,173 +1,149 @@
...
-    kube-dns-v20-3531996453-0tlv9              3/3       Running   0          28m
-    kube-proxy-8jthl                           1/1       Running   0          28m
-    kube-proxy-bnvgy                           1/1       Running   0          28m
-    kube-proxy-gkyx8                           1/1       Running   0          28m
-    kube-scheduler-2255275287-hti6w            1/1       Running   0          6m
-    kubelet-exe5k                              1/1       Running   0          28m
-    kubelet-p3g98                              1/1       Running   0          28m
-    kubelet-quhhg                              1/1       Running   0          28m
-    pod-checkpointer-node1.example.com         1/1       Running   0          28m
+If only the container image in a manifest has changed, update it.
```
aaronlevy: Let's put this note and "image only" option before the edit command option (so we start with simple option, then move to more detailed options).

So maybe a bit more description on the note as well

> If you have determined that the only change is the container image version (and not configuration, or flags, for example), then you can simply update the container image with a single command. This is more common in the case of patch version upgrades.

Then for the note about editing, maybe something like:

> If you are changing more of the manifest than just the image version, you can make the edits to the object directly

Documentation/upgrading.md
```diff
@@ -1,173 +1,149 @@
...
+kubectl set image daemonset kube-apiserver kube-apiserver=quay.io/coreos/hyperkube:v1.6.7_coreos.0
+```
 
-At this point, the control plane components have been upgraded to v1.4.3.
+If you only have one apiserver, the cluster may be momentarily unavailable.
 
-    $ kubectl version
-    Client Version: version.Info{Major:"1", Minor:"4", GitVersion:"v1.4.0", GitCommit:"a16c0a7f71a6f93c7e0f222d961f4675cd97a46b", GitTreeState:"clean", BuildDate:"2016-09-26T18:16:57Z", GoVersion:"go1.6.3", Compiler:"gc", Platform:"linux/amd64"}
-    Server Version: version.Info{Major:"1", Minor:"4", GitVersion:"v1.4.3+coreos.0", GitCommit:"7819c84f25e8c661321ee80d6b9fa5f4ff06676f", GitTreeState:"clean", BuildDate:"2016-10-17T21:19:17Z", GoVersion:"go1.6.3", Compiler:"gc", Platform:"linux/amd64"}
+### kube-scheduler
```
aaronlevy: Lets add the same image-only option for kube-scheduler and kube-controller-manager here too (but we can have an abridged note: "Again, if we are only updating the image, we can use the set image command:"

Documentation/upgrading.md
dghubble: True, matchbox calls only calls these out because these are the two tools used in Matchbox examples, for non-terraform and terraform examples respectively. I ported the docs from there.

Will change.

Documentation/upgrading.md
dghubble: So that would be v0.5.0...v0.5.1 if we wanna correspond to the Kubernetes v1.6.6 to v1.6.7 updates -  its comparing the assets produced at different bootkube versions corresponding to Kubernetes versions.

Documentation/upgrading.md
dghubble: agreed, its not very important for the person getting a high level summary

Documentation/upgrading.md
dghubble: The optimistic lead in sounds fine to me.

I'd still hope to recommend people "failover important workloads to another cluster in advance" or "weight the dangers against blue/green deployments". When managing clusters for teams and sufficiently important apps, there's no amount of "I tested in-place upgrades on a development cluster" that make this safe. At a minimum, there should be a 2nd cluster ready. Otherwise we're going to get panic'd questions like "site is down, how to recover now!".

Documentation/upgrading.md
aaronlevy: oh woops. But yeah - probably can't hurt to compare the actual bootkube versions to the example k8s versions.

Documentation/upgrading.md
dghubble: Updated to something vaguely in between. What do you think?

dghubble:
@aaronlevy Added edits as a 2nd commit so its easier to see. I'll rebase and squash before merge.

Documentation/upgrading.md
```diff
@@ -1,20 +1,22 @@
...
+"Self-hosted" Kubernetes clusters run the apiserver, scheduler, controller-manager, kube-dns, kube-proxy, and flannel or calico as pods, like ordinary applications. This allows upgrades to be performed in-place using (mostly) `kubectl`, as an alternative to re-provisioning.
 
 Let's upgrade a Kubernetes v1.6.6 cluster to v1.6.7 as an example.
 
 ## Status
 
-bootkube does not provide in-place patch diffs or migrations between releases. In-place upgrades are inherently more dangerous than a blue/green cluster deployments. Evaluate whether an in-place upgrade is appropriate for your Kubernetes cluster and risk tolerance before beginning. Failover important workloads to another cluster in advance.
+The process of in-place upgrading a self-hosted Kubernetes cluster can be straight-forward, but there may be complex underlying changes that affect your cluster. In most cases, patch version upgrades (e.g. v1.6.6 to v1.6.7) are safe.
+
+Evaluate if an in-place cluster is appropriate for your Kubernetes cluster and risk tolerance. You may wish to test the in-place upgrade on a development cluster, failover important workloads to another cluster, or be prepared to handle unforseen issues.
```
aaronlevy: I like the change. One nit: "Before beginning an upgrade, is is recommended to evaluate if an in-place..."

Documentation/upgrading.md
```diff
@@ -57,31 +59,43 @@ kube-scheduler                    2         2         2            2           2
 
 ### kube-apiserver
 
-Edit the `kube-apiserver` daemonset to rolling update the apiserver.
+If the only the container image version has changed, update the image with a single command.
```
aaronlevy: s/If the only/If only/

Documentation/upgrading.md
dghubble: fixed

Documentation/upgrading.md
dghubble: fixed

dghubble:
Squashed

PR #722, opened 2017-09-26
pkg/bootkube: Wait for CRD Kind rollout on create

derekparker:
Wait for the CRD kind to fully rollout and register before continuing.
This allows bootkube to also create instances of that kind reliably.


derekparker:
@diegs that's correct, that is the original approach I tried, but the way `Visit` is implemented, it will return an error when it hits that CRD instance object before even invoking the function passed in, which means we would have to retry that entire `Visit` loop again.

pkg/bootkube/create.go
```diff
@@ -144,6 +163,50 @@ func createAssets(manifestDir string) error {
...
+//
+// Example of apiGroup: "tco.coreos.com"
+// Example of version: "v1"
+// Example of namespace: "default"
+// Example of resourceKind: "ChannelOperatorConfig"
+func customResourceDefinitionKindURI(apiGroup, version, namespace, resourceKind string) string {
+	if namespace == "" {
+		namespace = metav1.NamespaceDefault
+	}
+	plural, _ := meta.UnsafeGuessKindToResource(schema.GroupVersionKind{
```
aaronlevy: This is kind of scary description: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/restmapper.go#L145

aaronlevy:
Also, maybe to alleviate some of your concerns @diegs you could still order objects (by name) that would get created before the CRD - if for some reason that was strictly needed. This just pauses when it happens to hit one of them.

pkg/bootkube/create.go
derekparker: Yeah, definitely agreed, however we've been using that function within operator-client (where this is taken from) without issue for a while now. Before that we were using `KindToResource` before it was removed in favor of this scarier named function.

derekparker:
@diegs @aaronlevy updated.

PR #723, opened 2017-09-28
pkg: Move experimental-network-policy to network-provider canal

dghubble:
* Canal is a setup that uses flannel for network connectivity and adds manifests for Calico in policy-only mode (i.e. user supplied networking)
* Rearranges flags as this was said to be desired before merging Calico support #714 

Author's note: I'd recommend users choose `--network-provider=calico` (added in #714) or stick with `--network-provider=flannel`. There are [known issues](#698) with choosing `experimental-canal` (flannel with Calico run in policy-only mode).

### DIFF
--- a/cmd/bootkube/render.go
+++ b/cmd/bootkube/render.go
@@ -50,8 +50,8 @@ var (
 		serviceCIDR         string
 		selfHostKubelet     bool
 		cloudProvider       string
+		networkProvider     string
 		selfHostedEtcd      bool
-		calicoNetworkPolicy bool
 	}
 
 	imageVersions = asset.DefaultImages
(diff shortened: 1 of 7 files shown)


dghubble:
This will fail the e2e-calico test of course. That is expected.

dghubble:
e2e-calico should not be passing.

dghubble:
Oh, I see. In the e2e-calico, we set `export TF_VAR_network_provider="calico"` which propagates through Terraform and scripts, but init-master has an else block. In this case, until #714, e2e-calico is only testing the flannel cluster. Bleh for how we have scripts calling terraform calling more scripts.

PR #731, opened 2017-10-06
Change polling frequency to reduce number of apiserver requests

thorfour:
Secrets and config maps are now pulled whenever the pod has changed or
after a timeout period whichever comes first.

### DIFF
--- a/cmd/checkpoint/main.go
+++ b/cmd/checkpoint/main.go
@@ -57,8 +57,7 @@ const (
 )
 
 var (
-	pollingFrequency time.Duration
-	lockfilePath     string
+	lockfilePath string
 
 	// TODO(yifan): Put these into a struct when necessary.
 	nodeName              string
@@ -68,7 +67,6 @@ var (
 	remoteRuntimeEndpoint string
 	runtimeRequestTimeout time.Duration
 	lastCheckpoint        time.Time
-	checkPointTimeout     time.Duration
 
 	// podSerializer is an encoder for writing checkpointed pods.
 	//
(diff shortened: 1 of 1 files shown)


diegs:
ok to test

cmd/checkpoint/main.go
```diff
@@ -52,9 +52,12 @@ const (
 
 	defaultRuntimeEndpoint       = "unix:///var/run/dockershim.sock"
 	defaultRuntimeRequestTimeout = 2 * time.Minute
+	defaultFrequency             = 3 * time.Second
+	defaultCheckpointTimeout     = 1 * time.Minute
 )
 
 var (
+	frequency    time.Duration
```
diegs: `pollingFrequency`?

cmd/checkpoint/main.go
```diff
@@ -324,59 +331,69 @@ func createCheckpointsForValidParents(client kubernetes.Interface, pods map[stri
...
+
+			_, err = checkpointSecretVolumes(client, pod)
+			if err != nil {
+				//TODO(aaron): This can end up spamming logs at times when api-server is unavailable. To reduce spam
+				//             we could only log error if api-server can't be contacted and existing secret doesn't exist.
+				glog.Errorf("Failed to checkpoint secrets for pod %s: %v", id, err)
+				continue
+			}
+
+			lastCheckpoint = time.Now()
```
diegs: This is a global timer but you're resetting it for the first pod you see, so the loop might not enter for the subsequent pods in this run. Perhaps create a struct to hold {pod, lastCheckpoint} pairs, or reset this after the loop runs?

thorfour: good catch

cmd/checkpoint/main.go
```diff
@@ -86,6 +91,8 @@ func init() {
 	flag.Set("logtostderr", "true")
 	flag.StringVar(&remoteRuntimeEndpoint, "container-runtime-endpoint", defaultRuntimeEndpoint, "[Experimental] The endpoint of remote runtime service. Currently unix socket is supported on Linux, and tcp is supported on windows.  Examples:'unix:///var/run/dockershim.sock', 'tcp://localhost:3735'")
 	flag.DurationVar(&runtimeRequestTimeout, "runtime-request-timeout", defaultRuntimeRequestTimeout, "Timeout of all runtime requests except long running request - pull, logs, exec and attach. When timeout exceeded, kubelet will cancel the request, throw out an error and retry later.")
+	flag.DurationVar(&frequency, "frequency", defaultFrequency, "Rate at which the kubelet and CRI shim is polled for running pods information")
+	flag.DurationVar(&checkPointTimeout, "api-poll-timeout", defaultCheckpointTimeout, "Rate at which the API server is polled for changes to secrets and configmaps")
```
aaronlevy: Are the defaults reasonable enough to not add flags?

thorfour: Since that's subjective I'm not sure that's a good metric. If we would never want someone to be able to modify them, then we maybe should consider removing them as a flag

aaronlevy: I'd say let's drop them unless it becomes clear that users need to be able to modify these

thorfour: Done

cmd/checkpoint/main.go
thorfour: Changed

diegs:

PR #733, opened 2017-10-09
Bumped checkpointer default image

thorfour:


thorfour:
Changes were to reduce api server polling from pod secrets: https://github.com/kubernetes-incubator/bootkube/pull/731

PR #734, opened 2017-10-11
fix system ClusterRoleBinding yaml

janwillies:
rbac is beta in 1.7, not sure if bugfixes to 1.7 are still relevant


dghubble:
This is about to get changed in #729

dghubble:
I don't think this is needed now that bootkube is on Kubernetes 1.8.x, where RBAC is v1.

PR #737, opened 2017-10-17
Reduce test flakiness for Kubernetes v1.8.1

dghubble:
See #729, this is just to try to reduce flakiness


e2e/smoke_test.go
```diff
@@ -2,11 +2,13 @@ package e2e
 
 import (
 	"testing"
+	"time"
 
 	"github.com/coreos/ktestutil/testworkload"
 )
 
 func TestSmoke(t *testing.T) {
+	testworkload.PollTimeoutForNginx = 5 * time.Minute
```
aaronlevy: I think this can be dropped with the ktestutil change

dghubble: Removing

Issue #735, opened 2017-10-13
Deprecate experimental self-hosted Kubelet


dghubble:
Originally, the experimental self-hosted kubelet feature (`--experimental-self-hosted-kubelet`) was an idea to run an on-host kubelet for bootstrapping and pivot to a scheduled, self-hosted kubelet daemonset. Today, the self-hosted Kubelet feature is not tested, not recommended, and not heavily used (or used at all?) in any [higher level projects](https://github.com/kubernetes-incubator/bootkube/blob/master/Documentation/users-integrations.md).

## Proposal

Phase out the `--experimental-self-hosted-kubelet` flag.

1. Rename `--experimental-self-hosted-kubelet` to `--deprecated-self-hosted-kubelet`. This acts as a forcing function so that any users relying on the setting discover a breakage, but can simply rename to continue using the feature and plan to migrate away from it and comment here. (e.g. 0.8.0?)
2. Remove the `--deprecated-self-hosted-kubelet` flag in a later release (e.g 0.9.0?)

## Feedback

If your organization is using this feature or has concerns, please raise them here. Thanks!

Closes #251 #355 
cc @aaronlevy

rockwyc992:
So that, if I want to update kubelet version of my cluster, how should I do?

I think I will update `/etc/kubernetes/kubelet.env` on every node manually, or I will Install new node with new version to replace old one.   

These are not the best way to use Kubernetes. Does there is any operator to help my do this? (Like what container-linux-update-operator do)

dghubble:
Yes, the kubelet should be run as a kubelet.service which may source a
kubelet.env file or set the version in the unit. Edit the version and
restart the service or just replace the node with another that has the
rev'd version.

It's your choice:
1. Configuration management to edit each host
2. Custom operator to edit each host
3. Immutable infrastructure to rolling replace nodes (i.e. new user-data)

I'm not aware of any public operator that will do this for you.

klausenbusk:
> Does there is any operator to help me do this work? (Like what container-linux-update-operator do)
> 2. Custom operator to edit each host

Hmm, that shouldn't be that hard to implement.
[...]

PR #739, opened 2017-10-18
cmd/bootkube: Mark --self-hosted-kubelet deprecated

dghubble:
* Rename self-hosted-kubelet flag to force users to take notice, but provide a window to migrate
* Phase 1 of #735


aaronlevy:
I'm inclined to disable the self-hosted etcd tests

aaronlevy:
I'd like to cut a release, and I think this should be in it. @dghubble any objections to merging this as-is. The etcd tests have nothing to do with this change

dghubble:
Yes, I was hoping it would be included so we can remove it in the following release. :+1:

dghubble:
Self-hosted etcd failure was the same known flake and its unrelated.

PR #742, opened 2017-10-23
Add /lib/modules mount to kube-proxy

### LINKED ISSUE #741: kube-proxy expects to load kernel modules in v1.8.x
```
# kube-proxy
time="2017-10-19T22:42:27Z" level=warning msg="Running modprobe ip_vs failed with message: ``, error: exec: \"modprobe\": executable file not found in $PATH"
time="2017-10-19T22:42:27Z" level=error msg="Could not get ipvs family information from the kernel. It is possible that ipvs is not enabled in your kernel. Native loadbalancing will not work until this is fixed."
```

Mostly benign for now, but should be fixed. The image contains `sbin/modprobe` (which can't be found?) and we do not currently load `/lib/modules` in kube-proxy.

dghubble:
Wait for Kubernetes v1.8.3 and test that this addresses the warning log in #741. This paves the way to use `ip_vs` in kube-proxy in future, but does not enable it for now.

rel: https://github.com/kubernetes/kubernetes/issues/17470, https://github.com/kubernetes/kubernetes/pull/52003
Closes #741

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -706,6 +706,9 @@ spec:
         securityContext:
           privileged: true
         volumeMounts:
+        - mountPath: /lib/modules
+          name: lib-modules
+          readOnly: true
         - mountPath: /etc/ssl/certs
           name: ssl-certs-host
           readOnly: true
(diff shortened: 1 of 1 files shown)


aaronlevy:
I don't think we need this - the kernel modules should be automatically loaded.

aaronlevy:
Oh nm see the linked issue. This is unfortunate. Can we at least set it to read-only?

dghubble:
Yeah, we can set it read-only. I linked to the upstream patch that does the same. The patch adding modprobe back into the image wasn't picked for v1.8.2 so the log warning will still be present. https://github.com/kubernetes/kubernetes/pull/54325

Wait on v1.8.3.

dghubble:
Rebased on #765 for Kubernetes v1.8.3

PR #745, opened 2017-10-24
Switch from quay.io to gcr.io hyperkube image

### LINKED ISSUE #744: Switch to upstream gcr.io hyperkube image
We no longer have a concrete need to use `quay.io/coreos/hyperkube:v1.8.x_coreos.y` over `gcr.io/google_containers/hyperkube:v1.8.x` starting with Kubernetes 1.8. The patches in the quay image are used for the Tectonic setup.

Perhaps we consider switching bootkube to the upstream images?

Pros:
* Faster turn around time to bump images when upstream releases them
* Possibly faster downloads for some users (a prior google groups thread mentioned quay.io vs gcr.io was a determining factor outside the US)
* Closer to "upstream" which may be a goal for a kubernetes-incubator project

Cons:
* The "unknown factor". What if we need to return to patching the image in future?
* Others?

dghubble:
Updated the conformance test job to use kubernetes/kubernetes when launching conformance clusters.

### Note

The rkt flag `--insecure-options=image` isn't as scary as it sounds. rkt had different ideas about image security and originally required GPG signed ACIs. When pulling `docker://` images, those signatures aren't present which is why rkt calls them insecure. In a post-OCI world, it doesn't have so much weight.

Closes #744

### DIFF
--- a/hack/multi-node/user-data.sample
+++ b/hack/multi-node/user-data.sample
@@ -8,15 +8,16 @@ coreos:
       content: |
         [Service]
         EnvironmentFile=/etc/environment
-        Environment=KUBELET_IMAGE_URL=quay.io/coreos/hyperkube
-        Environment=KUBELET_IMAGE_TAG=v1.8.1_coreos.0
+        Environment=KUBELET_IMAGE_URL=docker://gcr.io/google_containers/hyperkube
+        Environment=KUBELET_IMAGE_TAG=v1.8.1
         Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \
           --volume var-lib-cni,kind=host,source=/var/lib/cni \
           --volume opt-cni-bin,kind=host,source=/opt/cni/bin \
           --volume var-log,kind=host,source=/var/log \
           --mount volume=var-log,target=/var/log \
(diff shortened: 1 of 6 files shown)


dghubble:
Looks like workers didn't join, indicating the kubelet didn't start. They started fine when I tried it in https://github.com/coreos/matchbox/pull/665, need to figure out what is different.

dghubble:
Answer: Missing a "\" in RKT_RUN_ARGS in the kubelet.worker systemd unit. Should be fine now.

ericchiang:
Do the multi-node and single-node user data files have coverage in the e2e tests?

dghubble:
No, the e2e tests use the quickstart scripts to create clusters.

dghubble:
Broader context and discussion is in #744

dghubble:
Changed the repo in the conformance tests to use upstream kubernetes/kubernetes when launching conformance clusters.

dghubble:
Unsurprising since its nearly the identical image, but this passes the conformance tests.

```
[70809.588062] golang[5]: [ 4984.206434] golang[5]: Ran 147 of 693 Specs in 4165.177 seconds
[70809.588346] golang[5]: [ 4984.206611] golang[5]: SUCCESS! -- 147 Passed | 0 Failed | 0 Pending | 546 Skipped PASS
```

PR #755, opened 2017-10-28
checkpointer: implement state machine.

diegs:
This implements an explicit state machine to capture the various states
a checkpoint can be in. It recovers those states from disk upon startup
(if applicable) and then reconciles them with the states that are
fetched from the various apiservers.

It also adds 2 new states to attempt to workaround issues with
kubernetes 1.8 in which daemonset pods are deleted before being
recreated. For single-master clusters this can lead to a total outage
during apiserver upgrades since the checkpointer will aggressively
retire the checkpoint for the deleted apiserver pod before it has
a chance to schedule the new one.

### DIFF
--- a/pkg/checkpoint/manifest.go
+++ b/pkg/checkpoint/manifest.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/golang/glog"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -23,6 +24,16 @@ func getFileCheckpoints(path string) map[string]*v1.Pod {
 
 	for _, f := range fi {
 		manifest := filepath.Join(path, f.Name())
+
+		// Check for leftover temporary checkpoints.
+		if strings.HasPrefix(filepath.Base(manifest), ".") {
+			glog.V(4).Infof("Found temporary checkpoint %s, removing.", manifest)
+			if err := os.Remove(manifest); err != nil {
+				glog.V(4).Infof("Error removing temporary checkpoint %s: %v.", manifest, err)
+			}
+			continue
+		}
+
 		b, err := ioutil.ReadFile(manifest)
 		if err != nil {
 			glog.Errorf("Error reading manifest: %v", err)
@@ -71,9 +82,19 @@ func writeManifestIfDifferent(path, name string, data []byte) (bool, error) {
 }
 
 func writeAndAtomicRename(path string, data []byte, perm os.FileMode) error {
-	tmpfile := filepath.Join(filepath.Dir(path), "."+filepath.Base(path))
-	if err := ioutil.WriteFile(tmpfile, data, perm); err != nil {
+	// Ensure that the temporary file is on the same filesystem so that os.Rename() does not error.
+	tmpfile, err := ioutil.TempFile(filepath.Dir(path), ".")
+	if err != nil {
 		return err
 	}
-	return os.Rename(tmpfile, path)
+	if _, err := tmpfile.Write(data); err != nil {
+		return err
+	}
+	if err := tmpfile.Close(); err != nil {
+		return err
+	}
+	if err := tmpfile.Chmod(perm); err != nil {
+		return err
+	}
+	return os.Rename(tmpfile.Name(), path)
 }
--- a/pkg/checkpoint/manifest.go
+++ b/pkg/checkpoint/manifest.go
@@ -52,10 +52,6 @@ func writeCheckpointManifest(pod *v1.Pod) (bool, error) {
 		return false, err
 	}
 	path := filepath.Join(inactiveCheckpointPath, pod.Namespace+"-"+pod.Name+".json")
-	// Make sure the inactive checkpoint path exists.
-	if err := os.MkdirAll(filepath.Dir(path), 0600); err != nil {
-		return false, err
-	}
 	return writeManifestIfDifferent(path, podFullName(pod), buff.Bytes())
 }
 
--- a/pkg/checkpoint/manifest.go
+++ b/pkg/checkpoint/manifest.go
@@ -73,3 +73,11 @@ func writeManifestIfDifferent(path, name string, data []byte) (bool, error) {
 	glog.Infof("Writing manifest for %q to %q", name, path)
 	return true, writeAndAtomicRename(path, data, 0644)
 }
+
+func writeAndAtomicRename(path string, data []byte, perm os.FileMode) error {
+	tmpfile := filepath.Join(filepath.Dir(path), "."+filepath.Base(path))
+	if err := ioutil.WriteFile(tmpfile, data, perm); err != nil {
+		return err
+	}
+	return os.Rename(tmpfile, path)
+}
--- a/pkg/checkpoint/process.go
+++ b/pkg/checkpoint/process.go
@@ -1,6 +1,7 @@
 package checkpoint
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"time"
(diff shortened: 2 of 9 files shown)


diegs:

diegs:
cc @thorfour

diegs:

diegs:
@mxinden i am going to test this against a tectonic cluster this morning.

diegs:

pkg/checkpoint/process.go
```diff
@@ -9,135 +10,539 @@ import (
...
-			if _, ok := apiParentPods[id]; !ok {
-				glog.V(4).Infof("API GC: should remove active checkpoint %s", id)
+	for name, pod := range localParentPods {
+		if _, ok := cs.checkpoints[name]; !ok {
+			cs.checkpoints[name] = &checkpoint{
+				name:  name,
+				state: stateNone{},
+			}
+		}
+		// Always overwrite with the localParentPod since it is a better source of truth.
```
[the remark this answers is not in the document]
diegs: Well, it's using this pod to write to disk so we want to make sure we have the most updated copy of it right?

pkg/checkpoint/process.go
```diff
@@ -9,135 +10,539 @@ import (
...
-	// Can make decisions about starting/stopping checkpoints just with local state.
-	//
-	// If there is an inactive checkpoint, and no parent pod is running, or the checkpoint
-	// is the pod-checkpointer, start the checkpoint.
-	for id := range inactiveCheckpoints {
-		_, ok := localRunningPods[id]
-		if !ok || isPodCheckpointer(inactiveCheckpoints[id], checkpointerPod) {
-			glog.V(4).Infof("Should start checkpoint %s", id)
-			start = append(start, id)
+	// The apiserver pod is deleted, transition to stateInactiveGracePeriod.
```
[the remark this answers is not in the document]
diegs: Done

pkg/checkpoint/process.go
```diff
@@ -9,135 +10,539 @@ import (
...
+
+		// The parent pod exists but the kubelet is not running it, transition to stateActive.
+		return stateActive{}
 	}
 
-	// De-duped checkpoints to remove. If we decide to GC a checkpoint, we will clean up both inactive/active.
-	for k := range removeMap {
-		if k == podCheckpointerID {
-			continue
+	// The apiserver pod is still deleted, remain in stateInactiveGracePeriod.
```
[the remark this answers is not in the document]
diegs: Done

diegs:
@aaronlevy thanks for the comments. Made a few small changes, also testing it against tectonic and it seems to be surviving apiserver destruction nicely

diegs:

pkg/checkpoint/process.go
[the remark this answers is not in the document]
aaronlevy: oh. Duh.

pkg/checkpoint/manifest.go
```diff
@@ -73,3 +69,11 @@ func writeManifestIfDifferent(path, name string, data []byte) (bool, error) {
 	glog.Infof("Writing manifest for %q to %q", name, path)
 	return true, writeAndAtomicRename(path, data, 0644)
 }
+
+func writeAndAtomicRename(path string, data []byte, perm os.FileMode) error {
+	tmpfile := filepath.Join(filepath.Dir(path), "."+filepath.Base(path))
```
yifan-gu: Now I noticed that maybe we should not write the file under the directory.
Even with the prefix of ".", it will still be scanned by `ReadDir()` in `func getFileCheckpoints()`.

diegs: @yifan-gu since the checkpointer is single-threaded `getFileCheckpoints()` should not run concurrently. The only problem is if the program crashes in the middle of `writeAndAtomicRename` and we have a leftover `.foo.json` file sitting there.

I can adjust `getFileCheckpoints()` to ignore files that start with `.`, WDYT?

yifan-gu: I guess my question is why we are not just using a system temp dir here?

pkg/checkpoint/process_test.go
```diff
@@ -272,12 +319,24 @@ func TestProcess(t *testing.T) {
...
+		now = now.Add(checkpointGracePeriod)
+		c.update(tc.localRunning, tc.localParents, tc.apiParents, tc.activeCheckpoints, tc.inactiveCheckpoints, cp)
+		gotGraceStart, gotGraceStop, gotGraceRemove := c.process(now, tc.apiParents != nil, tc.localRunning, tc.localParents, tc.apiParents)
 		if !reflect.DeepEqual(tc.expectStart, gotStart) ||
 			!reflect.DeepEqual(tc.expectStop, gotStop) ||
-			!reflect.DeepEqual(tc.expectRemove, gotRemove) {
-			t.Errorf("For test: %s\nExpected start: %s Got: %s\nExpected stop: %s Got: %s\nExpected remove: %s Got: %s\n",
-				tc.desc, tc.expectStart, gotStart, tc.expectStop, gotStop, tc.expectRemove, gotRemove)
+			!reflect.DeepEqual(tc.expectRemove, gotRemove) ||
+			!reflect.DeepEqual(tc.expectGraceStart, gotGraceStart) ||
```
yifan-gu: nit: Maybe just print them one by one? I mean:
```go
if !reflect.DeepEqual() {
    t.Errorf()
}
if !reflect.DeepEqual() {
    t.Errorf()
}
...
```

pkg/checkpoint/manifest.go
diegs: @yifan-gu Atomic renames are only guaranteed in the same directory.

pkg/checkpoint/process.go
```diff
@@ -9,138 +10,542 @@ import (
...
+
+			glog.Infof("Checkpoint %s transitioning from state %s -> state %s", cp, cp.state, state)
+			cp.state = state
+		}
+	}
+
+	return starts, stops, removes
+}
+
+// stateSelfCheckpoint represents a checkpoint of the checkpointer itself, which has special
```
yifan-gu: maybe define all these states in a separate file? this `process.go` is getting big, I think it would be cleaner to just keep control logics here.

pkg/checkpoint/process.go
```diff
@@ -9,138 +10,542 @@ import (
...
+		return "stop"
+	case remove:
+		return "remove"
+	default:
+		return "[unknown action]"
+	}
+}
+
+// state represents the current state of a checkpoint.
+type state interface {
```
yifan-gu: Was a little confused by the name at first because we also have `apiState`, either change `apiState` to something else like `apiCondition`, or rename this to `podState`?

pkg/checkpoint/process.go
diegs: Done

pkg/checkpoint/process.go
diegs: Done (both)

pkg/checkpoint/process_test.go
diegs: If there are multiple states this can get a little verbose. I prefer to keep it as is.

pkg/checkpoint/manifest.go
diegs: @yifan-gu added some code in `getFileCheckpoints()` to clean up old files that start with `.`

diegs:
@yifan-gu thanks for the review! PTAL

diegs:

s-urbaniak:
Thanks a lot @diegs, first time https://github.com/coreos/tectonic-installer/pull/1938 passed green with this one 👍

pkg/checkpoint/manifest.go
yifan-gu: ok, I didn't found the where said it's only guaranteed in the same directory, but I am fine with the new commit.

diegs: https://godoc.org/os#Rename calls `rename()` says 'OS-specific restrictions may apply when oldpath and newpath are in different directories. If there is an error, it will be of type *LinkError.'

If you look at http://man7.org/linux/man-pages/man2/rename.2.html, `EXDEV` occurs if the old and new locations are on different filesystems (I've seen this happen before). It's safer to keep it in the same directory.

pkg/checkpoint/state.go
```diff
@@ -0,0 +1,337 @@
...
+	transition(time.Time, apiCondition) checkpointState
+	// action returns the action that should be taken for this state.
+	action() action
+}
+
+// stateSelfCheckpoint represents a checkpoint of the checkpointer itself, which has special
+// behavior.
+//
+// stateSelfCheckpoint can transition to stateInactiveGracePeriod.
+type stateSelfCheckpoint struct{}
```
yifan-gu: nit: stateSelfCheckpointActive ?

diegs: Done

diegs:
@yifan-gu thanks, I added a simple test to enforce only allowed transitions can occur. It doesn't actually check behavior (I was relying on the existing test for that) but I think it's a small improvement for now.

diegs:

pkg/checkpoint/manifest.go
rphillips: ioutil.TempFile would be a smart choice to create a temporary file within a certain directory

diegs: @rphillips the only difference would be that it will choose a random rather than deterministic filename, as we would be calling it with the current directory and `.` prefix, and then writing to it... I can make the change anyway though.

rphillips: it isn't a hard blocker... just a suggestion if we need to enumerate those files with a better prefix in the future...

diegs: Eh already changed it.

diegs:

diegs:

PR #760, opened 2017-11-01
Remove --deprecated-self-hosted-kubelet option

dghubble:
Don't merge until bootkube v0.9.0.


dghubble:
I believe with #767 and #777 we'd already said the next version will be v0.9.0 so this should be fine to merge now.

Issue #424, opened 2017-04-11
Build pod-checkpointer images for multiple architectures


aaronlevy:
For kubeadm to make use of the pod-checkpointer, we need to start building container images for upstream supported architectures.

Need a list of the architectures that are required - but opening this issue to track.

/cc @luxas

luxas:
amd64, arm, arm64, ppc64le and s390x are the architectures that are needed

Should be as easy as setting GOARCH really. Please prioritize this as using the checkpointer is a necessary requirement for reliable selfhosting in v1.7 :+1:

luxas:
friendly ping @aaronlevy

PR #762, opened 2017-11-02
Makefile: Cross-compile checkpointer for multiple archs

dghubble:
First part to #424 is cross-compiling checkpointer binary in `make all`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,17 @@ LDFLAGS=-X github.com/kubernetes-incubator/bootkube/pkg/version.Version=$(shell 
 all: \
 	_output/bin/linux/bootkube \
 	_output/bin/darwin/bootkube \
+	_output/bin/linux/checkpoint
+
+cross: \
+	_output/bin/linux/bootkube \
+	_output/bin/darwin/bootkube \
 	_output/bin/linux/checkpoint \
+	_output/bin/linux/amd64/checkpoint \
+	_output/bin/linux/arm/checkpoint \
+	_output/bin/linux/arm64/checkpoint \
+	_output/bin/linux/ppc64le/checkpoint \
+	_output/bin/linux/s390x/checkpoint
 
 release: \
 	clean \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -11,6 +11,16 @@ all: \
 	_output/bin/linux/bootkube \
 	_output/bin/darwin/bootkube \
 	_output/bin/linux/checkpoint \
+	_output/bin/linux/amd64/checkpoint \
+	_output/bin/linux/arm/checkpoint \
+	_output/bin/linux/arm64/checkpoint \
+	_output/bin/linux/ppc64le/checkpoint \
+	_output/bin/linux/s390x/checkpoint
+
+build: \
```
dghubble: Added this because it takes forever for `make all` to cross-compile every flavor and this is handy. We can rearrange these however people like.

Makefile
```diff
@@ -25,9 +35,12 @@ check:
 install: _output/bin/$(LOCAL_OS)/bootkube
 	cp $< $(GOPATH_BIN)
 
+_output/bin/%: GOOS=$(word 1, $(subst /, ,$*))
+_output/bin/%: GOARCH=$(word 2, $(subst /, ,$*))
+_output/bin/%: GOARCH:=amd64
```
dghubble: _output/OS/bootkube, _output/OS/checkpointer, and release tarball continue to default to amd64 to avoid breaking any of the other scripts in the repo or users of the tarball.

aaronlevy: Can you add this as a comment just to keep that clear the intention here.

dghubble:
coreosbot run e2e

dghubble:
Hm, checkpointer flakes? This PR can't be affecting it. @diegs

dghubble:
coreosbot run e2e

Makefile
diegs: It would be super awesome if `all` only built for your arch, and `release` built for all arches. But I am terrible at make and so this is fine if that approach is too hard.

dghubble: There's no difficulty to do that. Currently our "release" (bootkube.tar.gz) only includes:

* bin/linux/bootkube
* bin/darwin/bootkube
* bin/linux/checkpoint

The cross compilation is meant to pave the way for creating binaries for docker images for each architecture.

Maybe `all` should remain as-is to stay fast, and the full list of possible binaries should be in a temporary target like `cross`, at least until we have a target for the docker image. Right now the docker builds are done by a pile of scripts.

diegs: Sure, that's fine. I think in the future if `make release` includes all the arches, then it'd be nicer for `make all` to only include my current arch. Making darwin every time increases my build time by ~33%

Makefile
dghubble: added

dghubble:
coreosbot run e2e

dghubble:
Flakes seem to consistently relate to `FAIL: TestCheckpointerUnscheduleCheckpointer`, if it fails, so do all subsequent tests.

PR #764, opened 2017-11-02
checkpointer: two fixes (now that checkpointer tests work again)

diegs:
checkpointer: make sanitize function idempotent.

The new behavior reads back checkpoints from disk when the checkpointer
restarts, e.g. on node reboot. It was then round-tripping those
checkpoints through the sanitize() function which was not idempotent.
This changes that function to be idempotent.

checkpointer: don't move checkpointer to inactive.

It should move to activeGracePeriod after it is removed. Then it will
clean itself up after the grace period ends.

If it moves to inactive then it will simply get terminated while leaving
garbage around.

cc @thorfour

### DIFF
--- a/pkg/checkpoint/process_test.go
+++ b/pkg/checkpoint/process_test.go
@@ -1,7 +1,6 @@
 package checkpoint
 
 import (
-	"flag"
 	"reflect"
 	"testing"
 	"time"
@@ -11,9 +10,6 @@ import (
 )
 
 func TestProcess(t *testing.T) {
-	flag.Set("logtostderr", "true")
-	flag.Parse()
-
 	type testCase struct {
 		desc                string
 		localRunning        map[string]*v1.Pod
--- a/pkg/checkpoint/process_test.go
+++ b/pkg/checkpoint/process_test.go
@@ -56,6 +56,13 @@ func TestProcess(t *testing.T) {
 			expectStart:         []string{"AA"},
 		},
 		{
+			desc:                "Inactive checkpoint and only api and kubelet parents: should start",
+			inactiveCheckpoints: map[string]*v1.Pod{"AA": {}},
+			apiParents:          map[string]*v1.Pod{"AA": {}},
+			localParents:        map[string]*v1.Pod{"AA": {}},
+			expectStart:         []string{"AA"},
+		},
+		{
 			desc:              "Active checkpoint and no local running: no change",
 			activeCheckpoints: map[string]*v1.Pod{"AA": {}},
 		},
--- a/pkg/checkpoint/process_test.go
+++ b/pkg/checkpoint/process_test.go
@@ -166,8 +166,9 @@ func TestProcess(t *testing.T) {
 				},
 				"AA": {},
 			},
-			expectStop:        []string{"AA", "kube-system/pod-checkpointer"},
-			expectGraceRemove: []string{"AA", "kube-system/pod-checkpointer"},
+			expectStart:       []string{"kube-system/pod-checkpointer"},
+			expectStop:        []string{"BB"},
+			expectGraceRemove: []string{"AA", "BB", "kube-system/pod-checkpointer"},
 		},
 		{
 			desc:         "Inactive pod-checkpointer, no local parent, no api parent: should remove all",
@@ -183,7 +184,7 @@ func TestProcess(t *testing.T) {
 				},
 				"AA": {},
 			},
-			expectStop:        []string{"AA", "kube-system/pod-checkpointer"},
+			expectStart:       []string{"kube-system/pod-checkpointer"},
 			expectGraceRemove: []string{"AA", "kube-system/pod-checkpointer"},
 		},
 		{
@@ -200,7 +201,8 @@ func TestProcess(t *testing.T) {
 				},
 				"AA": {},
 			},
-			expectStop:        []string{"AA", "kube-system/pod-checkpointer"},
+			expectStart:       []string{"kube-system/pod-checkpointer"},
+			expectStop:        []string{"AA"},
 			expectGraceRemove: []string{"AA", "kube-system/pod-checkpointer"},
 		},
 		{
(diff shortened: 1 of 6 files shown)


diegs:

diegs:
I tested this locally using vagrant.

diegs:

diegs:

diegs:

diegs:

diegs:

diegs:

pkg/checkpoint/process_test.go
```diff
@@ -170,8 +166,9 @@ func TestProcess(t *testing.T) {
 				},
 				"AA": {},
 			},
-			expectStop:        []string{"AA", "kube-system/pod-checkpointer"},
-			expectGraceRemove: []string{"AA", "kube-system/pod-checkpointer"},
+			expectStart:       []string{"kube-system/pod-checkpointer"},
+			expectStop:        []string{"BB"},
+			expectGraceRemove: []string{"AA", "BB", "kube-system/pod-checkpointer"},
```
aaronlevy: There's a couple of these, but why would you expect a Start and GraceRemove for pod-checkpointer. And why would a particular pod have multiple states anyway?

diegs: Start is because if we're not stopping the checkpointer, we're starting it (was true in the old code too, but now we have two phases--before and after grace period).

The state transitions are: None (implicit before the test starts) => Active Grace Period (the parent is removed, but keep active for a bit, which leads to a "start" per my previous comment) => Remove (grace period has passed)

I had to reason through the other test changes but they made sense to me too: we always start from "no" state. The fix in my second commit was realizing that the checkpointer needed to keep in active grace period (rather than inactive) when its parent is removed to make sure it can clean up all its state before finally self-destructing.

pkg/checkpoint/process_test.go
```diff
@@ -183,7 +184,7 @@ func TestProcess(t *testing.T) {
 				},
 				"AA": {},
 			},
-			expectStop:        []string{"AA", "kube-system/pod-checkpointer"},
+			expectStart:       []string{"kube-system/pod-checkpointer"},
```
diegs: Reason for change: if the checkpoint isn't being removed we always start it. In this case it starts in active grace period (because the parent is gone) and moves to remove after the grace period ends.

pkg/checkpoint/process_test.go
```diff
@@ -200,7 +201,8 @@ func TestProcess(t *testing.T) {
...
```
diegs: Same reason as above for the checkpointer starting. We still stop AA because we stop a stop-then-remove.

pkg/checkpoint/process_test.go
```diff
@@ -264,8 +266,8 @@ func TestProcess(t *testing.T) {
 				},
 				"AA": {},
 			},
-			expectStop:        []string{"AA", "kube-system/pod-checkpointer"},
-			expectGraceRemove: []string{"AA", "kube-system/pod-checkpointer"},
+			expectStart:       []string{"kube-system/pod-checkpointer", "BB"},
```
diegs: Same as above, the checkpointer stays active during the grace period, and since BB is not locally running we also start its checkpoint as expected (and then promptly remove it).

pkg/checkpoint/process_test.go
```diff
@@ -282,7 +284,7 @@ func TestProcess(t *testing.T) {
 				},
 				"AA": {},
 			},
-			expectStop:        []string{"AA", "kube-system/pod-checkpointer"},
+			expectStart:       []string{"kube-system/pod-checkpointer"},
```
diegs: Same as above. Checkpointer now starts in Active Grace Period.

pkg/checkpoint/process_test.go
```diff
@@ -300,7 +302,8 @@ func TestProcess(t *testing.T) {
...
```
diegs: Same as above, including stopping AA gracefully.

diegs:
e2e-calico failure has nothing to do with this, since it's all checkpointer code...

pkg/checkpoint/process_test.go
aaronlevy: Hmm, it's not running locally, but its parent is -- so we shouldn't start it, right?

diegs: I thought localRunning (from the CRI shim) is the source of truth? The old code didn't actually look at localParent for anything actionable, my understanding is that it's just a cached mirror of apiParents.

yifan-gu:
Why the first results are `gotGrace*`, and the second results are `got*` ? Shouldn't they be swapped?
 
https://github.com/kubernetes-incubator/bootkube/pull/764/files#diff-afbd4361935a86196219850e070ec260R326

diegs:
@yifan-gu I don't follow, gotGrace* are a few lines down:

https://github.com/kubernetes-incubator/bootkube/pull/764/files#diff-afbd4361935a86196219850e070ec260R331

diegs:
coreosbot run e2e

pkg/checkpoint/process_test.go
aaronlevy: Ignore me. We added test case to old code and this is the same behavior

diegs:

diegs:

pkg/checkpoint/process_test.go
yifan-gu: What's the pod-checkpointer's behavior if its manifest is replaced by itself? Would it be restart?

yifan-gu: nvm, this should be fine.

pkg/checkpoint/process_test.go
```diff
@@ -170,8 +173,9 @@ func TestProcess(t *testing.T) {
...
 				"AA": {},
 			},
-			expectStop:        []string{"AA", "kube-system/pod-checkpointer"},
-			expectGraceRemove: []string{"AA", "kube-system/pod-checkpointer"},
+			expectStart:       []string{"kube-system/pod-checkpointer"},
+			expectStop:        []string{"BB"},
+			expectGraceRemove: []string{"AA", "BB", "kube-system/pod-checkpointer"},
 		},
 		{
 			desc:         "Inactive pod-checkpointer, no local parent, no api parent: should remove all",
```
yifan-gu: Is this case possible? For the pod checkpointer to run it will at least be in localRunning right?
But ok to leave as is.

yifan-gu:
LGTM overall, btw have we covered all cases in the test cases?

pkg/checkpoint/process_test.go
diegs: Strictly speaking this shouldn't be possible...

PR #767, opened 2017-11-09
pkg/asset: use service accounts for checkpointer and kube-proxy creds

ericchiang:
Today, both the checkpointer and kube-proxy re-use the kubelet's credentials to talk to the API server. This means we can't target them with specific RBAC profiles or identify them in audit logs. In addition, when we move to TLS bootstrapping (#663), the kubelet's kubeconfig will no longer hold real credentials.

Switch the checkpointer and kube-proxy to use service account credentials instead of the kubelet's kubeconfig.

Extracted from https://github.com/kubernetes-incubator/bootkube/pull/663 to make that PR smaller. 

cc @dghubble @rphillips

### DIFF
--- a/pkg/asset/asset.go
+++ b/pkg/asset/asset.go
@@ -36,7 +36,10 @@ const (
 	AssetPathKubeConfig                  = "auth/kubeconfig"
 	AssetPathManifests                   = "manifests"
 	AssetPathKubelet                     = "manifests/kubelet.yaml"
+	AssetPathKubeConfigInCluster         = "manifests/kubeconfig-in-cluster.yaml"
 	AssetPathProxy                       = "manifests/kube-proxy.yaml"
+	AssetPathProxySA                     = "manifests/kube-proxy-sa.yaml"
+	AssetPathProxyRoleBinding            = "manifests/kube-proxy-role-binding.yaml"
 	AssetPathKubeFlannel                 = "manifests/kube-flannel.yaml"
 	AssetPathKubeFlannelCfg              = "manifests/kube-flannel-cfg.yaml"
 	AssetPathCalico                      = "manifests/calico.yaml"
@@ -60,6 +63,9 @@ const (
 	AssetPathKubeDNSSvc                  = "manifests/kube-dns-svc.yaml"
 	AssetPathSystemNamespace             = "manifests/kube-system-ns.yaml"
 	AssetPathCheckpointer                = "manifests/pod-checkpointer.yaml"
+	AssetPathCheckpointerSA              = "manifests/pod-checkpointer-sa.yaml"
+	AssetPathCheckpointerRole            = "manifests/pod-checkpointer-role.yaml"
+	AssetPathCheckpointerRoleBinding     = "manifests/pod-checkpointer-role-binding.yaml"
 	AssetPathEtcdOperator                = "manifests/etcd-operator.yaml"
 	AssetPathEtcdSvc                     = "manifests/etcd-service.yaml"
 	AssetPathEtcdClientSecret            = "manifests/etcd-client-tls.yaml"
@@ -174,12 +180,11 @@ func NewDefaultAssets(conf Config) (Assets, error) {
 		}
 	}
 
-	// K8S kubeconfig
-	kubeConfig, err := newKubeConfigAsset(as, conf)
+	kubeConfigAssets, err := newKubeConfigAssets(as, conf)
 	if err != nil {
 		return Assets{}, err
 	}
-	as = append(as, kubeConfig)
+	as = append(as, kubeConfigAssets...)
 
 	// K8S APIServer secret
 	apiSecret, err := newAPIServerSecretAsset(as, conf.EtcdUseTLS)
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -390,6 +390,7 @@ spec:
         command:
         - /checkpoint
         - --lock-file=/var/run/lock/pod-checkpointer.lock
+        - --kubeconfig=/etc/checkpointer/kubeconfig
         env:
         - name: NODE_NAME
           valueFrom:
@@ -405,10 +406,13 @@ spec:
               fieldPath: metadata.namespace
         imagePullPolicy: Always
         volumeMounts:
+        - mountPath: /etc/checkpointer
+          name: kubeconfig
         - mountPath: /etc/kubernetes
           name: etc-kubernetes
         - mountPath: /var/run
           name: var-run
+      serviceAccountName: pod-checkpointer
       hostNetwork: true
       nodeSelector:
         node-role.kubernetes.io/master: ""
@@ -418,6 +422,9 @@ spec:
         operator: Exists
         effect: NoSchedule
       volumes:
+      - name: kubeconfig
+        secret:
+          secretName: kubeconfig-in-cluster
       - name: etc-kubernetes
         hostPath:
           path: /etc/kubernetes
@@ -428,6 +435,43 @@ spec:
     rollingUpdate:
       maxUnavailable: 1
     type: RollingUpdate
+`)
+
+var CheckpointerServiceAccount = []byte(`apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: kube-system
+  name: pod-checkpointer
+`)
+
+// TODO: Drop checkpointer RBAC resources to a Role and RoleBinding if
+// the checkpoint switches to only watching kube-system.
+
+var CheckpointerRole = []byte(`apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pod-checkpointer
+rules:
+- apiGroups: [""] # "" indicates the core API group
+  resources: ["pods"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: [""] # "" indicates the core API group
+  resources: ["secrets", "configmaps"]
+  verbs: ["get"]
+`)
+
+var CheckpointerRoleBinding = []byte(`apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: pod-checkpointer
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: pod-checkpointer
+subjects:
+- kind: ServiceAccount
+  name: pod-checkpointer
+  namespace: kube-system
 `)
 
 var ControllerManagerTemplate = []byte(`apiVersion: apps/v1beta2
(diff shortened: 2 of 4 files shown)


aaronlevy:
Can you add a description of why this is needed?

ericchiang:
@aaronlevy added a description.

dghubble:
We don't have any hard rules around this, but I'd aim to put this in a v0.9.0 followup to v0.8.2 since there is a new render asset.

dghubble:
Can we have kube-proxy and checkpointer each use their own named service account instead of a `service-account-kubeconfig.yaml`?

ericchiang:
@dghubble updated with RBAC roles.

pkg/asset/internal/templates.go
```diff
@@ -430,6 +437,40 @@ spec:
...
 
+var CheckpointerServiceAccount = []byte(`apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: kube-system
+  name: pod-checkpointer
+`)
+
+var CheckpointerRole = []byte(`apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
ericchiang: Note: if the pod-checkpointer only watches kube-system (e.g. its own namespace) we can drop this to a Role.

pkg/asset/internal/templates.go
```diff
@@ -729,15 +771,65 @@ spec:
...
+  name: system:node-proxier # Automatically created system role.
+subjects:
+- kind: ServiceAccount
+  name: kube-proxy
+  namespace: kube-system
+`)
+
+// ServiceAccountKubeConfig is used for pods that don't have access to the
+// KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT environment variables.
+// This is used for kube-proxy, since it configures these endpoints itself,
```
dghubble: Its also being used for the pod-checkpointer here. Whats the reason there?

pkg/asset/internal/templates.go
```diff
@@ -729,15 +771,65 @@ spec:
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:node-proxier # Automatically created system role.
+subjects:
+- kind: ServiceAccount
+  name: kube-proxy
+  namespace: kube-system
+`)
+
+// ServiceAccountKubeConfig is used for pods that don't have access to the
```
dghubble: Can you clarify "used for"? This still feels odd and kinda opaque. So before we relied on the host's kubeconfig which contained both endpoint info and credentials. With the service account approach you use an alternate kubeconfig which specifies the endpoint info and the local service account in the pod.

What's the tie in to these environment variables?

```
KUBERNETES_SERVICE_PORT=443
KUBERNETES_PORT=tcp://10.3.0.1:443
```

pkg/asset/asset.go
```diff
@@ -37,6 +37,9 @@ const (
 	AssetPathManifests                   = "manifests"
 	AssetPathKubelet                     = "manifests/kubelet.yaml"
 	AssetPathProxy                       = "manifests/kube-proxy.yaml"
+	AssetPathProxySA                     = "manifests/kube-proxy-sa.yaml"
+	AssetPathProxyRoleBinding            = "manifests/kube-proxy-role-binding.yaml"
+	AssetPathServiceAccountKubeConfig    = "manifests/service-account-kubeconfig.yaml"
```
dghubble: nit: everywhere else, we say the name followed by the type. `KubeconfigServiceAccount` and `kubeconfig-sa.yaml`?

pkg/asset/internal/templates.go
```diff
@@ -729,15 +771,65 @@ spec:
...
+// KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT environment variables.
+// This is used for kube-proxy, since it configures these endpoints itself,
+// and the checkpointer, which is run as a static pod.
+var ServiceAccountKubeConfigTemplate = []byte(`apiVersion: v1
+kind: Secret
+metadata:
+  name: service-account-kubeconfig
```
dghubble: nit: kubeconfig-service-account or just kubeconfig

pkg/asset/internal/templates.go
```diff
@@ -729,15 +771,65 @@ spec:
...
+  namespace: kube-system
+stringData:
```
dghubble: Cool, I missed this getting into upstream.

pkg/asset/internal/templates.go
```diff
@@ -729,15 +771,65 @@ spec:
       - name: ssl-certs-host
         hostPath:
           path: /usr/share/ca-certificates
-      - name: etc-kubernetes
-        hostPath:
-          path: /etc/kubernetes
+        name: ssl-certs-host
```
dghubble: accidental duplicate from a few lines above?

ericchiang: yes, thanks for catching.

pkg/asset/internal/templates.go
ericchiang: Clients compiled with client-go require the `KUBERNETES_SERVICE_PORT` and `KUBERNETES_PORT` to use the in-cluster config.

https://github.com/kubernetes/client-go/blob/kubernetes-1.8.2/rest/config.go#L318-L321

These aren't exposed to kube-proxy (since kube proxy sets up the endpoint) or the checkpointer, since it runs as a static pod.

This work:

```
// ServiceAccountKubeConfig instructs clients to use their service account token, but unlike an in-cluster client
// doesn't rely on the `KUBERNETES_SERVICE_PORT` and `KUBERNETES_PORT` to determine the API
// servers address. This kubeconfig is used bootstrapping pods that might not have access to these env vars,
// such as kube-proxy, which sets up the endpoint (chicken and egg), and the checkpointer, which needs to run
// as a static pod even if the API server isn't available.
```

ericchiang:
@dghubble thanks for the review. Comments addressed.

pkg/asset/asset.go
```diff
@@ -37,6 +37,9 @@ const (
 	AssetPathManifests                   = "manifests"
 	AssetPathKubelet                     = "manifests/kubelet.yaml"
 	AssetPathProxy                       = "manifests/kube-proxy.yaml"
+	AssetPathProxySA                     = "manifests/kube-proxy-sa.yaml"
+	AssetPathProxyRoleBinding            = "manifests/kube-proxy-role-binding.yaml"
+	AssetPathServiceAccountKubeConfig    = "manifests/kubeconfig-sa.yaml"
```
dghubble: Sorry, my nit doesn't help much. Its still confusing to read since this is actually a secret. eh, better idea?

pkg/asset/asset.go
dghubble: See below, got confused. Its not a service account.

pkg/asset/asset.go
dghubble: in-cluster-kubeconfig? which happens to be a secret or configmap?

ericchiang:
e2e passed but it looks like there's something wrong with the Jenkins GitHub plugin https://gist.github.com/ericchiang/1b80ef7c23ed4adf27527e728e04e2d3

pkg/asset/asset.go
ericchiang: Does `manifest/kubeconfig-in-cluster.yaml` work?

dghubble:
Also, `FAIL: TestCheckpointerUnscheduleCheckpointer`. Can you re-run? Possibly with the renamed in-cluster-kubeconfig

pkg/asset/asset.go
dghubble: sgtm

ericchiang:
/assign @diegs 

For final merge.

pkg/asset/internal/templates.go
diegs: I think this is ok to keep for now but add to the checkpointer README and the release note.

pkg/asset/k8s.go
```diff
@@ -40,6 +41,9 @@ func newStaticAssets(imageVersions ImageVersions) Assets {
 		MustCreateAssetFromTemplate(AssetPathControllerManagerDisruption, internal.ControllerManagerDisruptionTemplate, conf),
 		MustCreateAssetFromTemplate(AssetPathKubeDNSDeployment, internal.DNSDeploymentTemplate, conf),
 		MustCreateAssetFromTemplate(AssetPathCheckpointer, internal.CheckpointerTemplate, conf),
+		Asset{AssetPathCheckpointerSA, internal.CheckpointerServiceAccount},
```
diegs: I think it might be more uniform (and less surprising in the future) to pass these through `MustCreateAssetFromTemplate` even if they currently do not do any templating.

ericchiang:
coreosbot run e2e

ericchiang:
Last couple nits addressed.

pkg/asset/internal/templates.go
```diff
@@ -430,6 +437,43 @@ spec:
...
 
+var CheckpointerServiceAccount = []byte(`apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: kube-system
+  name: pod-checkpointer
+`)
+
+// TODO: Drop checkpointer RBAC resources to a Role and RoleBinding if
+// the checkpoint switches to only watching kube-system.
```
aaronlevy: Should we just make this change? It only needs to watch kube-system (also can be follow-up PR)

ericchiang: I can open a separate PR. We'd have to wait the checkpointer change to merged before we drop this to a role, so we can't do it in one PR anyway.

diegs: Why would the checkpointer have to be changed? It just wouldn't have access to non-kube-system right?

ericchiang: Checkpointer attempts to watch all namespaces. If we restrict it now that request will fail.

https://github.com/kubernetes-incubator/bootkube/blob/v0.8.2/pkg/checkpoint/apiserver.go#L18

ericchiang: Opened the checkpointer PR https://github.com/kubernetes-incubator/bootkube/pull/774

pkg/asset/internal/templates.go
```diff
@@ -729,15 +774,68 @@ spec:
...
+// KubeConfigInCluster instructs clients to use their service account token,
+// but unlike an in-cluster client doesn't rely on the `KUBERNETES_SERVICE_PORT`
+// and `KUBERNETES_PORT` to determine the API servers address.
+//
+// This kubeconfig is used by bootstrapping pods that might not have access to
+// these env vars, such as kube-proxy, which sets up the API server endpoint
+// (chicken and egg), and the checkpointer, which needs to run as a static pod
+// even if the API server isn't available.
+var KubeConfigInClusterTemplate = []byte(`apiVersion: v1
+kind: Secret
```
redbaron: Does it need to be a secret? There is no sensitive data in this resource, would you consider merging PR which changes it to ConfigMap?

PR #772, opened 2017-11-14
pkg/asset: Disable Calico termination grace period

dghubble:
* Disable termination grace period to account for Kubernetes v1.8 changes to DaemonSet behavior that can cause rolls to be slow. See https://github.com/projectcalico/calico/pull/1293
* Fix IPIP mode casing https://github.com/projectcalico/calico/pull/1233


aaronlevy:
cc @rithujohn191 (just heads up in case you hadn't seen this)

PR #775, opened 2017-11-16
Make it clear that only Go version 1.8 is supported now

mingzhaodotname:
Only Go version 1.8 is supported for building the project.

### DIFF
--- a/Documentation/development.md
+++ b/Documentation/development.md
@@ -1,5 +1,10 @@
 # Bootkube Development
 
+## Requirements
+
+* Go 1.8+
+* Configured [GOPATH](http://golang.org/doc/code.html#GOPATH)
+
 ## Building
 
 First, clone the repo into the proper location in your $GOPATH:
@@ -9,7 +14,7 @@ go get -u github.com/kubernetes-incubator/bootkube
 cd $GOPATH/src/github.com/kubernetes-incubator/bootkube
 ```
 
-Then, to build:
+Then, to build (only Go verson 1.8 is supported now):
 
 ```
 make clean all


Documentation/development.md
```diff
@@ -9,7 +9,7 @@ go get -u github.com/kubernetes-incubator/bootkube
 cd $GOPATH/src/github.com/kubernetes-incubator/bootkube
 ```
 
-Then, to build:
+Then, to build (only Go verson 1.8 is supported now):
```
dghubble: This will be out of date soon. Can you add something like this at the top?

```
## Requirements

* Go 1.8+
* Configured [GOPATH](http://golang.org/doc/code.html#GOPATH)
```

Builds are still done with 1.8, but we should try to address any issues if it breaks with 1.9.

mingzhaodotname: Thanks a lot for the quick review.

Agree. I tested 1.9 and it breaks. Maybe I can file an issue for that :-)

dghubble:
Actually, can you squash these two commits please?

mingzhaodotname:
Good point. Done :-)

Documentation/development.md
dghubble: Sorry. I should have clarified. Can you keep the Requirements section at the top and remove the modification to this line that says only Go 1.8 is supported? We do intend to support Go 1.8+

Yes, please open an issue for 1.9

dghubble:
One more comment above, thanks for the patience.

Documentation/development.md
mingzhaodotname: Yes, I thought about this. The reason I want to keep it is because 1.9 is not supported NOW. 

Imagine someone comes, and trying to compile it with Go 1.9, and then a new issue or pull request will probably be created :-) 

We can easily removed this when 1.9 is supported. 

Another option is to open an issue, and then refer the issue here by something like this:
"Go version 1.9 will be supported once issue xxx is resolved."

How do you think?

dghubble: We don't need to call out each bug in docs, these notes get lost over time. An issue is fine. If you prefer, change `Go 1.8+` to `Go 1.8` until addressed. While you're at it, can you fix your commit message to match the project style, maybe "Documentation/development.md: Clarify Go requirements"

dghubble:
This is so minor, I'll patch up the comment later.

dghubble:
Thanks!

PR #728, opened 2017-10-04
pkg/assets/internal: remove critical pod annotations

squat:
This PR mirrors the work done in https://github.com/coreos/tectonic-installer/pull/1702 to remove the critical pod annotations for control-plane components.
Question: do we also want to remove these annotations from kube-dns and calico?

cc @diegs cc @aaronlevy

Issue #519, opened 2017-05-17
Remove critical-pod annotations from scheduler & controller-manager


aaronlevy:
If we mark the scheduler as critical, for example, then it happens to get in a state where a pod is unscheduled because the scheduler is down (e.g. https://github.com/kubernetes-incubator/bootkube/issues/112) -- will the rescheduler start evicting other workloads?

This would be pretty dangerous as it would likely just keep evicting workloads even though it will never be scheduled (no scheduler to schedule it).

Ideally the rescheduler would only evict based on "known" scheduling failures (e.g. no more allocatable CPU), not just on "can't schedule" -- but I'm unsure of the exact behavior.

We don't currently deploy the rescheduler, but we should evaluate this behavior before doing so (possibly remove the critical pod annotation from controller-manager / scheduler).

xref: https://github.com/kubernetes-incubator/bootkube/issues/474

luxas:
cc @davidopp @bsalamat @timothysc

bsalamat:
When a critical pod cannot be scheduled, rescheduler tries to find a node for it and removes enough pods from the node to allow scheduling of the critical pod, but once it does so, it [adds the pod to a list of processed pods](https://github.com/kubernetes/contrib/blob/master/rescheduler/rescheduler.go#L174). Once the pod is added to the list, rescheduler will not remove more workloads from other nodes to make it schedule.
That said, after we add priority and preemption to Kubernetes, we will get rid of the way `critical` pods are annotated today and will probably retire rescheduler. In that world, scheduler and other critical system pods will get very high priority and won't get evicted/preempted.

bsalamat:
Actually, when I looked at the code again, I noticed that if the `processed pod` is not scheduled within a certain amount of time, it is removed from the processed list and, therefore, is processed again, which means that more pods may be removed to let it schedule. So, the scenario described in the issue can happen and is dangerous.
As explained above, the priority and preemption work will allow us to retire the existing rescheduler.

dghubble:
This is confusing. https://github.com/kubernetes-incubator/bootkube/pull/728#issuecomment-334240416 states the annotations are only for the rescheduler. https://github.com/kubernetes-incubator/bootkube/issues/474#issuecomment-322922093 states the annotations aren't safe if we *do* run the rescheduler. Upstream salt templates still include the critical-pod annotation, Tectonic does not. @diegs @aaronlevy 

[...]

PR #777, opened 2017-11-17
pkg/assets/internal: Remove critical pod annotations

dghubble:
Supercedes #728 and should remove all critical annotations on the control plane as requested.

Broader discussion and rationale:
* https://github.com/coreos/tectonic-installer/pull/1702
* #728

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -158,7 +158,6 @@ spec:
         k8s-app: kube-apiserver
       annotations:
         checkpointer.alpha.coreos.com/checkpoint: "true"
-        scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
       containers:
       - name: kube-apiserver
@@ -212,8 +211,6 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       tolerations:
-      - key: CriticalAddonsOnly
-        operator: Exists
       - key: node-role.kubernetes.io/master
         operator: Exists
         effect: NoSchedule
@@ -493,8 +490,6 @@ spec:
       labels:
         tier: control-plane
         k8s-app: kube-controller-manager
-      annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
       affinity:
         podAntiAffinity:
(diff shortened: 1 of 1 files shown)


dghubble:
Rebased.

redbaron:
Do I read it right, that annotations are removed just to make rescheduler happier?  It is confusing,  bootkube and rescheduler are different projects, bootkube doesn't install rescheduler into cluster by default, why bootkube is making changes, not rescheduler is improving?

I am concerned, because removed annotations are used by kuberneted itself, when `ExperimentalCriticalPodAnnotation` featuregate is enabled. How one would prevent these very important pods to be evicted by kubelet lets say under disk pressure? Is there a safe way to run self hosted setup in 1.8.x Kubernetes without these annotations at all?

diegs:
@redbaron as discussed in https://github.com/kubernetes-incubator/bootkube/issues/519#issuecomment-322854749, the rescheduler is actually dangerous in self-hosted clusters, and it will be replaced by priorities anyway (see comments from bsalamat in that thread).

Since bootkube has never shipped the rescheduler or enabled experimental features it was deemed prudent to remove the annotations. If you are deciding to run the rescheduler and accept the risks associated with it you are of course welcome to add the annotations to your deployment.

redbaron:
That is what I thought too, I am not running rescheduller.  My point is, that removed unnotations are used by kubernetes itself, not rescheduler.  These anontations are taken into account when kubelet evicts pods when it is under disk space pressure. So these annotations are useful and their removal makes self hosted setup less safe

Quentin-M:
@diegs @dghubble @redbaron @aaronlevy 

The `scheduler.alpha.kubernetes.io/critical-pod` annotation [is used by Kubernetes itself](https://github.com/kubernetes/kubernetes/search?utf8=%E2%9C%93&q=ExperimentalCriticalPodAnnotation&type=) (i.e. eviction, admission, daemonset, preemption controllers) if the `ExperimentalCriticalPodAnnotation` feature flag is enabled, in order to avoid evicting pods, and/or guarantee admission that are a/ static or b/ on resource-pressured nodes - regardless of priorities and whether the rescheduler is used or not.

This [comment and code block](https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/kubelet/eviction/eviction_manager.go#L543-L551) is particularly relevant and shows that without the feature flag and without the annotation, static pods could be evicted, and will not be re-admitted.

PR #784, opened 2017-11-28
e2e: make checkpointer tests more robust.

diegs:
Instead of extending the existing checkpointer daemonset to run on
worker nodes, start a new checkpointer daemonset instead. This makes
sure that the existing checkpointer (on the apiserver) is not disturbed.

Furthermore, this allows the checkpointer tests to work with the new
checkpointer that is only allowed to look at its own namespace.

PR #789, opened 2017-12-05
apiserver: run as non-root user.

diegs:
Changes the apiserver defaults to listen on port 6443 and use a non-root
user. The port matches the upstream default and reflects best practices.

Production users can put the apiserver behind a load-balancer to forward
443 to 6443.

### DIFF
--- a/Documentation/network-requirements.md
+++ b/Documentation/network-requirements.md
@@ -8,7 +8,7 @@ The information below describes a minimum set of port allocations used by Kubern
 
 | Protocol | Port Range | Source                                    | Purpose                |
 -----------|------------|-------------------------------------------|------------------------|
-| TCP      | 443        | Worker Nodes, API Requests, and End-Users | Kubernetes API server. |
+| TCP      | 6443       | Worker Nodes, API Requests, and End-Users | Kubernetes API server. |
 | UDP      | 8472       | Master & Worker Nodes                     | flannel overlay network - *vxlan backend* |
 
 ### etcd node(s) ingress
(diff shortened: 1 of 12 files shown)


rphillips:
though the build is failing

dghubble:
You might have a read through https://github.com/poseidon/terraform-render-bootkube/pull/17 where there was some discussion about the repercussions of this change. I'm not opposed to moving toward it, but it necessitates a v0.10.0 imo.

diegs:
@dghubble yes, v0.10.0 would be prudent, though this is basically just changing the default of an option that has always existed (and updating all the examples to use the new default).

PR tests are very flaky. I'm close to making them better in #784

redbaron:
Do we need to change port at all? Running as non root and  adding `CAP_NET_BIND_SERVICE` capability in security context should be enough

diegs:
I did some research on using `CAP_NET_BIND_SERVICE` in the past. Due to Docker limitations granting it to the user does not work. It must be added to the binary, which would require a modification to the hyperkube image, which upstream will not accept. They recommend using port 6443 which is actually a reasonable stance in my opinion.

redbaron:
Googled,  found your comments :) Can't initcontainer help here? Set capability on hyperkube binary at start time. Hmm, probably  just binding to a non privileged port is easier.

diegs:

diegs:

xiang90:
@diegs 

are all the tests failure tracked? it is useful to link the failure issues in the rerun comment so we know all flakes are tracked. thanks!

diegs:
Flake is #816. Was just doing some testing to see if #814 solves this use case (it does!)

diegs:
coreosbot run e2e

PR #792, opened 2017-12-06
Enable portmap CNI plugin

redbaron:
Makes `hostPort:` setting to work when specified in Pods


weikinhuang:
This is covered in https://github.com/kubernetes-incubator/bootkube/pull/711

PR #797, opened 2017-12-06
Pass ServiceCIDR to controller-manager

redbaron:
In theory PodCIDR and ServiceCIDR can overlap, letting controller-manager
know service cidr allows it to behave correctly in that case

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -407,6 +407,7 @@ spec:
         - --allocate-node-cidrs=true
         - --cloud-provider={{ .CloudProvider }}
         - --cluster-cidr={{ .PodCIDR }}
+        - --service-cluster-ip-range={{ .ServiceCIDR }}
         - --configure-cloud-routes=false
         - --leader-elect=true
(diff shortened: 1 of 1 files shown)


ericchiang:
> In theory PodCIDR and ServiceCIDR can overlap

Do people actually do this? Are there any other benefits from passing this flag or is it just to make this case work?

dghubble:
> behave correctly

How is that defined? Overlapping ranges will surely produce collisions, is the controller trusted detect and use only the non-overlapping parts of the range?

redbaron:
@dghubble , yes it is blocking overlapping part, see https://github.com/kubernetes/kubernetes/blob/v1.8.4/pkg/controller/node/ipam/range_allocator_test.go#L75-L96

redbaron:
> Do people actually do this? Are there any other benefits from passing this flag or is it just to make this case work?

I don't know if people do it or not, but AFAIK it is supported configuration and there is no harm passing extra flag.

dghubble:
coreosbot run e2e

PR #798, opened 2017-12-11
logging: add log gathering scripts

### LINKED ISSUE #783: Improve log collection for e2e tests
- [ ] Organize log artifacts based on which node they we captured from.
- [x] All logs should include timestamps.
- [ ] Stop depending on the fluentd manifests compiled into https://github.com/coreos/ktestutil

We should consider removing fluentd since we still have to SSH to the master to grab them off disk (it's not configured to actively send logs to s3 or something like that). If the master restarts and fails to boot the control plane, fluentd doesn't come back up and we lose all logs for the test.

Consider just SSH'ing to the nodes to fetch logs directly from Docker and journald. Here's a script that does that pretty well.

```bash
#!/bin/bash -e

for CID in $( docker ps -a -q ); do
  if [ "$( docker inspect $CID | jq '.[0].Path' )" == "\"/pause\"" ]; then
    # Ignore pause pods
    continue
  fi

  NAME=$( docker inspect $CID | jq '.[0].Name' -r | sed 's/^\/k8s_//g' )
  echo $NAME
  docker logs -t $CID > "container-$NAME.log" 2>&1
done

for UNIT in $( echo "kubelet docker" ); do
  echo $UNIT
  journalctl \
    -o short-iso \
    --no-hostname \
    -u $UNIT > "service-$UNIT.log"
done
```

rphillips:
fixes #783


ericchiang:
Since the terraform scripts are the ones that run e2e, can we add this there too?

This is the jenkins configuration for those without access

```
#!/bin/bash
set -euo pipefail

#build bootkube
./build/build-release.sh

cd hack/terraform-quickstart

# install terraform
curl -L -O https://releases.hashicorp.com/terraform/0.9.6/terraform_0.9.6_linux_amd64.zip
unzip terraform_0.9.6_linux_amd64.zip
export PATH=$PATH:$PWD

export TF_VAR_access_key_id=$ACCESS_KEY_ID
export TF_VAR_access_key=$ACCESS_KEY
export TF_VAR_instance_tags="bootkube-pr-e2e"
export TF_VAR_resource_owner="bootkube-pr-e2e"
export TF_VAR_ssh_key="jenkins-core-services"
export TF_VAR_additional_masters=0


# these are default values but lets be explicit
export TF_VAR_num_workers=1
export TF_VAR_self_host_etcd="false"
export TF_VAR_region="us-east-1"

# bring up compute
./terraform apply

# sleep so ssh works with start-cluster
sleep 30

#avoid some IPs being blank bootkube/issues/552
./terraform refresh

#launch bootkube via quickstart scripts
./start-cluster.sh

#run tests -- mount /tmp for access to ssh agent 
BOOTKUBE_ROOT=$(git rev-parse --show-toplevel)
sudo rkt run \
    --volume bk,kind=host,source=${BOOTKUBE_ROOT} \
    --mount volume=bk,target=/go/src/github.com/kubernetes-incubator/bootkube \
    --volume tmp,kind=host,source=/tmp \
    --mount volume=tmp,target=/tmp \
    --set-env SSH_AUTH_SOCK=$SSH_AUTH_SOCK \
    --insecure-options=image docker://golang:1.8.3 --exec /bin/bash -- -c \
    "cd /go/src/github.com/kubernetes-incubator/bootkube/e2e && \
    go test -v -timeout 45m \
    --kubeconfig=../hack/quickstart/cluster/auth/kubeconfig \
    --expectedmasters=1 \
    ./e2e/"

```

rphillips:
good idea. will do

ericchiang:
We might also want to add a post-build action step to jenkins.

hack/scripts/copylogs
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
ericchiang: Probably want to name this something vagrant specific?

hack/scripts/copylogs
```diff
@@ -0,0 +1,17 @@
...
+
+set -euo pipefail
+
+LOG_DIR=logs
+
+mkdir -p $LOG_DIR
+
+for m in `vagrant status | grep running | awk '{print $1}'`; do 
+  echo Getting logs for $m
+  vagrant ssh $m -- sudo /tmp/gatherlogs &> /dev/null
```
ericchiang: Read from stdin so we don't have to include the `gatherlogs` script at provision time?

```
vagrant ssh $m -- sudo /bin/bash < gatherlogs
```

ericchiang:
@rphillips maybe add something to the pod-build script like:

```
if [ -f "gatherlogs" ]
then
  ./gatherlogs
fi
```

Then use this PR to test your changes?

README.md
```diff
@@ -86,3 +86,4 @@ You can find us at the bootkube channel on [Kubernetes slack](https://github.com
 ## License
 
 bootkube is under the Apache 2.0 license. See the [LICENSE](LICENSE) file for details.
+
```
ericchiang: ?

ericchiang: Ah nvm, read the commit message.

ericchiang:
@rphillips I don't see anything in the jenkins job to actually run these scripts

hack/scripts/copylogs
ericchiang: @rphillips was this addressed?

rphillips:
@ericchiang i started testing this right before my vacation. I'll get back to this today and this week.

hack/single-node/bootkube-up
```diff
@@ -39,3 +39,6 @@ echo
 echo "Bootstrap complete. Access your kubernetes cluster using:"
 echo "kubectl --kubeconfig=cluster/auth/kubeconfig get nodes"
 echo
+echo "To copy logs to the host run:"
+echo "../scripts/vagrant-copylogs"
```
ericchiang: I don't think we need to add these to the vagrant examples

hack/scripts/copylogs
rphillips: removed the vagrant integration.

hack/single-node/bootkube-up
rphillips: removed

rphillips:
removed the readme tweak, and overall cleaned this up.

rphillips:
@ericchiang is there an option to turn off the reaping?

ericchiang:
Thoughts from IRL conversation.

* Execute the script by piping it into bash via SSH instead of copying the file.
* The kubelet flags are for image collection, not container collection.[0]
* Potentially re-run glide for the deleted code.

Also please squash as you're pushing. Ideally, the commits should be whatever we expect to merge. Something like:

```
e2e: use scripts for gathering logs
e2e: remove unused log gathering package
vendor: glide update
``` 

[0] https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/#container-collection

rphillips:
@ericchiang all your comments addressed

rphillips:
coreosbot run e2e

hack/multi-node/Vagrantfile
```diff
@@ -181,6 +185,12 @@ Vagrant.configure("2") do |config|
 
       worker.vm.provision :file, :source => CA_CERT_PATH, :destination => "/tmp/ca.crt"
       worker.vm.provision :shell, :inline => "mv /tmp/ca.crt /etc/kubernetes/ca.crt", :privileged => true
+
+      worker.vm.provision :file, :source => CA_CERT_PATH, :destination => "/tmp/ca.crt"
```
ericchiang: Why is this part of the PR?

rphillips:
coreosbot run e2e

hack/terraform-quickstart/genlogs.sh
```diff
@@ -0,0 +1,20 @@
...
+set -euo pipefail
+
+export IDENT=${IDENT:-${HOME}/.ssh/id_rsa}
+export WORKER_IPS=`terraform output -json worker_ips | jq -r '.value[]'`
+export MASTER_IPS=`terraform output -json master_ips | jq -r '.value[]'`
+export BOOTSTRAP_IP=`terraform output bootstrap_node_ip`
+export LOGS_DIR=${LOGS_DIR:-/tmp/logs}
+export REMOTE_PORT=${REMOTE_PORT:-22}
+export REMOTE_USER=${REMOTE_USER:-core}
+export IDENT=${IDENT:-${HOME}/.ssh/id_rsa}
```
ericchiang: This is provided twice. Also, shouldn't ssh just work? We already ssh to the box in the terraform apply.

hack/terraform-quickstart/genlogs.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export IDENT=${IDENT:-${HOME}/.ssh/id_rsa}
+export WORKER_IPS=`terraform output -json worker_ips | jq -r '.value[]'`
+export MASTER_IPS=`terraform output -json master_ips | jq -r '.value[]'`
+export BOOTSTRAP_IP=`terraform output bootstrap_node_ip`
+export LOGS_DIR=${LOGS_DIR:-/tmp/logs}
+export REMOTE_PORT=${REMOTE_PORT:-22}
```
ericchiang: I'd like to avoid options that don't ever take more than one value. Can we remove a few of these?

hack/terraform-quickstart/main.tf
```diff
@@ -45,6 +67,27 @@ resource "aws_instance" "worker_node" {
...
+  provisioner "file" {
+    source = "environment.txt"
+    destination = "/tmp/environment"
+
+    connection {
+      user = "core"
+    }
+  }
+
+  provisioner "remote-exec" {
```
ericchiang: Since we expect these terraform setups to be usable to bring up a real cluster, we should probably make this conditional based on a terraform variable.

hack/multi-node/Vagrantfile
rphillips: good catch... there was a conflict on a rebase... I will remove this change.

hack/terraform-quickstart/genlogs.sh
rphillips: `more than one value`... those are default values if the variable does not have an overriding environment variable

hack/terraform-quickstart/genlogs.sh
rphillips: fixed in 9939c6d8

rphillips: the identity needs to be passed to ssh... we could let ssh guess, but the other scripts do it this way. My guess as to why is to allow the ssh scripts to be run locally to a provisioned cloud vm not managed by terraform.

hack/terraform-quickstart/main.tf
rphillips: terraform does not support conditionals per se, but I added an environment variable to select which environment to push (by default a blank file, and e2e when running in the e2e environment)

ericchiang: perfect, thanks

rphillips:
coreosbot run e2e

rphillips:
coreosbot run e2e

rphillips:
tests are finally green... I'm going to do one more test with a tweak to the environment to set the ttl duration to 30 minutes or so.

rphillips:
coreosbot run e2e

rphillips:
coreosbot run e2e

ericchiang:
e2e failed again :/

```
17:07:38 [211647.616257] golang[6]: --- FAIL: TestReboot (716.16s)
17:07:39 [211647.619712] golang[6]: 	reboot_test.go:25: rebooting 2 nodes
17:07:39 [211647.620132] golang[6]: 	reboot_test.go:43: waiting for control plane: get pods in kube-system: Get https://52.23.158.184:6443/api/v1/namespaces/kube-system/pods: dial tcp 52.23.158.184:6443: getsockopt: connection refused
```

I'm inclined to merge because we've still been seeing these failures and more logs are better than fewer.

Any thoughts?

rphillips:
I agree

PR #802, opened 2017-12-13
hack/quickstart: Use env file to keep track of master label/taint

klausenbusk:
Inspired by tectonic kubelet.service https://github.com/coreos/tectonic-installer/blob/89a6298bd974eaa59b391c9e6cd8218d9e70665c/modules/ignition/resources/services/kubelet.service

Not tested, as I'm not sure if this is out-of-scope for the hack scripts?


rphillips:
Merged via https://github.com/kubernetes-incubator/bootkube/pull/981

Issue #800, opened 2017-12-12
bootkube should tolerate uninitialized nodes


klausenbusk:
I think all the `bootkube` manifest should tolerate the `node.cloudprovider.kubernetes.io/uninitialized` taint.

This would make it possible to bootstrap a cluster with a external CCM (Cloud Controller Manager) and `--cloud-provider=external` and should make it safe to switch a running cluster to a CCM.

Per the design docs: https://github.com/kubernetes/community/blob/930ce65595a3f7ce1c49acfac711fee3a25f5670/contributors/design-proposals/cloud-provider/cloud-provider-refactoring.md , I don't think this would cause any issue.
```
Among these controller loops, the following are cloud provider dependent.
    nodeController
    volumeController
    routeController
    serviceController
```
```
    Node Controller: For checking the cloud provider to determine if a node has been deleted in the cloud after it stops responding
    Route Controller: For setting up routes in the underlying cloud infrastructure
    Service Controller: For creating, updating and deleting cloud provider load balancers
    Volume Controller: For creating, attaching, and mounting volumes, and interacting with the cloud provider to orchestrate volumes
```
per https://kubernetes.io/docs/concepts/overview/components/

klausenbusk:
See also: https://github.com/projectcalico/calico/pull/1286

ericchiang:
@colemickens I know you got CCM running with bootkube before. Can you comment on this?

colemickens:
I would agree with this assesment: https://github.com/colemickens/tectonic-installer/commit/4e948bd7326a0729dd18d31dd22fd475b096cc4a#diff-83beaecec190e3958db827dd8a570b14R60

PR #803, opened 2017-12-13
Tolerate uninitialized node

klausenbusk:
This ensure master components gets scheduled, even if CCM[1] hasn't
initiazlied the node yet.

[1] https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/

Fix: #800

----

Not tested yet...

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -189,6 +189,9 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       tolerations:
+      - key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
+        effect: NoSchedule
       - key: node-role.kubernetes.io/master
         operator: Exists
         effect: NoSchedule
@@ -322,6 +325,9 @@ spec:
         node-role.kubernetes.io/master: ""
       restartPolicy: Always
       tolerations:
+      - key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
+        effect: NoSchedule
       - key: node-role.kubernetes.io/master
         operator: Exists
         effect: NoSchedule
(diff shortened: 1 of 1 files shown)


ericchiang:
coreosbot test this please

colemickens:
@klausenbusk Is that taint only set when the CCM is activated? Is there any downsides to tolerating this taint when CCM is not activated on KCM, or running in the cluster? (I can't think, but wanted to explicitly make sure you'd thought about it).

klausenbusk:
> @klausenbusk Is that taint only set when the CCM is activated?

Per the [doc](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager):
> kubelets specifying `--cloud-provider=external` will add a taint `node.cloudprovider.kubernetes.io/uninitialized` with an effect `NoSchedule` during initialization. This marks the node as needing a second initialization from an external controller before it can be scheduled work. Note that in the event that cloud controller manager is not available, new nodes in the cluster will be left unschedulable. The taint is important since the scheduler may require cloud specific information about nodes such as their region or type (high cpu, gpu, high memory, spot instance, etc).

> Is there any downsides to tolerating this taint when CCM is not activated on KCM, or running in the cluster? (I can't think, but wanted to explicitly make sure you'd thought about it).

The taint is only set by `kubelet` when started with `--cloud-provider=external`, so I can't think of any downsides when not-in-use.

---

Please also see: https://github.com/kubernetes-incubator/bootkube/pull/804, which add the require logic to the `hack/quickstart` script to support CCM.

ericchiang:
coreosbot run e2e

xiang90:
```
[1052217.486576] golang[6]: --- FAIL: TestSmoke (1200.62s)
[1052217.493801] golang[6]: 	smoke_test.go:13: Test nginx creation failed: deployment nginx-7dd8f is not ready: timed out waiting for the condition
[1052217.494023] golang[6]: FAIL
```

coreosbot run e2e

xiang90:
coreosbot run e2e

aaronlevy:
Very delayed response - but I'm wondering if we should just have critical control-plane components tolerate all taints?

Issue #453, opened 2017-04-19
Add `--advertise-address` to bootstrap-apiserver manifest


aaronlevy:
In the case of vagrant/virtualbox (hack/multi-node) the temporary apiserver is advertising itself on the incorrect interface (with IP 10.0.2.15).

The default interface on virtualbox machines isn't actually routable between nodes, so what ends up happening is that the api-server ends up advertising itself on an IP that it cant be reached. Similar scenarios could occur in environments with multiple ethernet devices (not just virtualbox).

A workaround is to set the `--advertise-address` to the downward api pod-ip. This means that at runtime the kubelet will populate the same host address that the kubelet is using (because hostnetwork=true, this means the same host address the kubelet has resolved, or been told to use).

Hopefully we can just drop the same thing into the static pod manifest: 
https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L165
https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L187-L191

PR #804, opened 2017-12-13
hack/quickstart: enable setting kubelet cloud provider separately from controller manager

klausenbusk:
Running a cloud-controller-manager require kubelet to be started
with `--cloud-provider=external` and `kube-apiserver` and
`kube-controller-manager` without the `--cloud-provider` flag or a
empty string [1].
This commit make it possible to specific `--cloud-provider` for
`kubelet` without specifying a cloud-provider for `kube-apiserver`
and `kube-controller-manager` with the new env option
CLOUD_PROVIDER_KUBELET, the option use CLOUD_PROVIDER as fallback
if not set.

[1] https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager

---

Not tested yet.

As this is the future, I think it make sense to add a extra option.

### DIFF
--- a/hack/quickstart/init-master.sh
+++ b/hack/quickstart/init-master.sh
@@ -9,6 +9,7 @@ IDENT=${IDENT:-${HOME}/.ssh/id_rsa}
 SSH_OPTS=${SSH_OPTS:-}
 SELF_HOST_ETCD=${SELF_HOST_ETCD:-false}
 CLOUD_PROVIDER=${CLOUD_PROVIDER:-}
+CLOUD_PROVIDER_KUBELET=${CLOUD_PROVIDER_KUBELET:-${CLOUD_PROVIDER}}
 NETWORK_PROVIDER=${NETWORK_PROVIDER:-flannel}
 
 function usage() {
@@ -85,7 +86,7 @@ function init_master_node() {
     fi
 
     # Set cloud provider
-    sed -i "s/cloud-provider=/cloud-provider=$CLOUD_PROVIDER/" /etc/systemd/system/kubelet.service
+    sed -i "s/cloud-provider=/cloud-provider=$CLOUD_PROVIDER_KUBELET/" /etc/systemd/system/kubelet.service
 
     # Start the kubelet
     systemctl enable kubelet; sudo systemctl start kubelet
@@ -123,7 +124,7 @@ if [ "${REMOTE_HOST}" != "local" ]; then
     fi
     # Copy self to remote host so script can be executed in "local" mode
     scp -i ${IDENT} -P ${REMOTE_PORT} ${SSH_OPTS} ${BASH_SOURCE[0]} ${REMOTE_USER}@${REMOTE_HOST}:/home/${REMOTE_USER}/init-master.sh
-    ssh -i ${IDENT} -p ${REMOTE_PORT} ${SSH_OPTS} ${REMOTE_USER}@${REMOTE_HOST} "sudo REMOTE_USER=${REMOTE_USER} CLOUD_PROVIDER=${CLOUD_PROVIDER} SELF_HOST_ETCD=${SELF_HOST_ETCD} NETWORK_PROVIDER=${NETWORK_PROVIDER} /home/${REMOTE_USER}/init-master.sh local"
+    ssh -i ${IDENT} -p ${REMOTE_PORT} ${SSH_OPTS} ${REMOTE_USER}@${REMOTE_HOST} "sudo REMOTE_USER=${REMOTE_USER} CLOUD_PROVIDER=${CLOUD_PROVIDER} CLOUD_PROVIDER_KUBELET=${CLOUD_PROVIDER_KUBELET} SELF_HOST_ETCD=${SELF_HOST_ETCD} NETWORK_PROVIDER=${NETWORK_PROVIDER} /home/${REMOTE_USER}/init-master.sh local"
 
     # Copy assets from remote host to a local directory. These can be used to launch additional nodes & contain TLS assets
     mkdir ${CLUSTER_DIR}


colemickens:
Doesn't the flag need to be updated on KCM (and apiserver, though it's not used for anything) as well?

klausenbusk:
> Doesn't the flag need to be updated on KCM (and apiserver, though it's not used for anything) as well?

Not per the doc: https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager
> kube-apiserver and kube-controller-manager MUST NOT specify the --cloud-provider flag. This ensures that it does not run any cloud specific loops that would be run by cloud controller manager. In the future, this flag will be deprecated and removed.
> kubelet must run with --cloud-provider=external. This is to ensure that the kubelet is aware that it must be initialized by the cloud controller manager before it is scheduled any work.
> kube-apiserver SHOULD NOT run the PersistentVolumeLabel admission controller since the cloud controller manager takes over labeling persistent volumes. To prevent the PersistentVolumeLabel admission plugin from running, make sure the kube-apiserver has a --admission-control flag with a value that does not include PersistentVolumeLabel.
For the cloud-controller-manager to label persistent volumes, initializers will need to be enabled and an InitializerConifguration needs to be added to the system. Follow these instructions to enable initializers. Use the following YAML to create the InitializerConfiguration:

colemickens:
Very interesting. I've seen it working with apiserver and kcm specifying `external` but the docs are clear and the other usage of CCM strips off the `--cloud-provider` flags from KCM and apiserver, which also supports what you're saying.

Let me touch base with the `sig` to make sure this text is still accurate. For some reason I thought there was discussion of using `--cloud-provider=external` on KCM to change the behavior of volumes, but that might require another flag to get KCM to do external mode, but still provide azure|gce|aws disk attach support.

I guess you're not stripping the flag, so it's expected that the user then set `export CLOUD_PROVIDER=` and `export CLOUD_PROVIDER_KUBELET=external` ?

klausenbusk:
> I guess you're not stripping the flag, so it's expected that the user then set export CLOUD_PROVIDER= and export CLOUD_PROVIDER_KUBELET=external?

`CLOUD_PROVIDER` default to "", so setting `CLOUD_PROVIDER_KUBELET` should be enough.

klausenbusk:
> I've seen it working with apiserver and kcm specifying external

The text was changed in 1.8 I think, please keep me updated I'm also a bit confused.

colemickens:
Looks like the PR I'm thinking of is here: https://github.com/kubernetes/kubernetes/pull/52371

For now, let's go ahead and leave this as-is. When this PR merges and 1.10 is out, we can revisit and activate the extra flag to keep volumes working until CCM+CSI is ready and the flag is deprecated again.

Source for the info, @dims, from Slack. Thanks @dims! For historical sake:

---

colemickens [1:55 PM] 
is assuming the docs are still correct that kcm/apiserver are to be run with no --cloud-provider flag.


colemickens
[1:55 PM] 
I think there was discussion of changing KCM so it could run controller loops, but IIRC that would be via a new flag and isn't there yet.


dims [2:18 PM] 
@colemickens this one is merging. was that the one you were thinking of? https://github.com/kubernetes/kubernetes/pull/52371
 
colemickens [2:23 PM] 
Yes, that's exactly what I'm looking for. We have a PR in bootkube that doesn't set the --cloud-provider flag on KCM and I was thinking we wanted to set it to `external` so that we could eventually utilize this.

[2:23] 
However there is text that says to explicitly NOT set --cloud-provider on apiserver/kcm today.

[2:23] 
So that will probably need to get updated once that PR merges. (edited)

dims [2:24 PM] 
@colemickens right. you should only set --cloud-provider=external if you really know you are using / need it. else don’t set it

colemickens:
coreosbot test this please

ericchiang:
coreosbot test this please

dghubble:
Has anyone set this up? Where are the materials on it? It's still listed as untested. Is the understanding that this flag is to enable experimenting with CCM?

ericchiang:
> Has anyone set this up? Where are the materials on it? It's still listed as untested. Is the understanding that this flag is to enable experimenting with CCM?

CCM is still experimental, I don't think we would accept a PR to add it to bootkube yet.

This PR just lets you set the kubelet cloud controller separately from the controller manager. It seems fine in that respect.

dghubble:
Let's not have a title like "Support CCM (Cloud Controller Manager)" then.

dghubble:
Otherwise, no complaints with allowing it in the hack scripts to facilitate experiments / adoption.

colemickens:
Updated the PR title. Will amend the commit and push as well.

klausenbusk:
So I just did a quick testing (after rebasing) of this and it isn't working. When `kubelet` is started with `--cloud-provider=external`, `status.podIP` is set to `` and `bootstrap-apiserver` then can't start.

klausenbusk:
> So I just did a quick testing (after rebasing) of this and it isn't working. When kubelet is started with --cloud-provider=external, status.podIP is set to `` and bootstrap-apiserver then can't start.

Someone has already reported this upstream: https://github.com/kubernetes/kubernetes/issues/54965 (but the "bug" was lated closed), so at the moment it seems we end up in a chicken-egg situation no matter what we do.

klausenbusk:
> so at the moment it seems we end up in a chicken-egg situation no matter what we do.

The help page seems to be incorrect at the moment (a `--advertise-address` blank should work), I have opened a issue upstream: https://github.com/kubernetes/kubernetes/issues/58686 .

Let set this on hold until then.
/hold

klausenbusk:
> Let set this on hold until then.
> /hold

[Upstream](https://github.com/kubernetes/kubernetes/issues/58686) is gonna fix the comment not the behavior.

I'm not sure how we can fix this chicken-egg situation then, `--advertise-address` was added for a [reason](https://github.com/kubernetes-incubator/bootkube/issues/453).

colemickens:
@klausenbusk Can you please elaborate? There's confusion about a couple of things:

- From my reading and understanding, it sounds like `--advertise-address=` (as in, specified, but blank) does "work", but defaults to all interfaces. (Which is undesirable in bootkube for the reason you linked)

- Why does CCM/external cloud provider affect this? You're saying that disabling the cloudprovider on kubelet causes the `podIP` to be empty? Do you know why?

colemickens:
Also, have you filed an upstream bug for the CCM/podIP behavior, or this general issue? I presume others would be affected by this, if it's a fundamental problem, especially as kubeadm matures and embraces self-hosting.

klausenbusk:
> From my reading and understanding, it sounds like --advertise-address= (as in, specified, but blank) does "work", but defaults to all interfaces. (Which is undesirable in bootkube for the reason you linked)

The help text is wrong, a blank `--advertise-address=` doesn't work (`Error: invalid argument "" for --bind-address=: failed to parse IP: ""`), upstream is gonna update the help text: https://github.com/kubernetes/kubernetes/issues/58686#issuecomment-359783776


> Why does CCM/external cloud provider affect this? You're saying that disabling the cloudprovider on kubelet causes the podIP to be empty? Do you know why?

It is the CCM job to run the `Node controller`, which I assume (?) set the `hostIP` (which is then used by `podIP`), so until the CCM has initialized the node `hostIP` is "blank".


> Also, have you filed an upstream bug for the CCM/podIP behavior, or this general issue? I presume others would be affected by this, if it's a fundamental problem, especially as kubeadm matures and embraces self-hosting.

See: https://github.com/kubernetes/kubernetes/issues/55633 and https://github.com/kubernetes/kubernetes/pull/50545

colemickens:
I guess I'm still a bit confused because I had CCM working with bootkube in [a `tectonic-installer` WIP branch](https://github.com/colemickens/tectonic-installer/commits/ccm-stash), and I know the Azure folks are having successes running CCM as well. Maybe you or I could try to look at it and see where the diff is.

I don't really understand why hostIP needs to be known to know the podIP. The pod and host network are often different anyway (overlay/underlay/physical/etc).

klausenbusk:
> I guess I'm still a bit confused because I had CCM working with bootkube in a tectonic-installer WIP branch, and I know the Azure folks are having successes running CCM as well. Maybe you or I could try to look at it and see where the diff is.

Firstly, they don't seems to be using [TLS bootstrapping](https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/), which isn't supported at the moment with a CCM (https://github.com/kubernetes/kubernetes/issues/55633).

> I don't really understand why hostIP needs to be known to know the podIP.

When the pod is started with `hostNetwork: true` it use `hostIP` as `podIP`.

klausenbusk:
> - but not sure if there were still some outstanding issues.

I think it should work now with https://github.com/kubernetes/kubernetes/pull/65594 merged, but I'm not sure and nevertheless we need to wait for 1.12.

PR #808, opened 2017-12-18
add root requirements doc

rphillips:
Documents the root user requirements of the bootkube components.

### DIFF
--- a/Documentation/root-requirements.md
+++ b/Documentation/root-requirements.md
@@ -1,0 +1,40 @@
+# Information about services running as root
+
+Running services as a non-root user is typically better for security, since
+permissions can be controlled.
+
+Some Kubenetes services can be run as a non-root user, but most require root
+privileges in some manner. This document will try out outline why certain
+Kubernetes services require root privileges.
+
+## Services that do *not* run as root
+
+* Scheduler
+* Controller Manager
+
+## Services requiring root
+
+### Flannel
+
+Flannel is a layer 3 network fabric. Flannel needs access to low level
+networking interfaces to be able to dynamically configure them.
+
+### Proxy
+
+kube-proxy is a network proxy found on every Kubernetes node. Kube-proxy often
+needs to open and manage privileged ports (< 1024) in addition to managing iptables.
+
+### API Server
+
+Currently runs as root, but is in the process of being
+[migrated](https://github.com/kubernetes-incubator/bootkube/pull/789) to use
+non-root ports.
+
+### DNS
+
+DNS needs to bind to privileged port 53 for UDP and TCP.
+
+### Checkpointer
+
+Checkpointer is a service to recover a cluster after a reboot or loss of a node.
+This service writes to `/etc/kubernetes/manifests`.


Documentation/root-requirements.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Flannel
+
+Flannel is a layer 3 network fabric. Flannel needs access to low level
+networking interfaces to be able to dynamically configure them.
+
+### Proxy
+
+kube-proxy is a network proxy found on every Kubernetes node. Kube-proxy often
+needs to open and manage privileged ports (< 1024).
```
aaronlevy: nit: also because it needs to interact with iptables in general

Documentation/root-requirements.md
```diff
@@ -0,0 +1,38 @@
...
+
+### API Server
+
+The API server needs read access to `/etc/kubernetes/manifests`.
+
+### DNS
+
+DNS needs to bind to privileged port 53 for UDP and TCP.
```
aaronlevy: I wonder if we should change this. Typically you would only hit dns through the service, which could target a different port on the dns pod. Just wanted to raise question - but can be address separately (it's true that it needs to point at 53 right now).

rphillips: created an issue to track the DNS moving to an unprivileged port.

Documentation/root-requirements.md
```diff
@@ -0,0 +1,38 @@
...
+networking interfaces to be able to dynamically configure them.
+
+### Proxy
+
+kube-proxy is a network proxy found on every Kubernetes node. Kube-proxy often
+needs to open and manage privileged ports (< 1024).
+
+### API Server
+
+The API server needs read access to `/etc/kubernetes/manifests`.
```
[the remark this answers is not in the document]
rphillips: after closer inspection we read /etc/kubernetes/secrets... Does this mount point currently require root?

Documentation/root-requirements.md
rphillips: fixed

Documentation/root-requirements.md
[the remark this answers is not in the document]
rphillips: hmm. no it does not... perhaps we could migrate api server to use non root

rphillips:
tweaked the API server section

Documentation/root-requirements.md
```diff
@@ -0,0 +1,40 @@
...
+
+### API Server
+
+Currently runs as root, but is in the process of being
+[migrated](https://github.com/kubernetes-incubator/bootkube/pull/789) to use
+non-root ports.
+
+### DNS
+
+DNS needs to bind to privileged port 53 for UDP and TCP.
```
xiang90: will CAP_NET_BIND_SERVICE solve this issue?

aaronlevy: It would, but it might entail us having to build our own kube-dns container (to add that capability to the binary) then also still have a PSP that allows the container to have that capability.

If we don't have to directly access the dns container, we can always reference it through a service that is bound to port 53 (which is already true). I just don't know for sure if there is a reason we need the container to bind to 53 specifically.

xiang90: > which is already true

@aaronlevy 

What does this me specifically? Do you mean that we already access the dns container through a service that is bound to port 53 so that we can map it to a different port that the actual dns server could potentially bind to?

aaronlevy: There is a dns service already listening on port 53. We would map the container to a different (>1024) port as long as nothing expects direct access the dns pod via port 53

xiang90: @aaronlevy Thanks!

@rphillips maybe the doc should also capture the potential alternatives we discussed here.

xiang90: i see https://github.com/kubernetes-incubator/bootkube/issues/809 exist :).

dghubble: Isn't upstream migrating to CoreDNS gradually anyway? These questions may have been raised then. Discussion of modifying a kube-dns image doesn't seem too relevant going forward.

@aaronlevy any thoughts on the ramifications of running the DNS service on a non-standard port?

dghubble: Or you're envisioning kube-proxy is in-place and its only access through the service?

aaronlevy: Right, dns would still be accessed through standard port (via the service). This is already true to the best of my knowledge (e.g. in kubelet flags you point to service-ip:53, and that's what gets injected into all pods). There might just be a separate reason that the container also binds to 53 -- but none that I know of.

Not sure that CoreDNS would actually replace recommended (non-alpha) option within 1.10 timeframe at least. And I'm unfamiliar with the project so don't know how much this matters.

We can punt on changing the dns pod for now - as long as the PSP allows it to run as root (or we change it to have the capability - but if that requires managing our own releases of dnsmasq container -- not inclined to go that route).

xiang90: @aaronlevy @dghubble 

I would not prefer changing the dns port, which can introduce risks, than tedious work that can be automated but has less risk. Or we need to 100% understand the ramifications of changing the port.

Just to be clear, I raised this up not because I want to get a conclusion in this PR. I just want to document the options we have. We can revisit this later.

aaronlevy: This isn't about automating tedious work - it's whether the pod needs to run as roon. Preferably it does not.

It really comes down to "is there any reason that you would need to contact the pod directly, and not through a service". Which to my knowledge, the answer is "no". But that just needs to be further evaluated.

As it is configured today - this document is correct. When someone has time they can evaluate if we can change the port - which is what https://github.com/kubernetes-incubator/bootkube/issues/809 is for.

aaronlevy:
@rphillips last changes that I'd ask before merging:

- Open an issue to track removing need for checkpointer to have root. It happens to use /etc/kubernetes/manifests - but no strict reason it needs to use that dir or be root (that I know of).

- Add links to the above issue and #809 to their respective sections just so it's clear we're looking into options about removing root requirement (also for apiserver)

Documentation/root-requirements.md
xiang90: > This isn't really about automating tedious work - it's whether the pod needs to run as root. 

I think you misunderstood what I said. I was comparing changing port vs set capabilities rather than changing port vs keeping the container run as root.

> Which to my knowledge, the answer is "no". But that just needs to be further evaluated.

Correct. I want us to confirm this before making any decision.

Thanks.

rphillips:
Thanks @aaronlevy. I'll get this updated.

rphillips:
@aaronlevy fixed

PR #814, opened 2017-12-20
checkpointer: write secrets and configmaps with uid.

diegs:
This reads the uid from the security context and uses it when
checkpointing secrets and configmaps.

If there is a conflict between containers or the outer pod security it
returns an error for now. In the future we could make multiple copies,
one with each uid.

### DIFF
--- a/pkg/checkpoint/config_map.go
+++ b/pkg/checkpoint/config_map.go
@@ -12,13 +12,18 @@ import (
 
 // checkpointConfigMapVolumes ensures that all pod configMaps are checkpointed locally, then converts the configMap volume to a hostpath.
 func (c *checkpointer) checkpointConfigMapVolumes(pod *v1.Pod) (*v1.Pod, error) {
+	uid, gid, err := podUserAndGroup(pod)
+	if err != nil {
+		return nil, fmt.Errorf("failed to checkpoint configMap for pod %s/%s: %v", pod.Namespace, pod.Name, err)
+	}
+
 	for i := range pod.Spec.Volumes {
 		v := &pod.Spec.Volumes[i]
 		if v.ConfigMap == nil {
 			continue
 		}
 
-		_, err := c.checkpointConfigMap(pod.Namespace, pod.Name, v.ConfigMap.Name)
+		_, err := c.checkpointConfigMap(pod.Namespace, pod.Name, v.ConfigMap.Name, uid, gid)
 		if err != nil {
 			return nil, fmt.Errorf("failed to checkpoint configMap for pod %s/%s: %v", pod.Namespace, pod.Name, err)
 		}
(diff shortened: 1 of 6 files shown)


diegs:

diegs:
Flake on `e2e-calico` which is irrelevant since this code only touched the checkpointer (tested by `e2e w/checkpointer`)

pkg/checkpoint/manifest.go
```diff
@@ -78,10 +78,10 @@ func writeManifestIfDifferent(path, name string, data []byte) (bool, error) {
 		return false, nil
 	}
 	glog.Infof("Writing manifest for %q to %q", name, path)
-	return true, writeAndAtomicRename(path, data, 0644)
+	return true, writeAndAtomicRename(path, data, 0644, rootUID, rootGID)
 }
 
-func writeAndAtomicRename(path string, data []byte, perm os.FileMode) error {
+func writeAndAtomicRename(path string, data []byte, uid, gid int, perm os.FileMode) error {
```
xiang90: this func is busy enough i feel. probably should not couple chown into this function.

xiang90:
> Flake on e2e-calico which is irrelevant since this code only touched the checkpointer (tested by e2e w/checkpointer)

Do we have an issue tracking the calico test failure?

diegs:
Filed https://github.com/kubernetes-incubator/bootkube/issues/816

rphillips:
@xiang90 not yet.

xiang90:
@diegs Thanks!

pkg/checkpoint/config_map.go
```diff
@@ -12,13 +12,18 @@ import (
...
+	}
+
 	for i := range pod.Spec.Volumes {
 		v := &pod.Spec.Volumes[i]
 		if v.ConfigMap == nil {
 			continue
 		}
 
-		_, err := c.checkpointConfigMap(pod.Namespace, pod.Name, v.ConfigMap.Name)
+		_, err := c.checkpointConfigMap(pod.Namespace, pod.Name, v.ConfigMap.Name, uid, rootGID)
```
squat: Why use the root GID here? The `podUser` func could return the `fsGroup` from the container's securityContext if it is available, and default to `rootGID` otherwise.

pkg/checkpoint/secret.go
```diff
@@ -12,13 +12,18 @@ import (
...
+	}
+
 	for i := range pod.Spec.Volumes {
 		v := &pod.Spec.Volumes[i]
 		if v.Secret == nil {
 			continue
 		}
 
-		_, err := c.checkpointSecret(pod.Namespace, pod.Name, v.Secret.SecretName)
+		_, err := c.checkpointSecret(pod.Namespace, pod.Name, v.Secret.SecretName, uid, rootGID)
```
squat: same here re `rootGID`

pkg/checkpoint/config_map.go
diegs: Done.

pkg/checkpoint/manifest.go
```diff
@@ -78,10 +78,10 @@ func writeManifestIfDifferent(path, name string, data []byte) (bool, error) {
 		return false, nil
 	}
 	glog.Infof("Writing manifest for %q to %q", name, path)
-	return true, writeAndAtomicRename(path, data, 0644)
+	return true, writeAndAtomicRename(path, data, 0644, rootUID, rootGID)
```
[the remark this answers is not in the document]
diegs: Fixed, thanks!

pkg/checkpoint/secret.go
diegs: Done.

pkg/checkpoint/manifest.go
diegs: I think it's better practice to force all callers to specify the UID and GID explicitly.

diegs:

diegs:
@squat PTAL

diegs:
coreosbot run e2e

diegs:
Again reiterating that "e2e w/checkpointer" is the only test result that matters for this PR.

PR #817, opened 2017-12-20
Update code-of-conduct.md

spiffxp:
Refer to kubernetes/community as authoritative source for code of conduct

ref: kubernetes/community#1527


aaronlevy:
lgtm

side note @spiffxp should we maybe have everything point to CNCF code-of-conduct? (otherwise the k8s one is just essentially redirect to that anyway). I know you just opened this on a bunch of repos so no worries - just a thought.

Issue #824, opened 2018-01-08
Reduce e2e test flakes


ericchiang:
# PR e2e Calico

Between 259 and 228 inclusive 12 builds failed (~35%).

TestCheckpointer/UnscheduleCheckpointer

2018-01-08 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/259/console
2017-12-20 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/243/console
2017-12-20 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/241/console
2017-12-19 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/236/console
2017-12-19 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/233/console
2017-12-19 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/230/console
2017-12-18 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/228/console

TestDeleteAPI

2018-01-02 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/253/console

Network failures

2018-01-02 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/254/console

Unit test failures

2017-12-20 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/239/console

Log collection

2017-12-19 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/232/console

Control plane failed

2017-12-18 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e-calico/229/console

diegs:
Smoke test (probably actually reboot test):

2018-01-08 https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e/683/console

ericchiang:
Even with Cole's PR we're still seeing flakes in TestReboot and consequentially TestSmoke. https://github.com/kubernetes-incubator/bootkube/issues/864

[...]

PR #831, opened 2018-01-12
templates: increase API server verbosity

ericchiang:
This PR is intended to help debug our e2e test flakes https://github.com/kubernetes-incubator/bootkube/issues/824. It's not intended to be merged.

### DIFF
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/golang/glog"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


xiang90:
maybe our e2e test should always start the API server with v6?

ericchiang:
@xiang90 yeah but I don't think the audit changes are appropriate for everyone. Would require a knob.

xiang90:
> @xiang90 yeah but I don't think the audit changes are appropriate for everyone.

agree

> Would require a knob.

i hate to add the knob just for testing in general. we probably can figure out something better. one ugly approach is to add it in testing scripts after running bootkube render.

ericchiang:
Found https://github.com/kubernetes-incubator/bootkube/issues/837

Closing for now.

PR #832, opened 2018-01-12
log in UTC time

rphillips:


rphillips:
@ericchiang @colemickens Sorry, Needs a review again. The default flags include '`Ldate | Ltime`'

PR #849, opened 2018-01-22
pkg/asset: Remove kubenetes-etcd-network-checkpointer

dghubble:
* kenc was associated with self-hosted etcd and is no longer used


xiang90:
lgtm

PR #850, opened 2018-01-22
use the lease endpoint reconciler for the API

rphillips:
Turns on the lease endpoint reconciler for the API server.


ericchiang:
A little hesitant because this feature is technically alpha, and because we don't have HA test coverage. Any thoughts here?

klausenbusk:
Another option is using the `none` reconciler and adding a regular `kubernetes` service which use a selector.

xiang90:
I agree that we need some testing coverage on this before turning it on. Probably on k8s upstream.

But as a side note, we know the current setup is pretty broken anyway...

PR #863, opened 2018-01-24
add more bootkube recover docs

rphillips:
* Add some documentation on how bootkube recover works
* Remove self hosted etcd variables

### DIFF
--- a/Documentation/disaster-recovery.md
+++ b/Documentation/disaster-recovery.md
@@ -47,10 +47,41 @@ In the event of partial or total self-hosted control plane loss, `bootkube
 recover` may be able to assist in re-bootstrapping the self-hosted control
 plane.
 
-The `bootkube recover` subcommand does not recover a cluster directly. Instead,
-it extracts the control plane configuration from an available source and
-renders manifests in a format that `bootkube start` can use invoked to reboot
-the cluster.
+The `bootkube recover` subcommand does not recover a cluster directly. The
+recovery is a two step process: `bootkube recover` then `bootkube start`. The
+recovery command extracts the control plane configuration from an available
+source and renders manifests to the local filesystem. These resulting manifests
+can be passed to `bootkube start`.
+
+There are two available sources to choose from in `recover`: etcd or API server.
+
+### What does bootkube recover do?
+
+`bootkube recover`attempts to read the configuration from an existing backend etcd or
+API server. On success, `bootkube recover` writes manifests for a modified
+bootstrap control plane to a directory. The second phase of the recover can be
+initiated by an administrator by running `bootkube start` on these manifests.
+
+`bootkube recover` modifies bootstrap pod specs in the following ways:
+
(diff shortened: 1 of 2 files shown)


Documentation/disaster-recovery.md
```diff
@@ -50,7 +50,26 @@ plane.
...
 it extracts the control plane configuration from an available source and
 renders manifests in a format that `bootkube start` can use invoked to reboot
-the cluster.
+the cluster. There are two available sources to choose from: etcd or an API
+server.
+
+### What does bootkube recover do?
+
+Bootkube attempts to read the configuration from an existing backend etcd or API
+server. On success, the tool renders a modified bootstrap control plane to
```
xiang90: what does `modified` mean? also can we clarify that the tool renders manifests (basically change control plane to control plane manifests).

pkg/recovery/recover.go
```diff
@@ -80,11 +80,6 @@ type controlPlane struct {
 	daemonSets  v1beta1.DaemonSetList
 	deployments v1beta1.DeploymentList
 	secrets     v1.SecretList
-
-	// for self hosted etcd recovery if not nil
-	bootEtcd        *asset.Asset
```
xiang90: can we separate this out to another commit? want to keep this commit cleaner.

rphillips: split into two commits... still working on addressing the comments.

Documentation/disaster-recovery.md
```diff
@@ -47,10 +47,31 @@ In the event of partial or total self-hosted control plane loss, `bootkube
...
+renders manifests to the local filesystem in a format that `bootkube start` can
+be invoked to reboot the cluster.
+
+There are two available sources to choose from: etcd or an API server.
+
+### What does bootkube recover do?
+
+Bootkube attempts to read the configuration from an existing backend etcd or API
+server. On success, the tool renders a modified bootstrap control plane to
+support direct injection into a kubelet.
```
colemickens: Clarify what "direct injection into a kubelet" means. I assume it means, roughly, "in a format that `bootkube start` can provide to the kubelet, as static pods, to restart the control plane bootstrap process".

colemickens: Or maybe more simply "in a format that `bootkube start` can handle", since the user doesn't really "directly inject" static pods, but rather just calls `bootkube start` (which is documented elsewhere).

rphillips: good call.. @xiang90 @colemickens I believe the current commits address your comments.

xiang90:
lgtm. defer to @colemickens or @ericchiang. thanks!

rphillips:
Does the recovery control plane need to be torn down?

ericchiang:
> Does the recovery control plane need to be torn down?

From slack. `bootkube start` already tears down the temp control plane. It'll do the same thing for the recovery control plane.

PR #866, opened 2018-01-25
Fail release build on gofmt diffs

colemickens:
The current Makefile doesn't fail builds on `gofmt` errors. This changes causes it to fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ release: \
 	_output/release/bootkube.tar.gz \
 
 check:
-	@find . -name vendor -prune -o -name '*.go' -exec gofmt -s -d {} +
+	@gofmt -l -s $(GOFILES) | read; if [ $$? == 0 ]; then gofmt -s -d $(GOFILES); exit 1; fi
 	@go vet $(shell go list ./... | grep -v '/vendor/')
 	@go test -v $(shell go list ./... | grep -v '/vendor/\|/e2e')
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -28,7 +28,7 @@ release: \
 	_output/release/bootkube.tar.gz \
 
 check:
-	@find . -name vendor -prune -o -name '*.go' -exec gofmt -s -d {} +
+	@gofmt -s -d $$(find . -name '*.go' -print -o -name vendor -prune); if [ "$$?" = "0" ]; then exit 1; fi
```
rphillips: we could use `@gofmt -s -d $$(find . -name '*.go' -print -o -name vendor -prune) && exit 1`

colemickens:
@rphillips Turns out exit code is unreliable for `gofmt`. Fighting make+shell was frustrating. Latest push is what I came up with that works for both pass/nopass cases.

Makefile
colemickens: I tried capturing the output, checking if empty, and then printing the output, instead of invoking `gofmt` twice, but it was more trouble than it was worth to get make/shell to do what I wanted.

ericchiang:
Control plane didn't come back :(

https://jenkins-kube-lifecycle.prod.coreos.systems/job/bootkube-pr-e2e/793/console

ericchiang:
coreosbot e2e test this

ericchiang:
coreosbot run e2e

colemickens:
coreosbot run e2e

colemickens:
coreosbot run e2e

colemickens:
coreosbot run e2e

PR #876, opened 2018-02-05
 *: switch to dep for dependency management

ericchiang:


scripts/verify-gopkg.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
```
colemickens: Did you notice anything that fails this check so far? Or is this an assert style sanity check?

This also doesn't seem to assert that the assets in the vendor/ are valid corresponding to either the toml/lock. Is that worth doing as well?

scripts/verify-gopkg.sh
```diff
@@ -0,0 +1,15 @@
...
+
+TOML=$( grep -P '( |^)name = ".*"' Gopkg.toml | tr -d ' ' | sort )
+LOCK=$( grep -P '( |^)name = ".*"' Gopkg.lock | tr -d ' ' | sort )
+
+# NOTE(ericchiang): For some reason you need this "export" here for bash reasons
+# I don't understand.
+export DIFF=$( diff <(echo "$TOML") <(echo "$LOCK") )
```
colemickens: Given you comment, and a quick test, did you ensure this works in the failure case?

You don't wrap the value in quotes, and `DIFF=` and `export DIFF=` work differently if the values have spaces, at least in my quick little experiment it seemed to.

It seems like maybe `export` is papering over the fact that your value has whitespace and allowing for false positives?

rphillips: might need a `-q` in there as well to only print when files differ

scripts/verify-gopkg.sh
ericchiang: Assert style sanity check.

scripts/verify-gopkg.sh
ericchiang: I confirmed that this works when the files are out of sync.

If I comment out the `export` and modify one of the Gopkg files I can replicate locally that `DIFF` becomes set to the empty string no matter what.

I can add some quotes, though I tested that and it didn't work to fix this problem. @thorfour and I were staring at this for a while and don't have any good answer here.

rphillips: should be:

```
DIFF=$( diff <(echo "$TOML") <(echo "$LOCK") || true )
```

`diff` will return an error code if there is a difference causing the script to exit.

ericchiang: it worked!

ericchiang:
Gonna wait for https://github.com/kubernetes-incubator/bootkube/pull/877 to merge to push the change.

rphillips:
merged #877

PR #882, opened 2018-02-09
[WIP] heptio-authenticator-aws side car with apiserver

abhinavdahiya:
This adds https://github.com/heptio/authenticator as a side car with the kube-apiserver. This will allow users to use their aws credentials.

- `aws-authenticator-config.yaml` file configures the authenticator with required AWS resource names to kubernetes group,username

/cc: @ericchiang

### DIFF
--- a/hack/quickstart/aws-authenticator-config.yaml
+++ b/hack/quickstart/aws-authenticator-config.yaml
@@ -1,0 +1,42 @@
+# a unique-per-cluster identifier to prevent replay attacks (see above)
+clusterID: bootkube-cluster
+server:
+  # state directory for generated TLS certificate and private keys
+  stateDir: /var/aws-authenticator
+  generateKubeconfig: /var/aws-authenticator/kubeconfig.yaml
+  # each mapRoles entry maps an IAM role to a username and set of groups
+  # Each username and group can optionally contain template parameters:
+  #  1) "{{AccountID}}" is the 12 digit AWS ID.
+  #  2) "{{SessionName}}" is the role session name.
+  mapRoles:
+  # statically map arn:aws:iam::000000000000:role/KubernetesAdmin to cluster admin
+  #- roleARN: arn:aws:iam::000000000000:role/KubernetesAdmin
+  #  username: kubernetes-admin
+  #  groups:
+  #  - system:masters
+  # map EC2 instances in my "KubernetesNode" role to users like
+  # "aws:000000000000:instance:i-0123456789abcdef0". Only use this if you
+  # trust that the role can only be assumed by EC2 instances. If an IAM user
+  # can assume this role directly (with sts:AssumeRole) they can control
+  # SessionName.
+  #- roleARN: arn:aws:iam::000000000000:role/KubernetesNode
+  #  username: aws:{{AccountID}}:instance:{{SessionName}}
+  #  groups:
+  #  - system:bootstrappers
+  #  - aws:instances
+  # map federated users in my "KubernetesAdmin" role to users like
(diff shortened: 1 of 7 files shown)


abhinavdahiya:
@ericchiang does this look like a workable path towards integrating this?

hack/terraform-quickstart/main.tf
```diff
@@ -16,17 +16,18 @@ resource "aws_instance" "bootstrap_node" {
...
+  )}"
 
   root_block_device {
     volume_type = "gp2"
     volume_size = "30"
   }
 
   provisioner "file" {
-    source = "environment_${var.environment}.txt"
+    source      = "environment_${var.environment}.txt"
```
ericchiang: Ah, we should make sure we `terraform fmt` all of these files as part of CI.

cc @rphillips

pkg/asset/images.go
```diff
@@ -2,14 +2,15 @@ package asset
...
+	Flannel:          "quay.io/coreos/flannel:v0.10.0-amd64",
+	FlannelCNI:       "quay.io/coreos/flannel-cni:v0.3.0",
+	Calico:           "quay.io/calico/node:v2.6.6",
+	CalicoCNI:        "quay.io/calico/cni:v1.11.2",
+	Hyperkube:        "gcr.io/google_containers/hyperkube:v1.9.2",
+	KubeDNS:          "gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.14.8",
+	KubeDNSMasq:      "gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64:1.14.8",
+	KubeDNSSidecar:   "gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.8",
+	PodCheckpointer:  "quay.io/coreos/pod-checkpointer:3cd08279c564e95c8b42a0b97c073522d4a6b965",
+	AWSAuthenticator: "quay.io/abhinavdahiya/aws-authenticator:dev",
```
ericchiang: Why `quay.io/abhinavdahiya`? We shouldn't pull images from an individual developer.

pkg/asset/internal/templates.go
```diff
@@ -184,6 +187,25 @@ spec:
...
+        - mountPath: /var/aws-authenticator
+          name: var-aws-authenticator
+          readonly: true
+      - args:
+        - server
+        - --config=/etc/kubernetes/aws-authenticator/config.yaml
+        image: {{ .Images.AWSAuthenticator }}
+        name: aws-authenticator
+        volumeMounts:
+        - mountPath: /etc/ssl/certs
```
ericchiang: Why do we need this?

ericchiang:
Awesome.

We need to think about how we're actually going to merge this, it plays pretty heavily into "what's the role of render?". We probably don't want to add arbitrary flags to render, it needs to remain a reference architecture. Do we want to switch to something like https://github.com/poseidon/terraform-render-bootkube/ ?

I'm going to open an issue describing what we want to do with this TLS bootstrapping attestation work, and how it fits into bootkube. Hopefully we can decide if this a set of experimental flags on bootkube, if it's going to be an internal only thing, if we want to switch to more sophisticated rendering, etc.

hack/quickstart/aws-authenticator-config.yaml
```diff
@@ -0,0 +1,42 @@
+# a unique-per-cluster identifier to prevent replay attacks (see above)
```
ericchiang: In the current setup we probably still want bootkube rendering this. Let me open an issue before we address that.

PR #888, opened 2018-02-12
Remove PersistentVolumeLabel admission controller flag

dghubble:
* PersistentVolumeLabel admission controller is deprecated in 1.9

I noticed warnings in apiserver logs that this admission controller is deprecated. It is indeed mentioned in 1.9 release deprecation [notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md#deprecations) and details are found in [#52618](https://github.com/kubernetes/kubernetes/pull/52618).

PR #889, opened 2018-02-13
render: increase checkpointer grace period to 10m

ericchiang:
We seem to hit this a lot in e2e. Bumping to see if it impacts the flakes.

cc @dghubble @rphillips @diegs

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -293,6 +293,7 @@ spec:
         - /checkpoint
         - --lock-file=/var/run/lock/pod-checkpointer.lock
         - --kubeconfig=/etc/checkpointer/kubeconfig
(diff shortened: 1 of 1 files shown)


ericchiang:
ref https://github.com/kubernetes-incubator/bootkube/pull/888#issuecomment-365101907

ericchiang:
According to @diegs the grace period is the time between the checkpointer believing the API server is alive and it actually deleting the checkpointed API server (or whatever other pod it's checkpointing). "Hitting the grace period" is normal and isn't something we should worry about.

diegs:
The states (including the grace period ones) have some godoc that describes how they fit together:

https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/checkpoint/state.go

PR #899, opened 2018-02-16
jenkins: add e2e script

colemickens:
This is the script from Jenkins with modifications to support multiple clusters side-by-side.

The TF state and a per-cluster SSH key are saved into a cluster directory. This is instead of relying on the user having a specially named ssh key already active in their AWS subscription.

cc: @ericchiang @rphillips

### DIFF
--- a/hack/.gitignore
+++ b/hack/.gitignore
@@ -1,0 +1,1 @@
+_clusters/
--- a/hack/quickstart/init-master.sh
+++ b/hack/quickstart/init-master.sh
@@ -109,11 +109,6 @@ function init_master_node() {
 
 [ "$#" == 1 ] || usage
 
-[ -d "${CLUSTER_DIR}" ] && {
-    echo "Error: CLUSTER_DIR=${CLUSTER_DIR} already exists"
-    exit 1
-}
-
 # This script can execute on a remote host by copying itself + bootkube binary + kubelet service unit to remote host.
 # After assets are available on the remote host, the script will execute itself in "local" mode.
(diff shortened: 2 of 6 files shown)


ericchiang:
```
17:54:17 
17:54:17 Error: Error asking for user input: missing required value for "ssh_public_key"
17:54:17 
17:54:17 
17:54:17 Generating Logs
17:54:17 The module root could not be found. There is nothing to output.
17:54:17 The module root could not be found. There is nothing to output.
17:54:17 The module root could not be found. There is nothing to output.
17:54:17 Gathering Logs
17:54:17 The module root could not be found. There is nothing to output.
17:54:17 The module root could not be found. There is nothing to output.
17:54:17 The module root could not be found. There is nothing to output.
17:54:17 	zip warning: name not matched: logs*
17:54:17
```

colemickens:
@ericchiang I can't repro that yet, but am patching up something else caused by a need for a newer version of terraform after rebasing (I'm actually surprised you made it this far(?)). I assume that was a fast-fail on `terraform apply` and not after a successful deployment?

ericchiang:
Can we be sure to bump terraform in an independent PR?

colemickens:
@ericchiang The place where I've bumped it locally is in this (new) script...

I'm actually confused how this is passing e2e in Jenkins (unless that job was updated since I forked off the script). I'll look into it more.

colemickens:
Yeah, looks like the current Jenkins job had been updated. Things are looking fine for me locally again, I've pushed it.

PR #901, opened 2018-02-17
add golang 1.10

rphillips:


rphillips:
could put quotes around 1.10, but I like having the latest stable 1.10

Issue #902, opened 2018-02-21
bootkube recovery doesn't set hostNetwork: true on recovered controller-manager and scheduler assets.


coresolve:
Command I used to test: 
```
docker run -it -v /etc/ssl:/etc/ssl \
-v /root/recovery:/recovery quay.io/coreos/bootkube:v0.8.1 \
/bootkube recover --etcd-ca-path /etc/ssl/etcd/ca.crt \
--etcd-certificate-path /etc/ssl/etcd/client.crt \
--etcd-private-key-path /etc/ssl/etcd/client.key \
--recovery-dir /recovery/now \
--etcd-servers=https://testrecovery-etcd-0.dcooley.se.k8s.work:2379 --kubeconfig=/recovery/kubeconfig
```

Then grep for hostNet:
```
ip-10-0-42-175 now # grep -ir hostNetwork bootstrap-manifests/*
bootstrap-manifests/bootstrap-kube-apiserver.yaml:  hostNetwork: true
```

This means that the assets created by recovery will not work for recovery as there is nothing to satisfy the cni implementation without a controller manager and a scheduler.

coresolve:
@diegs

PR #905, opened 2018-02-22
bootkube/pkg/recovery/recover fix hostNetwork

coresolve:
Fixes: https://github.com/kubernetes-incubator/bootkube/issues/902
Adds hostNetwork: true to all generated assets and tests for same.

### DIFF
--- a/pkg/recovery/recover_test.go
+++ b/pkg/recovery/recover_test.go
@@ -203,6 +203,23 @@ func TestFixUpBootstrapPods(t *testing.T) {
 			APIVersion: "v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


diegs:
ok to test

pkg/recovery/recover_test.go
```diff
@@ -186,6 +186,7 @@ func TestFixUpBootstrapPods(t *testing.T) {
 					ReadOnly:  true,
 				}},
 			}},
+			HostNetwork: true,
```
diegs: Should actually not set these in the `pods` array--the `fixUpBootstrapPods` method should add them.

rphillips:
coreosbot run e2e calico

PR #906, opened 2018-02-22
checkpointer: change activeGracePeriod logic.

diegs:
This changes activeGracePeriod to move to inactiveGracePeriod when the
apiserver is up. This is to allow the checkpointer to get out of the
way of the real apiserver when it comes back, while still protecting
against that real apiserver going back down shortly after.

Before this change, a checkpoint could remain active for the entire
grace period unnecessarily if it was replaced by a new pod with a
different name that was successfully running.

After this change, a checkpoint will only get cleaned up after the
apiserver has been up and stable for an entire grace period's length.


diegs:

diegs:
Looks like the e2e checkpointer test no longer works:

```
+ ./terraform apply
1 error(s) occurred:

* provider.aws: : invalid or unknown key: version
```

ericchiang:
cc @rphillips any idea what that's about? ^

rphillips:

rphillips:
updated terraform to 0.11.3 on jenkins.

rphillips:

PR #908, opened 2018-02-22
recover: relax verification of etcd CAPath

### LINKED ISSUE #903: bootkube recovery requires an etcd client cert when used with an ca cert.
bootkube recovery requires that you provide a etcd-client cert and key when you specify a etcd-ca-cert.

This should not be a requirement. Etcd can be secured by a CA and not require client certs.

rphillips:
etcd can be secured with only a CA, and not require client certificates. This
fix relaxes the argument verification.

Fixes #903

### DIFF
--- a/cmd/bootkube/recover.go
+++ b/cmd/bootkube/recover.go
@@ -88,8 +88,8 @@ func validateRecoverOpts(cmd *cobra.Command, args []string) error {
 	if recoverOpts.recoveryDir == "" {
 		return errors.New("missing required flag: --recovery-dir")
 	}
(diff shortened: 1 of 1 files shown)


rphillips:
I will dig in, the lower level code needs some tweaks.

ericchiang:
Are there no tests for these packages? Could we add one?

diegs:
There are tests, but not for the `cmd` stuff... not sure if cobra has a good cmd testing framework?

rphillips:
After some research, there is not a great way with cobra to do argument testing.

ericchiang:
Was thinking something like

```
type recoverOpts struct {
    // ... fields
}

func validateRecoverOpts(o *recoverOpts) error {
    // ...
}
```

Then a test like

```
func TestValidateRecoverOpts(t *testing.T) {
    // ...
}
```

Looks like that'd be a big change though, so let's skip adding the tests.

ericchiang:
coreosbot run e2e calico

rphillips:
coreosbot run calico

rphillips:
coreosbot run e2e-calico

rphillips:
coreosbot run e2e calico

PR #915, opened 2018-03-05
jenkins: add initial bootkube-e2e jobs

colemickens:
Adds a new tree `hack/jenkins` to the tree:

```
hack/jenkins
├── images
│   └── bootkube-e2e
│       └── Dockerfile
├── jobs
│   └── bootkube_e2e.groovy
├── pipelines
│   └── bootkube-e2e
│       └── Jenkinsfile
└── scripts
    ├── e2e.sh
    ├── tqs-down.sh
    └── tqs-up.sh
```

This holds the majority of the configuration for all bootkube jobs. (Somethings are captured in our Jenkins configuration repo, instead of here, because they are "global" settings. With rare exception, all job-specific config is and belongs here.)

The jobs added here, `bootkube-e2e-{calico,flannel}` will execute on Kubernetes agents. This was chosen to minimize the amount of special care needed on the hosts. Even for a fairly self-contained project like `bootkube`, there is value in having a container with known-good versions of tools in place. Fortunately, if we need docker or something, we have a number of options, including provisioning "physical" agents, but I'd strongly prefer to avoid this.

This change will also enable easier testing to changes in our tests. Rarely should changes ever need to be made to the initial job Job DSL description, as most of the test logic is fully encoded in the Jenksfile or shell scripts. Thus a PR can modify them and the modifications will be tested as part of the test run. One could radically re-organize the hack/terraform/quickstart configs, for example, and still get test passes without needing manual merges or modifications to the Jenkins instance.

There's just a couple TODOs in here: changing the repo/bot accounts (will be done in a followup), as well as one last enhancement I'd like to make to `tqs-up.sh` to fully remove the `ssh-agent` stuff.

### DIFF
--- a/hack/jenkins/README.md
+++ b/hack/jenkins/README.md
@@ -1,0 +1,60 @@
+# hack/jenkins
+
+Job configurations, pipelines, scripts, and docker images used for `bootkube` validation.
+
+### Overvew
+
+This contains all specifications for `bootkube` Jenkins jobs. No changes are made to these jobs
+after they are deployed. They are re-deployed each time Jenkins is restarted. The goal is for
+Jenkins and its job configurations to be as fungible as possible.
+
+### Jenkins Quick Information
+
+-   Declarative, Verison-Controlled Jenkins Jobs
+    -   (via [Job DSL Plugin](https://github.com/jenkinsci/job-dsl-plugin))
+    -   It defines a groovy DSL for specifying Jenkins job in a declarative-looking imperative syntax that emits Jenkins job xml.
+    -   It provides a "Build Action" that is used in a "Seed Job" (defined in a separate repo) to instantiate jobs.
+    -   Jobs for this repo are defined in `./jobs/`.
+-   Jenkinsfile / Jenkins Pipelines
+    -   (built-in to Jenkins)
+    -   Jenkins has pipelines for defining workflows for jobs in a version-control friendly manner (as opposed to the non-friendly XML files that it uses internally)
+    -   Confusingly, they come in two variants - [Declarative Pipelines](https://jenkins.io/doc/book/pipeline/syntax/#declarative-pipeline) and [Scripted Pipelines](https://jenkins.io/doc/book/pipeline/syntax/#scripted-pipeline).
+        Be sure you are reading the docs for the right kind. They have slightly different steps, and even they have same-named steps, they can have behavioral differences.
+    -   This repository favors "Declarative Pipelines" where possible.
+        -   It seems to be where Jenkins is trending and trying to push people toward.
+        -   The syntax is more convincingly declarative.
+        -   It has an escape hatch that allows you to enter a scripting block and use the full scripting syntax.
+    -   Pipelines for this repo are defined in `./pipelines/`.
+-   Kubernetes Plugin
+    -   (via [Kubernetes Plugin](https://github.com/jenkinsci/kubernetes-plugin))
+    -   Allows using a Kubernetes cluster as a Jenkins "Cloud" (resource from which worker Agents can be spawned).
+    -   _Note:_ All jobs here are expected to run on the default `kubernetes` cloud.
+
+### Structure
+
+-   `images`: Contains `Dockerfile`s for any images used in validation. (_Note:_ images used for CI are not considered or supported as part of any bootkube release.)
+-   `jobs`: Contains top-level groovy scripts containing the Job DSL configurations. (these are recommeneded to be (but are not necessarily) Pipeline Jobs.)
+-   `pipelines`: Contains `Jenkinsfile` pipelines used as part of pipeline jobs.
+-   `scripts`: Contains scripts that are used in Pipelines, or in Jobs directly. (May or may not be usable outside of Jenkins. Only supported as part of CI.)
+
+This shows an example directory structure, added as part of the `bootkube-e2e-*` jobs:
+
+    hack/jenkins
+    ├── images
+    │   └── bootkube-e2e
+    │       └── Dockerfile
+    ├── jobs
+    │   └── bootkube_e2e.groovy
+    ├── pipelines
+    │   └── bootkube-e2e
+    │       └── Jenkinsfile
+    └── scripts
+        ├── e2e.sh
+        ├── tqs-down.sh
+        └── tqs-up.sh
+
+### Currently Defined Jobs
+
+-   bootkube-e2e-\*
+    -   calico: tests a standard multi-master bootkube cluster with calico
+    -   flannel: tests a standard multi-master bootkube cluster with flannel
--- a/hack/jenkins/images/bootkube-e2e/Dockerfile
+++ b/hack/jenkins/images/bootkube-e2e/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:xenial
+
+RUN apt-get update && apt-get upgrade -y && \
+    apt-get install -y unzip wget curl git make jq openssh-client && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN wget https://godeb.s3.amazonaws.com/godeb-amd64.tar.gz && \
+    tar xvzf godeb-amd64.tar.gz && \
+    ./godeb install 1.10 && \
+    rm -rf godeb* *.deb
+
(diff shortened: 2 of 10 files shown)


colemickens:
coreosbot run e2e

colemickens:
cc: @rithujohn191 (you're not in the org/repo yet, so I can't request a review directly)

hack/jenkins/pipelines/bootkube-e2e/Jenkinsfile
```diff
@@ -0,0 +1,81 @@
+pipeline {
```
ericchiang: nit: note what Jenkinsfile version this is using

hack/jenkins/jobs/bootkube_e2e.groovy
```diff
@@ -0,0 +1,61 @@
...
+    }
+    definition {
+      triggers {
+        githubPullRequest {
+          admins(job_admins)
+          userWhitelist(user_whitelist)
+          orgWhitelist(org_whitelist)
+          useGitHubHooks(true)
+          onlyTriggerPhrase(false)
+          triggerPhrase("coreosbot run ${job_name}")
```
ericchiang: nit: make sure this can't conflict with the existing e2e jobs so we can run them in parallel

colemickens:
coreosbot run e2e

hack/jenkins/images/bootkube-e2e/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:xenial
+
```
rphillips: we could do:

```
ARG GOLANG
ENV GOLANG ${GOLANG:-1.10}
ARG TERRAFORM_VERSION
ENV TERRAFORM_VERSION ${TERRAFORM_VERSION:-0.11.3}

File: hack/jenkins/pipelines/bootkube-e2e/Jenkinsfile [hunk 4]
```diff
@@ -0,0 +1,81 @@
...
+            sh "./hack/jenkins/scripts/tqs-up.sh"
+          }
+        }
+      }
+    }
+    stage('e2e') {
+      steps {
+        container("default") {
+          dir("${WORKDIR}") {
+            sh "./hack/jenkins/scripts/e2e.sh"
```
rphillips: we should execute these scripts with bash, since the shebangs are bash.

rphillips: nevermind... sh is part of the dsl

hack/jenkins/scripts/tqs-up.sh
```diff
@@ -0,0 +1,64 @@
...
+#export TF_VAR_instance_tags="${CLUSTER_NAME}" # unused??
+export TF_VAR_resource_owner="${CLUSTER_NAME}"
+export TF_VAR_ssh_public_key="$(cat "${IDENT}.pub")"
+export TF_VAR_ssh_key="jenkins-core-services"
+export TF_VAR_additional_masters=0
+export TF_VAR_num_workers=${NUM_WORKERS}
+export TF_VAR_region="${REGION}"
+
+# bring up compute
+terraform init
```
rphillips: ${TERRAFORM} here

hack/jenkins/images/bootkube-e2e/Dockerfile
ericchiang: I'm for hard coding these instead of relying on the developer's env.

rphillips: i'm ok with that as well.

hack/jenkins/pipelines/bootkube-e2e/Jenkinsfile
colemickens: The default shell for Jenkins is `/bin/bash` and `sh` is actually a pipeline command, rather than the unix `sh`. AFAICT everything is working as expected, including bash expansions.

hack/jenkins/images/bootkube-e2e/Dockerfile
colemickens: Will do asap, thanks, I should use `ARGS` more often.

hack/jenkins/scripts/tqs-up.sh
colemickens: :+1:

hack/jenkins/jobs/bootkube_e2e.groovy
colemickens: The prefix in `context()` assures this. In general, that `context` is the identifier to GitHub for commit statuses. (aka, any client can clobber another, yes, etc.)

ericchiang: sgtm, as long as it's documented here https://github.com/kubernetes-incubator/bootkube/blob/master/Documentation/development.md#running-pr-tests

hack/jenkins/images/bootkube-e2e/Dockerfile
colemickens: Hm, build args don't directly come from the environment, the dev would have to specify them. That having been said, for things like versions, I think hard-coding is better. (I've found build args to more be useful when using a single Dockerfile with slightly different internal behavior based on a build arg.)

I'm just going to leave it for now. If it's even useful to someone they can send it.

hack/jenkins/jobs/bootkube_e2e.groovy
colemickens: Ah. There's two things then. We don't want the build commit statuses to conflict, but we also want to trigger them independently.

In this case, I will be prefixing both the commit status, and the retrigger commands with "tku-", so that they are the same and easy to infer from one another, hopefully also alleviating the need to look them up as they'll be right there on the page.

colemickens:
@ericchiang @rphillips Thanks! Feedback addressed.

colemickens:
coreosbot run e2e

ericchiang:
```
11:27:17 terraform destroy -force
11:27:17 [bootkube-pr-e2e] $ /bin/bash /tmp/jenkins5410195304671793502.sh
11:27:17 var.ssh_public_key
11:27:17   SSH Public Key
11:27:17 
11:27:17   Enter a value: 
11:27:17 var.ssh_public_key
11:27:17   SSH Public Key
11:27:17 
11:27:17   Enter a value: 
11:27:17 var.ssh_public_key
11:27:17   SSH Public Key
11:27:17 
11:27:17   Enter a value: 
11:27:17 var.ssh_public_key
11:27:17   SSH Public Key
11:27:17 
11:27:17   Enter a value: 
11:27:17 
11:27:17 Error: Error asking for user input: missing required value for "ssh_public_key"
```

colemickens:
coreosbot run e2e

colemickens:
coreosbot run e2e

colemickens:
coreosbot run e2e
coreosbot run e2e-calico

colemickens:
coreosbot run e2e

colemickens:
coreosbot run e2e calico

colemickens:
BTW this has been passing for a while and should generally be ready to merge. @rithujohn191 / @rphillips please take another look.

hack/jenkins/README.md
```diff
@@ -0,0 +1,74 @@
+# hack/jenkins
+
+Job configurations, pipelienes, scripts, and docker images used for `bootkube` validation.
```
rithujohn191: pipelienes --> pipelines

hack/jenkins/README.md
```diff
@@ -0,0 +1,74 @@
...
+└── scripts
+    ├── e2e.sh
+    ├── tqs-down.sh
+    └── tqs-up.sh
+```
+
+### Currently Defined Jobs
+
+* bootkube-e2e-*
+    * calico: tests a standard multi-master bootkube cluster with calico
```
rithujohn191: Does this test for Calico IPIP overlay network as well?

colemickens: I'm not sure. It looks like it only tests the IPIP configuration? https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L1191

PR #919, opened 2018-03-12
update Kubernetes dependency to 1.9.3

ericchiang:
Kubernetes related changes:
* Replace kubectl import with custom code in pkg/bootkube/create.go
* API imports now refer to k8s.io/api
* Migrate codec uses to k8s.io/client-go/kubernetes/scheme.Codec

Gopkg.toml changes:
* Update Kubernetes to 1.9.3
* Move Kubernetes imports to their own block
* Various updates to gRPC and golang.org/x repos

We still need to import github.com/kubernetes/kubernetes since the
checkpointer uses the CRI gRPC interface advertised there. In the
future we should copy the proto file and generate it in this repo.


ericchiang:
Most of this is mechanical. The changes to `pkg/bootkube/create.go` needs the most eyes (maybe I should make that a separate commit?).

ericchiang:
Travis is failing on Go 1.8. I'm inclined to remove it instead of trying to debug since Go 1.8 isn't supported by the Go team anymore https://golang.org/doc/devel/release.html#policy

pkg/bootkube/create.go
```diff
@@ -205,24 +211,175 @@ func customResourceDefinitionKindURI(apiGroup, version, namespace, plural string
...
+
+// parseManifests parses a YAML or JSON document that may contain one or more
+// kubernetes resoures.
+func parseManifests(r io.Reader) ([]manifest, error) {
+	reader := yaml.NewYAMLReader(bufio.NewReader(r))
+	var manifests []manifest
+	for {
+		data, err := reader.Read()
+		if err != nil {
+			if err == io.EOF {
```
rphillips: this seems odd to me... usually we would want to read all the chunks until EOF, then parse the data.

ericchiang: `NewYAMLReader` is a little special. It's not a stream implementation, it returns discrete chunks of YAML https://godoc.org/k8s.io/apimachinery/pkg/util/yaml#YAMLReader.Read

Maybe renaming `data` to `rawManifest` would make that clearer?

The following Go idiom is also pretty common:

```go
if err != nil {
    if err == io.EOF {
        // successful return
    }
    // unsuccessful return
}
```

Take a look at the Go std library:

```
grep -r '== io.EOF' -A8 "$(go env GOROOT)/src/" | grep -v '_test.go' | less
```

pkg/bootkube/create.go
```diff
@@ -43,143 +59,133 @@ func CreateAssets(config clientcmd.ClientConfig, manifestDir string, timeout tim
...
-		obj, err := resource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)
-		if err != nil {
-			if apierrors.IsAlreadyExists(err) {
-				count++
-				return nil
-			}
-			return cmdutil.AddSourceToErr("creating", info.Source, err)
+func (c *creater) createManifests(manifests []manifest) error {
+	var crds, other []manifest
+	for _, m := range manifests {
```
aaronlevy: This is a bit of a change in behavior. Prior, if the resource it just created happened to be a CRD, it would then wait until the api resource was available (so we knew that creating subsequent instances of that CRD would be successful).

If we scan the list of manifests and create all CRDs upfront - that does remove some of the ability to ensure strict ordering of all asset creation (e.g. `01-foo, 02-foo`). Off the top of my head I don't know that this would be an issue - but I wonder if the minor speed improvement is worth removing that ability to order all object creation.

pkg/bootkube/create.go
```diff
@@ -205,24 +211,175 @@ func customResourceDefinitionKindURI(apiGroup, version, namespace, plural string
...
-	if err != nil {
-		return err
+
+	err = c.client.Post().
+		AbsPath(m.urlPath(plural)).
+		Body(m.raw).
+		SetHeader("Content-Type", "application/json").
+		Do().Error()
+	if err != nil && errors.IsAlreadyExists(err) {
+		// retry with a PUT
```
aaronlevy: I believe the behavior of "do not replace if exists" is assumed in some places (e.g. recover) - so I'm not sure this is a safe assumption to make.

pkg/bootkube/create.go
rphillips: gotcha. Thanks for the explanation. parseManifests seems fine.

pkg/bootkube/create.go
```diff
@@ -43,143 +59,133 @@ func CreateAssets(config clientcmd.ClientConfig, manifestDir string, timeout tim
...
-	timeout = timeout - time.Since(start)
-	if err := wait.PollImmediate(5*time.Second, timeout, createFn); err != nil {
-		err = fmt.Errorf("Failed to create assets: %v", err)
-		glog.Error(err)
-		return err
+	if err := creater.createManifests(m); err != nil {
+		UserOutput("%v\n", err)
+		UserOutput("\nNOTE: Bootkube failed to create some cluster assets. It is important that manifest errors are resolved and resubmitted to the apiserver.\n")
+		UserOutput("For example, after resolving issues: kubectl create -f <failed-manifest>\n\n")
+		return fmt.Errorf("failed to create assets: %v", err)
```
aaronlevy: We previously had allowed bootkube to continue even if creation events failed. This was primarily because failing early actually made it more difficult (often times) to debug the issue (usually unrelated to actually bringing up a cluster). So if bootkube continued until completion (and left you with an accessible apiserver) you could then debug the failures through kubectl. If you just exit early - you might have nothing to look into.

This is also somewhat related to comment below about retrying the creation with a PUT (at very least shouldn't exit on already exist errors).

rphillips:
code looks good... the calico failure doesn't look like a flake.

ericchiang:
Calico failure is

```
16:56:54 Error: failed to create assets: creating /home/core/assets/manifests/calico-role.yaml ClusterRole kube-system/calico-node: the server could not find the requested resource
16:56:54 Tearing down temporary bootstrap control plane...
16:56:54 Error: failed to create assets: creating /home/core/assets/manifests/calico-role.yaml ClusterRole kube-system/calico-node: the server could not find the requested resource
16:56:54 Error: failed to create assets: creating /home/core/assets/manifests/calico-role.yaml ClusterRole kube-system/calico-node: the server could not find the requested resource
16:56:54 failed to create assets: creating /home/core/assets/manifests/calico-role.yaml ClusterRole kube-system/calico-node: the server could not find the requested resource
```

It looks like our calico resources are including a namespace on cluster wide resources?

```
$ head cluster/manifests/calico-role.yaml 
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: calico-node
  namespace: kube-system
rules:
  - apiGroups: [""]
    resources:
      - namespaces
    verbs:
```

I'll update the create code to accept this, then do a follow up PR to fix our rendered calico manfiests.

pkg/bootkube/create.go
ericchiang: The `manifest` type has the path of the file, so I'll sort in lexicographical order to preserve this behavior. Still plan on creating the CRDs before the other manifests.

pkg/bootkube/create.go
ericchiang: :+1: will modify to not error if creating the resource fails and avoid the PUT.

pkg/bootkube/create.go
```diff
@@ -43,143 +59,133 @@ func CreateAssets(config clientcmd.ClientConfig, manifestDir string, timeout tim
...
 		}
+		UserOutput("Created %s\n", m)
+	}
 
-		count++
-		shortOutput := false
-		if !shortOutput {
-			f.PrintObjectSpecificMessage(info.Object, util.GlogWriter{})
+	for _, crd := range crds {
+		if err := c.waitForCRD(crd); err != nil {
```
colemickens: What is `waitForCRD` doing?

ericchiang: It got brought over from the old code. It's just waiting for the API server to respond to a "list" on the custom resource. Will add a comment to the function.

colemickens:
coreosbot run tku-bootkube-e2e-calico
coreosbot run tku-bootkube-e2e-flannel

pkg/bootkube/create.go
```diff
@@ -43,143 +60,150 @@ func CreateAssets(config clientcmd.ClientConfig, manifestDir string, timeout tim
...
-		return err
 	}
-	if count == 0 {
-		return fmt.Errorf("no objects passed to create")
+
+	for _, m := range other {
+		create(m)
 	}
-	return nil
+	return true
```
ericchiang: Woops, this needs to be `return ok`. Going to wait for CI to go green before fixing.

ericchiang:
Comments addressed and just had one successful set of e2e runs. Going to merge on another green.

PR #922, opened 2018-03-13
bump to k8s 1.9.4

rphillips:


rphillips:
regression in 1.9.4. we should punt until a fixed release:

https://github.com/kubernetes/kubernetes/issues/61076

PR #924, opened 2018-03-14
Jenkins: Collect e2e logs & archive in public S3 bucket

colemickens:
jenkins: bootkube-e2e: collect, archive logs:
* Collects systemd+docker logs from the machines, as the existing Jenkins was doing.
* Archives all artifacts at the end, making them easily accessible from the job UI

jenkins: bootkube-e2e: upload logs to s3:
* Tees into log files in each stage (Jenkins makes it impossible to get the raw build log in a declarative pipeline from scm job, this was the best option I could come up with)
* Makes a `tar.gz` of all archived assets.
* Uploads the archived assets, and the `tar.gz` (convenience) in a public s3 bucket
* Sets the commit status to point to the public s3 bucket instead of to our private Jenkins

### DIFF
--- a/hack/jenkins/scripts/gather-logs.sh
+++ b/hack/jenkins/scripts/gather-logs.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+set -x
+set -euo pipefail
+
+export IDENT="${IDENT:-"${HOME}/.ssh/id_rsa"}"
+
+cd ${DIR}/../../terraform-quickstart/
+
+# TODO: unclear why ssh-agent isn't still running from tqs-up.sh...
+if [ -z "${SSH_AUTH_SOCK:-}" ] ; then
+  ssh-agent -s > "/tmp/bootkube-tqs-sshagent.env"
+  source "/tmp/bootkube-tqs-sshagent.env"
+  ssh-add "${IDENT}"
+fi
+
+./genlogs.sh
+./copylogs.sh
--- a/hack/jenkins/scripts/tqs-up.sh
+++ b/hack/jenkins/scripts/tqs-up.sh
@@ -23,8 +23,10 @@ if [ -z "${SSH_AUTH_SOCK:-}" ] ; then
   ssh-add "${IDENT}"
 fi
 
+set +x
 export TF_VAR_access_key_id="${ACCESS_KEY_ID}"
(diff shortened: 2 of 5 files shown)


hack/jenkins/scripts/tqs-up.sh
```diff
@@ -23,8 +23,10 @@ if [ -z "${SSH_AUTH_SOCK:-}" ] ; then
   ssh-add "${IDENT}"
 fi
 
+set +x
```
rphillips: we should remove the +x's around the credentials.

hack/jenkins/scripts/gather-logs.sh
```diff
@@ -0,0 +1,18 @@
...
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+set -x
+set -euo pipefail
+
+export IDENT="${IDENT:-"${HOME}/.ssh/id_rsa"}"
+
+cd ${DIR}/../../terraform-quickstart/
+
+# TODO: unclear why ssh-agent isn't still running from tqs-up.sh...
+if [ -z "${SSH_AUTH_SOCK:-}" ] ; then
```
rithujohn191: Is the next task to move the ssh-agent to the tqs-up.sh?

colemickens: It already is in `tqs-up.sh`. I don't trap/kill it, so it ought to still be running, but `ps aux | grep ssh` clearly shows that it isn't, so I just start it up again.

I probably will leave this workaround for the time being. My intuition is that it's another quirk of how the kubernetes-plugin works in Jenkins, and I feel it could take a lot of time to chase down with very little net benefit.

rithujohn191: Thanks for the clarification

rphillips: are we using the ssh agent plugin in jenkins? it tears down the agent after the 'build' step for some reason.

colemickens: No, in this case, I manually start the ssh-agent inside `tqs-up.sh` which should start a process and drop an `env` file. I checked to see if maybe it was just that the environment was getting reset between `stage(s)` but there was no relevant ssh process running at this point anyway, so somehow that process was getting cleaned up.

We could potentially consider using the sshAgent pipeline extension, but I'm reluctant to add back more plugins. Every single plugin, so far, has come with at least one sharp edge.

rphillips: Agreed... +1 to use the least amount of plugins

PR #925, opened 2018-03-14
create: fix newlines on missing UserOutputs

rphillips:
Performed a multi-node up. Something in create failed and the error was not flushed to stdout

### DIFF
--- a/pkg/bootkube/create.go
+++ b/pkg/bootkube/create.go
@@ -164,7 +164,7 @@ func (c *creater) createManifests(manifests []manifest) (ok bool) {
 	create := func(m manifest) {
 		if err := c.create(m); err != nil {
 			ok = false
-			UserOutput("Failed creating %s: %v", m, err)
+			UserOutput("Failed creating %s: %v\n", m, err)
 		} else {
 			UserOutput("Created %s\n", m)
 		}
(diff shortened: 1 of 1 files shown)


pkg/bootkube/create.go
```diff
@@ -164,7 +164,7 @@ func (c *creater) createManifests(manifests []manifest) (ok bool) {
 	create := func(m manifest) {
 		if err := c.create(m); err != nil {
 			ok = false
-			UserOutput("Failed creating %s: %v", m, err)
+			UserOutput("Failed creating %s: %v\n", m, err)
```
ericavonb: Since everywhere that calls `UserOutput` ends with a newline, why don't we have the function include it instead?

rphillips: In some cases, two newlines are passed for display purposes, and knowing the function adds newlines does not conform to the fmt.Printf API (Printf does not add newlines by default). Being explicit on the output formatting is generally better, I think.

ericchiang: I'd be happy with replacing this with fmt.Printf (so we'd get vet warnings) or a logger struct that we passed. UserOutput feels a bit unnecessary.

rphillips:
added an issue https://github.com/kubernetes-incubator/bootkube/issues/936

PR #927, opened 2018-03-19
pkg/asset: Migrate from Calico v2.6.6 to to 3.0.3

rithujohn191:
Upstream changes: [v3.0.3](https://github.com/projectcalico/calico/releases/tag/v3.0.3)


ericchiang:
Are these manifests copied from an upstream source? Can we add a link to the templates.go file?

rithujohn191:
Upstream Calico template can be found [here](https://github.com/projectcalico/calico/blob/master/master/getting-started/kubernetes/installation/hosted/kubernetes-datastore/calico-networking/1.7/calico.yaml).

NOTE: [Typha docs](https://github.com/projectcalico/typha) state that the feature is listed as beta so I have not included the Typha service manifest. 
Also `[hostendpoints](https://github.com/projectcalico/calico/blob/master/master/getting-started/kubernetes/installation/hosted/kubernetes-datastore/calico-networking/1.7/calico.yaml#L396-L408)` support is targeted for v3.1.0: [](https://github.com/projectcalico/calico/pull/1651) so I have not included it in the template file.

ericchiang:
```
Upstream Calico template can be found here.

NOTE: Typha docs state that the feature is listed as beta so I have not included the Typha service manifest.
Also hostendpoints support is targeted for v3.1.0: projectcalico/calico#1651 so I have not included it in the template file.
```

Awesome. Mind including that as a comment in the code? :)

colemickens:
Looking at the docs for Typha, I see this:

>Once you have a Typha service running, you can tell Felix v2.3.0+ (calico/node:v1.3.0+) to connect to it by setting the following environment variable in your calico/node pod spec, which tells Felix to discover Typha using the Kubernetes service API:

    - name: FELIX_TYPHAK8SSERVICENAME
       value: "calico-typha"

But, it looks like you are setting that in the diff (from the configMap). Is that intentional / safe? Maybe there's another option that you set that disables the Typha feature entirely?

colemickens:
(Sorry, wasn't trying to close, hit the wrong button...)

rithujohn191:
@colemickens but the `calico-config` has the following value set:
```
+  typha_service_name: "none"
```

So the service would not start. But if its safer to just leave the whole thing out and add it at a later stage I could remove it

colemickens:
That's good enough for me, I didn't see that (I guess `"none"` is special cased...). No reason to change it if it works, in my view.

pkg/asset/k8s.go
```diff
@@ -94,9 +98,14 @@ func newDynamicAssets(conf Config) Assets {
 			MustCreateAssetFromTemplate(AssetPathCalicoRoleBinding, internal.CalicoRoleBindingTemplate, conf),
 			MustCreateAssetFromTemplate(AssetPathCalicoSA, internal.CalicoServiceAccountTemplate, conf),
 			MustCreateAssetFromTemplate(AssetPathCalicoPolicyOnly, internal.CalicoPolicyOnlyTemplate, conf),
-			MustCreateAssetFromTemplate(AssetPathCalicoBGPConfigsCRD, internal.CalicoBGPConfigsCRD, conf),
-			MustCreateAssetFromTemplate(AssetPathCalicoFelixConfigsCRD, internal.CalicoFelixConfigsCRD, conf),
+			MustCreateAssetFromTemplate(AssetPathCalicoBGPConfigurationsCRD, internal.CalicoBGPConfigurationsCRD, conf),
+			MustCreateAssetFromTemplate(AssetPathCalicoBGPPeersCRD, internal.CalicoBGPPeersCRD, conf),
+			MustCreateAssetFromTemplate(AssetPathCalicoFelixConfigurationsCRD, internal.CalicoFelixConfigurationsCRD, conf),
+			MustCreateAssetFromTemplate(AssetPathCalicoGlobalNetworkPoliciesCRD, internal.CalicoGlobalNetworkPoliciesCRD, conf),
+			MustCreateAssetFromTemplate(AssetPathCalicoGlobalNetworkSetsCRD, internal.CalicoGlobalNetworkSetsCRD, conf),
```
colemickens: So did we just not support global network policies before now? Or are these new features to Calico 3.x?

rithujohn191: It seems to be a new resource in v3 according to the docs: https://docs.projectcalico.org/v3.0/reference/calicoctl/resources/globalnetworkpolicy

colemickens: :+1: thanks!

rithujohn191:
@ericchiang  added a comment about disabling Typha and the reasoning for it

pkg/asset/internal/templates.go
```diff
@@ -1192,14 +1193,17 @@ spec:
...
+                  name: calico-config
+                  key: typha_service_name
             - name: NODENAME
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
             - name: IP
-              valueFrom:
-                fieldRef:
-                  fieldPath: status.podIP
```
ericchiang: I think there was a reason this was here. Probably something to do with vagrant examples and such. Why was this changed? We should probably test the local VM examples.

rithujohn191: This is part of the v3 changes: https://github.com/projectcalico/calico/pull/1588/files

ericchiang: Okay thanks 👍

Issue #932, opened 2018-03-21
reports "failed to create assets", but no output about which


aaronlevy:
using /hack/multi-node:

```
Creating self-hosted assets...
Created /home/core/cluster/manifests/csr-approver-role-binding.yaml ClusterRoleBinding system-bootstrap-approve-node-client-csr
Created /home/core/cluster/manifests/csr-bootstrap-role-binding.yaml ClusterRoleBinding system-bootstrap-node-bootstrapper
Created /home/core/cluster/manifests/csr-renewal-role-binding.yaml ClusterRoleBinding system-bootstrap-node-renewal
Created /home/core/cluster/manifests/flannel-cfg.yaml ConfigMap kube-system/kube-flannel-cfg
Created /home/core/cluster/manifests/flannel-cluster-role-binding.yaml ClusterRoleBinding flannel
Created /home/core/cluster/manifests/flannel-cluster-role.yaml ClusterRole flannel
Created /home/core/cluster/manifests/flannel-sa.yaml ServiceAccount kube-system/flannel
Created /home/core/cluster/manifests/flannel.yaml DaemonSet kube-system/kube-flannel
Created /home/core/cluster/manifests/kube-apiserver-secret.yaml Secret kube-system/kube-apiserver
Created /home/core/cluster/manifests/kube-apiserver.yaml DaemonSet kube-system/kube-apiserver
Created /home/core/cluster/manifests/kube-controller-manager-disruption.yaml PodDisruptionBudget kube-system/kube-controller-manager
Created /home/core/cluster/manifests/kube-controller-manager-role-binding.yaml ClusterRoleBinding controller-manager
Created /home/core/cluster/manifests/kube-controller-manager-secret.yaml Secret kube-system/kube-controller-manager
Created /home/core/cluster/manifests/kube-controller-manager-service-account.yaml ServiceAccount kube-system/kube-controller-manager
Created /home/core/cluster/manifests/kube-controller-manager.yaml Deployment kube-system/kube-controller-manager
Created /home/core/cluster/manifests/kube-dns-deployment.yaml Deployment kube-system/kube-dns
Created /home/core/cluster/manifests/kube-dns-svc.yaml Service kube-system/kube-dns
Created /home/core/cluster/manifests/kube-proxy-role-binding.yaml ClusterRoleBinding kube-proxy
Created /home/core/cluster/manifests/kube-proxy-sa.yaml ServiceAccount kube-system/kube-proxy
[...]

PR #933, opened 2018-03-21
pkg/bootkube: fix create always printing an error

ericchiang:
closes https://github.com/kubernetes-incubator/bootkube/issues/932

### DIFF
--- a/pkg/bootkube/create.go
+++ b/pkg/bootkube/create.go
@@ -145,6 +145,7 @@ func newCreater(c *rest.Config) (*creater, error) {
 }
 
 func (c *creater) createManifests(manifests []manifest) (ok bool) {
+	ok = true
 	// Bootkube used to create manifests in named order ("01-foo" before "02-foo").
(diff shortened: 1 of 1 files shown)


colemickens:
Has this been broken for some time then?

ericchiang:
@colemickens yep!

colemickens:
Okay. Did we miss it because bootkube is exitting with zero? Is that another bug? Or did we just miss it because `bootkube render` is happening remotely in our e2e and it silently keeps going?

rphillips:
We continue running if manifest creation fails; it isn't supposed to be a hard error based on the comment here:

https://github.com/kubernetes-incubator/bootkube/blob/3fd1fd850ec3189df3445ee1913cf03426a65631/pkg/bootkube/create.go#L75-L76

aaronlevy:
Opened https://github.com/kubernetes-incubator/bootkube/issues/937 as we should probably surface these errors at least to e2e -- could revisit the behavior of ignoring in normal case as well

Issue #934, opened 2018-03-21
Checkpointer fails to get pod status using v1.10-rc1


aaronlevy:
Getting repeated errors:

```
E0321 23:44:16.429921       1 runtime_service.go:47] failed to list running sandboxes: rpc error: code = 12 desc = unknown service runtime.RuntimeService
E0321 23:44:21.447718       1 runtime_service.go:125] ListPodSandbox with filter "&PodSandboxFilter{Id:,State:&PodSandboxStateValue{State:SANDBOX_READY,},LabelSelector:map[string]string{},}" from runtime sevice
 failed: rpc error: code = 12 desc = unknown service runtime.RuntimeService
E0321 23:44:21.447821       1 runtime_service.go:47] failed to list running sandboxes: rpc error: code = 12 desc = unknown service runtime.RuntimeService
E0321 23:44:26.461873       1 runtime_service.go:125] ListPodSandbox with filter "&PodSandboxFilter{Id:,State:&PodSandboxStateValue{State:SANDBOX_READY,},LabelSelector:map[string]string{},}" from runtime sevice
 failed: rpc error: code = 12 desc = unknown service runtime.RuntimeService
```

The checkpointer then never pivots from the checkpointed copy of the apiserver, to the "real" apiserver pod.

Tested using v1.9.3 and it did not have this same behavior

ericchiang:
This error is coming from our use of CRI

https://github.com/kubernetes-incubator/bootkube/blob/8a9a588f6c19223f00c6c91d88c307e1387eb288/pkg/checkpoint/runtime_service.go#L107-L115

ericchiang:
The CRI service switched from v1alpha1 to v1alpha2 between 1.9 and 1.10

https://github.com/kubernetes/kubernetes/tree/v1.9.6/pkg/kubelet/apis/cri
https://github.com/kubernetes/kubernetes/tree/v1.10.0-rc.1/pkg/kubelet/apis/cri

ericchiang:
CRI made a breaking change over 1.10 https://github.com/kubernetes/kubernetes/pull/58973

https://github.com/kubernetes-incubator/cri-tools/issues/236

From what I can see this is expected. I don't see a way for us to support both 1.9 and 1.10 CRI.

PR #938, opened 2018-03-22
pkg/checkpoint: support both CRI v1alpha1 and v1alpha2

ericchiang:
Updates https://github.com/kubernetes-incubator/bootkube/issues/934

Generated proto files taken from the following paths:

https://github.com/kubernetes/kubernetes/tree/v1.9.6/pkg/kubelet/apis/cri
https://github.com/kubernetes/kubernetes/tree/v1.10.0-rc.1/pkg/kubelet/apis/cri


rphillips:
could you add a README in the v1alpha1 directory where those files came from?

ericchiang:
> could you add a README in the v1alpha1 directory where those files came from?

Yeah will do. Going to let CI see if this works first.

pkg/checkpoint/runtime_service.go
```diff
@@ -93,53 +88,123 @@ func (r *remoteRuntimeService) localRunningPods() map[string]*v1.Pod {
...
+				},
+			},
+		})
+		if err != nil {
+			glog.Errorf("ListContainers with filter from runtime service failed: %v", err)
+			return nil, err
+		}
+		for _, c := range resp.Containers {
+			if c.Metadata == nil {
+				glog.V(4).Infof("Container does not have metadata: %+v", c)
```
colemickens: Why do we skip here? When does this happen?

ericchiang: Just copied from the previous behavior.

colemickens: Yeah, I saw the removed block too, was just curious. not a big deal.

ericchiang:
Testing this locally with v1.10-rc.1

ericchiang:
This appears to work. I was able to get a v1.10-rc.1 cluster running and the checkpointer behaves fine.

pkg/checkpoint/cri/README.md
```diff
@@ -0,0 +1,10 @@
+# Vendored CRI APIs
+
+We copy generated CRI APIs for a couple reasons:
+* When CRI promotes a version, it sometimes deletes the old alpha version.
+* Prevents importing `k8s.io/kubernetes`.
```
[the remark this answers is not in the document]
ericchiang: I was gonna do that in a follow up (or let someone else do it). It's a much bigger diff :)

PR #939, opened 2018-03-22
add strict mode for start

### LINKED ISSUE #937: Surface asset creation errors to e2e tests
So we don't end up with issues like https://github.com/kubernetes-incubator/bootkube/issues/932 - our e2e tests should expose any asset creation errors as test failures.

We don't fail on these errors in the normal case as it's usually easier to debug the resulting cluster if we don't fail early (https://github.com/kubernetes-incubator/bootkube/pull/933#issuecomment-375161442).

Not sure best course here - maybe something like a "strict" mode on start that fails on any error? The bash -set euo equivalent?

rphillips:
Fixes #937

### DIFF
--- a/cmd/bootkube/start.go
+++ b/cmd/bootkube/start.go
@@ -21,6 +21,7 @@ var (
 	startOpts struct {
 		assetDir        string
 		podManifestPath string
+		strict          bool
 	}
 )
 
@@ -28,12 +29,14 @@ func init() {
 	cmdRoot.AddCommand(cmdStart)
 	cmdStart.Flags().StringVar(&startOpts.assetDir, "asset-dir", "", "Path to the cluster asset directory. Expected layout genereted by the `bootkube render` command.")
 	cmdStart.Flags().StringVar(&startOpts.podManifestPath, "pod-manifest-path", "/etc/kubernetes/manifests", "The location where the kubelet is configured to look for static pod manifests.")
+	cmdStart.Flags().BoolVar(&startOpts.strict, "strict", false, "Strict mode will cause bootkube to exit early if any manifests in the asset directory cannot be created.")
 }
 
 func runCmdStart(cmd *cobra.Command, args []string) error {
 	bk, err := bootkube.NewBootkube(bootkube.Config{
 		AssetDir:        startOpts.assetDir,
 		PodManifestPath: startOpts.podManifestPath,
+		Strict:          startOpts.strict,
 	})
 	if err != nil {
 		return err
--- a/pkg/bootkube/create.go
+++ b/pkg/bootkube/create.go
@@ -33,7 +33,7 @@ const (
 	crdRolloutTimeout  = 2 * time.Minute
 )
 
-func CreateAssets(config clientcmd.ClientConfig, manifestDir string, timeout time.Duration) error {
+func CreateAssets(config clientcmd.ClientConfig, manifestDir string, timeout time.Duration, strict bool) error {
 	if _, err := os.Stat(manifestDir); os.IsNotExist(err) {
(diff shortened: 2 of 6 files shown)


pkg/bootkube/create.go
```diff
@@ -164,23 +170,28 @@ func (c *creater) createManifests(manifests []manifest) (ok bool) {
...
 		}
+		UserOutput("Created %s\n", m)
+		return nil
 	}
 
 	// Create all namespaces first
 	for _, m := range namespaces {
-		create(m)
+		if err := create(m); err != nil && c.strict {
+			return
```
colemickens: Do you need to set ok=false here?

Or better, I think we want to explicitly `return false` here, right?

rphillips: thanks. fixed it to explicitly return false

cmd/bootkube/start.go
```diff
@@ -21,19 +21,22 @@ var (
...
 		podManifestPath string
+		strict          bool
 	}
 )
 
 func init() {
 	cmdRoot.AddCommand(cmdStart)
 	cmdStart.Flags().StringVar(&startOpts.assetDir, "asset-dir", "", "Path to the cluster asset directory. Expected layout genereted by the `bootkube render` command.")
 	cmdStart.Flags().StringVar(&startOpts.podManifestPath, "pod-manifest-path", "/etc/kubernetes/manifests", "The location where the kubelet is configured to look for static pod manifests.")
+	cmdStart.Flags().BoolVar(&startOpts.strict, "strict", false, "Strict mode.")
```
ericchiang: nit: More of a description than just "Strict mode". Maybe

> When run in strict mode bootkube fails if it can't create any of the manifests in the asset directory.

aaronlevy:
Can we have the flag used by default only on the tools used in e2e? e.g. I don't think it needs to be in the hack dir (I pretty regularly would mess up manifests during dev - and then want to debug them live vs having the cluster being taken down).

rphillips:
@aaronlevy fixed that up. I needed to add a BOOTKUBE_OPTS variable so jenkins can pass down the strict flag during the e2e tests.

ericchiang:
@colemickens is the correct place to make the Jenkins changes here? https://github.com/kubernetes-incubator/bootkube/blob/master/hack/jenkins/pipelines/bootkube-e2e/Jenkinsfile

rphillips:
added the environment option to the jenkinsfile

colemickens:
@ericchiang Yes, that's where I would put it. You could parameterize it up through the job, rather than the pipeline, but that's only useful if you wanted to queue builds with strict mode off. For the e2e jobs, I would probably have done the same and put it in the Jenkinsfile as it is now. LGTM.

cmd/bootkube/start.go
```diff
@@ -21,19 +21,22 @@ var (
...
 		podManifestPath string
+		strict          bool
 	}
 )
 
 func init() {
 	cmdRoot.AddCommand(cmdStart)
 	cmdStart.Flags().StringVar(&startOpts.assetDir, "asset-dir", "", "Path to the cluster asset directory. Expected layout genereted by the `bootkube render` command.")
 	cmdStart.Flags().StringVar(&startOpts.podManifestPath, "pod-manifest-path", "/etc/kubernetes/manifests", "The location where the kubelet is configured to look for static pod manifests.")
+	cmdStart.Flags().BoolVar(&startOpts.strict, "strict", false, "Strict mode will cause bootkube to exit early if any manifests in the asset directory cannot be started.")
```
aaronlevy: nit: s/cannot be started/cannot be created/

rphillips: good comment. tweaked.

PR #943, opened 2018-03-28
Documentation/root-requirements: API server no longer needs root

wking:
Catch up with #789.


wking:
Heh, maybe there are a number of @coreosbot running in parallel?

PR #946, opened 2018-03-28
DCO: Remove unused file

wking:
This originally landed in #3 along with a reference from `CONTRIBUTING.md`.  That `CONTRIBUTING.md` reference was removed in #255, where it was replaced with references to Linux Foundation [CNCF CLAs][1].  Since we no longer use the DCO, this commit removes the local copy.

This doesn't quite match with the [CNCF charter][2], which has (§11.c):

> All new inbound code contributions to the CNCF shall be (i) accompanied by a Developer Certificate of Origin sign-off (http://developercertificate.org) and (ii) made under the Apache License, Version 2.0 (available at http://www.apache.org/licenses/LICENSE-2.0), such license to be in addition to, and shall not supersede, obligations undertaken under the contribution license agreement(s) provided for in (b) above.

But perhaps we want to punt on resolving that until [kubernetes-template-project][3] decides how it should be handled.

[1]: https://github.com/cncf/cla
[2]: https://www.cncf.io/about/charter/
[3]: https://github.com/kubernetes/kubernetes-template-project


wking:
> But perhaps we want to punt on resolving that until kubernetes-template-project decides how it should be handled.

I've filed kubernetes/kubernetes-template-project#16 upstream for this.

wking:
> Going to wait on the result of kubernetes/kubernetes-template-project#16

It was [closed][1] with "I don't think there is any intention to add a DCO requirement to Kubernetes given the CLA requirement...".  I think that's a green light for this PR, although you might want to wait to see if the CNCF charter wording gets updated.

[1]: https://github.com/kubernetes/kubernetes-template-project/issues/16#issuecomment-411148950

PR #948, opened 2018-03-28
Makefile: Drop hard-coded Bash and simplify gofmt

wking:
I'm reverting #870 here and also removing the code that seemed to have motivated it.  Details on the individual changes in the commit messages.  The `find` commit is something of a drive-by; I'm happy to split it out into its own pull request if that would make review easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ export GOARCH:=amd64
 export PATH:=$(PATH):$(PWD)
 
 LOCAL_OS:=$(shell uname | tr A-Z a-z)
-GOFILES:=$(shell find . -name '*.go' | grep -v -E '(./vendor)')
+GOFILES:=$(shell find . -name '*.go' ! -path './vendor/*')
 LDFLAGS=-X github.com/kubernetes-incubator/bootkube/pkg/version.Version=$(shell $(CURDIR)/build/git-version.sh)
 TERRAFORM:=$(shell command -v terraform 2> /dev/null)
 
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,7 @@ release: \
 	check \
 	_output/release/bootkube.tar.gz \
 
-check:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -41,6 +39,10 @@ endif
 	@./scripts/verify-gopkg.sh
 	@go test -v $(shell go list ./... | grep -v '/vendor/\|/e2e')
 
+gofmt:
+	gofmt -s -w $(GOFILES)
```
ericchiang: We might be able to replace this with just `go fmt ./...`

ericchiang: It automatically does as of 1.9 :)

See https://golang.org/doc/go1.9#vendor-dotdotdot

wking: > It automatically does as of 1.9 :)

Are we comfortable making that our minimum release, as documented [here][1]?  I see we dropped 1.8 from Travis in b64b287dfe643 (#919).

[1]: https://github.com/kubernetes-incubator/bootkube/blob/bff16e2afc34df7469e7a108cf5d87098809249a/Documentation/development.md#requirements

ericchiang:
Yeah, let's bump it to 1.10 and add a comment to the travis file to update
that number when we're bumping the travis versions. This can be a separate
PR.

Makefile
wking: > We might be able to replace this with just `go fmt ./...`

I was taking another look at this now that #955 has bumped our minimum Go version.  `go fmt …` calls `gofmt` with `-l -w` by default, which means we'd be dropping the `-s` we're currently using.  Is that acceptable?  Or do you want to keep the current, direct `gofmt` invocation?

wking:
While I wait for maintainer feedback about [whether we need to keep `-s`][1], I've rebased this onto master (around #949, #956) with 99b62c23 → 9815a948.

[1]: https://github.com/kubernetes-incubator/bootkube/pull/948#discussion_r178906706

PR #949, opened 2018-03-29
Makefile: Respect GOBIN

wking:
`GOBIN` (documented [here][1]) defaults to `DIR/bin`, but it's a configurable variable in its own right.  This command updates our calculations, using `?=` (documented [here][2]) to avoid expensive shell operations when we don't actually need the result (because a different target was called or because the user provided their own value).

I've also used `go env GOPATH` to get the ([platform-specific][1]) default value when the environment variable is not set.

And I've used [`cut`][3] instead of [`awk`][4] to pull out the first component of `GOPATH`.  Both are in POSIX, but `cut` is a simpler tool for this particular problem.

This will have trivial context conflicts with the in-flight #948.  I'm happy to rebase the other if/when one of them lands.

[1]: https://golang.org/cmd/go/#hdr-GOPATH_environment_variable
[2]: https://www.gnu.org/software/make/manual/html_node/Setting.html
[3]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/cut.html
[4]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html

### DIFF
--- a/Documentation/development.md
+++ b/Documentation/development.md
@@ -3,15 +3,14 @@
 ## Requirements
 
 * Go 1.10+
-* Configured [GOPATH](http://golang.org/doc/code.html#GOPATH)
 
 ## Building
 
-First, clone the repo into the proper location in your $GOPATH:
+First, clone the repo into the proper location in your [`GOPATH`][GOPATH]:
 
 ```
 go get -u github.com/kubernetes-incubator/bootkube
-cd $GOPATH/src/github.com/kubernetes-incubator/bootkube
+cd $(go env GOPATH | cut -d: -f1)/src/github.com/kubernetes-incubator/bootkube
 ```
 
 Then build:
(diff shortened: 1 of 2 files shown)


wking:
> Should we just call `go install` instead?

That sounds good to me, but I ran into golang/go#18981 attempting it.  That issue was fixed by golang/go@8f70e1f8a91 in Go 1.10.  Are we comfortable bumping our minimum version?  Would you prefer on of the other workarounds mentioned in the upstream image?  Or are you ok with this PR as it stands until we are comfortable making Go 1.10 the minimum version?

ericchiang:
Yes. I think it's fine to bump to 1.10

wking:
Rebased around #955 and [converted to `go install …`][1] with ba3f8ef9 → e4e23885.

[1]: https://github.com/kubernetes-incubator/bootkube/pull/949#issuecomment-378292372

rphillips:
ok to test

wking:
Hmm, I seem to have missed the `go install …` commit (maybe in a rebase?) :/.  I'll file a fixup PR.

PR #953, opened 2018-04-02
add hyperkube release scripts

rphillips:
* Builds and optionally pushes the built hyperkube image


hack/jenkins/pipelines/hyperkube-release/Jenkinsfile
```diff
@@ -0,0 +1,30 @@
+
+podTemplate(label: 'hyperkube-build',
```
colemickens: If you're only using a single container, is it worth trying to switch back to the declarative pipeline format? Maybe it doesn't matter as much here, since we don't have critical cleanup steps that must run (like we do with running terraform destroy in the e2e jobs).

ericchiang:
Note: we should consider finding a different long term home for this. Bootkube is fine for now.

colemickens:
Can you update the README to list the new job(s)?

rphillips:
tweaked the README.

PR #954, opened 2018-04-03
bump to 1.10.0

rphillips:


rphillips:
testing on AWS

ericchiang:
/hold

blocked on https://github.com/kubernetes-incubator/bootkube/pull/957

dghubble:
@rphillips there were a few apiserver flags you might consider tweaking. Not technically required, but they're mentioned in 1.10 deprecation notes. https://github.com/poseidon/terraform-render-bootkube/commit/5f3546b66ffb9946b36e612537bb6a1830ae7746

dghubble:
Btw, admission control order doesn't matter now. I tried various orders and the apiserver logs mention the order its actually going to use. I chose to pass them in that same order so nobody gets worried.

rphillips:
@dghubble thanks, I'll get those in. Currently having an issue with 1.10 on AWS. The KubernetesCluster tag in this PR is required currently. I didn't notice a breaking change in the changelog.

rphillips:
this PR is ready for another review... tags have been created correctly...

I still have a pending question with sig-aws on if there is an upgrade path for old clusters.

rphillips:
using /proc/sys/kernel/random/uuid for the UUID, since uuidgen does not exist.

rphillips:
coreosbot run e2e

rphillips:
Looks like the new builders did not run.

hack/jenkins/scripts/tqs-up.sh
```diff
@@ -27,12 +28,17 @@ set +x
 export TF_VAR_access_key_id="${ACCESS_KEY_ID}"
 export TF_VAR_access_key="${ACCESS_KEY_SECRET}"
 set -x
+export TF_VAR_kubernetes_id="$(cat /proc/sys/kernel/random/uuid)"
```
ericchiang: Please add a comment here.

rphillips: commented

PR #955, opened 2018-04-03
Documentation/development: Bump minimum Go version to 1.10

wking:
This [gets us][2] a fixed `CGO_ENABLED=0 go install`][1] and avoids 1.8 (which is [no longer supported upstream now that 1.10 has been released][3]).

I've also dropped the version information from the build comment (it's mentioned just before in the “Requirements” section) to DRY things up a bit.

[1]: https://github.com/golang/go/issues/18981
[2]: https://github.com/golang/go/commit/8f70e1f8a91
[3]: https://golang.org/doc/devel/release.html#policy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ before_script:
   - unzip terraform_0.11.3_linux_amd64.zip
   - export PATH=$PWD:$PATH
 go:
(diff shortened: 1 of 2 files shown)


wking:
CC @ericchiang.

PR #956, opened 2018-04-03
Makefile: Actually use 'go install ...' for the install target

wking:
This was my intention with #949, but I seem to have missed the actual code :/.  Instead, that pull-request landed an earlier implementation which we intended to drop based on #949 [review][1].

This commit has the intended implementation.  For detailed motivation, see the description in dbf0b6a5.

[1]: https://github.com/kubernetes-incubator/bootkube/pull/949#issuecomment-378292372


wking:
CC @rphillips, @ericchiang.

rphillips:
thanks!

PR #957, opened 2018-04-03
Replace Jenkins `sh` with a `bash` step that sets `set -euo pipefail`

colemickens:
Jenkins is currently marking stages with failing scripts as "succeeded". It looks like this is happening because of our piping into `tee`.

Replace `sh` with a new, custom `bash` step that will `set -euo pipefail` initially to safely run scripts and surface failures correctly.

### DIFF
--- a/hack/jenkins/pipelines/bootkube-e2e/Jenkinsfile
+++ b/hack/jenkins/pipelines/bootkube-e2e/Jenkinsfile
@@ -1,5 +1,7 @@
 // Declarative Pipeline (used by `bootkube-e2e-*` jobs)
 
+def bash(String cmd) { sh("#/usr/bin/env bash\nset -euo pipefail\n${cmd}") }
+
 pipeline {
   agent {
     kubernetes {
@@ -37,8 +39,8 @@ pipeline {
     stage('checkout') {
       steps {
         // jnlp slave runs as "jenkins" user, use the escape hatch. (https://hub.docker.com/r/jenkins/slave/~/dockerfile/)
-        sh "chmod -R go+rw /home/jenkins"
-        sh "mkdir -p \"${ARTIFACT_DIR}\""
+        bash "chmod -R go+rw /home/jenkins"
+        bash "mkdir -p \"${ARTIFACT_DIR}\""
         dir("${WORKDIR}") {
           checkout scm
         }
@@ -47,21 +49,21 @@ pipeline {
     stage('build') {
       steps {
         dir("${WORKDIR}") {
-          sh "make release |& tee -a ${ARTIFACT_DIR}/build.log"
+          bash "make release |& tee -a ${ARTIFACT_DIR}/build.log"
         }
       }
     }
     stage('deploy') {
       steps {
         dir("${WORKDIR}") {
-          sh "./hack/jenkins/scripts/tqs-up.sh |& tee -a ${ARTIFACT_DIR}/deploy.log"
+          bash "./hack/jenkins/scripts/tqs-up.sh |& tee -a ${ARTIFACT_DIR}/deploy.log"
         }
       }
     }
     stage('e2e') {
       steps {
         dir("${WORKDIR}") {
-          sh "./hack/jenkins/scripts/e2e.sh |& tee -a ${ARTIFACT_DIR}/e2e.log"
+          bash "./hack/jenkins/scripts/e2e.sh |& tee -a ${ARTIFACT_DIR}/e2e.log"
         }
       }
     }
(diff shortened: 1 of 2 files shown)


wking:
> Jenkins is currently marking stages with failing scripts as "succeeded". It looks like this is happening because of our piping into `tee`.

Can you provide more details on this failure or link to the Jenkins build?  I'd rather stick with POSIX where possible (e.g. see #948), and am happy to help troubleshoot any issues with POSIX-shell compat.

colemickens:
@wking Can you explain why or point me to existing discussion of this?

I need Jenkins steps to execute with `pipefail` or someone to rewrite all of our scripts that work with `set -euo pipefail`.

The problem is that the piping we do [here](https://github.com/kubernetes-incubator/bootkube/blob/1bb203f1bff3fd0f64f88ca4538fb35e0aa5b35d/hack/jenkins/pipelines/bootkube-e2e/Jenkinsfile#L64) is swallowing failures. This needs to be resolved immediately. Please ping me in #bootkube in Slack if you want to discuss interactively, I can share details back here.

wking:
> Can you explain why or point me to existing discussion of this?

I just like sticking to standards ;).

> The problem is that the piping we do [here][1] is swallowing failures.

Ah, ok.

> This needs to be resolved immediately.

Don't wait on me, but I'll ping you in Slack if I find a way to stick with POSIX for this case that is simpler than [this][2] ;).

[1]: https://github.com/kubernetes-incubator/bootkube/blob/1bb203f1bff3fd0f64f88ca4538fb35e0aa5b35d/hack/jenkins/pipelines/bootkube-e2e/Jenkinsfile#L64
[2]: https://unix.stackexchange.com/a/76171

colemickens:
Updated. Explicitly tested with a failure case (seen with [this commit](https://github.com/colemickens/bootkube/commits/39eac210dcc841bf8f4d50281808ce046fb011bc) (until GitHub GCs it) and [this Jenkins failure](https://jenkins-kube-upstream.prod.coreos.systems/blue/organizations/jenkins/tku-bootkube-e2e-calico/detail/tku-bootkube-e2e-calico/125/pipeline)).

@rphillips ptal

@wking I've not added the change you sent in Slack, as I don't see it demonstrating that pipefailures are caught, and I have previously ruled against redirecting+catting as is a drastic change in the amount of visibility we get in Jenkins (streaming output, timestamps, etc)

wking:
> @wking I've not added the change you sent in Slack, as I don't see it demonstrating that pipefailures are caught…

```sh
echo hi >/tmp/test.log; EXIT=$?; cat /tmp/test.log && exit ${EXIT}
```

isn't catching pipe failures.  It's writing to the temp file, capturing the exit code, writing the temp file to stdout, and then failing if either the `cat` or the initial command (`echo` in the example above) failed.

> … and I have previously ruled against redirecting+catting as is a drastic change in the amount of visibility we get in Jenkins (streaming output, timestamps, etc)

Ah, I didn't realize folks were watching streaming output, and I didn't realize Jenkins was timestamping lines as they were written to stdout.  For a POSIX alternative that accomplishes that, you could turn my original proposal around and cache the exit code instead of stdout:

```sh
(echo hi; echo $?> /tmp/status) | tee >/tmp/test.log && exit $(cat /tmp/status)
```

Testing a few cases, success:

```console
$ sh -c '(echo hi; echo $?> /tmp/status) | tee >/tmp/test.log && exit $(cat /tmp/status)'
$ echo $?
0
```

Failing primary command:

```console
$ sh -c '(bad-command; echo $?> /tmp/status) | tee >/tmp/test.log && exit $(cat /tmp/status)'
sh: bad-command: command not found
$ echo $?
127
```

Failing `tee`:

```console
$ sh -c '(echo hi; echo $?> /tmp/status) | bad-command >/tmp/test.log && exit $(cat /tmp/status)'
sh: bad-command: command not found
$ echo $?
127
```

My only concern would be that if you have several tests running in parallel you'd want unique status filenames, but you could accomplish that in this case by naming them after the log files (`build.log` and `build.status`, `deploy.log` and `deploy.status`, etc.).

My proposal here is going to be slightly more typing per current `tee` pipe vs. what you have in 3992f278.  I'm not sure how much the maintainers care about staying POSIX compatible if you already have Bash installed on the Jenkins workers anyway.

colemickens:
I'm going to leave this implementation as-is, for now:

* We are not POSIX compliant in a number of other places (including places used in the flow that this PR involves), and we it's not a particular goal of the project.
* Our Jenkins agents all run in a container anywhere. For now, this is sufficiently portable.

PR #963, opened 2018-04-11
Remove deprecated admission plugin flag

odacremolbap:
As of kubernetes 1.10 `--admission-control` has been deprecated in favor of `--enable-admission-plugins`

This new flag is already in place for the bootstrap rendered manifest
https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L227

This PR also sets that flag for the apiserver template

Note: trying 1.9 clusters won't work using the new flag, this could be fixed doing just the opposite, replacing the new flag at the rendered bootstrap manifest with the old one, since it is still working. I'd stick to the new flag, and let 1.9 users modify the rendered manifests if using bootkube > 0.12.0

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -147,7 +147,7 @@ spec:
         command:
         - /hyperkube
         - apiserver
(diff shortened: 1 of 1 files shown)


colemickens:
@coreosbot test this please

colemickens:
@rphillips this LGTM but can you confirm? It looks like you updated the bootstrap apiserver; wanted to make sure it wasn't intentional that the real apiserver was left with the older flag.

rphillips:
missed that... lgtm

PR #983, opened 2018-05-17
pkg/poll: Add a Poll helper that respects Context

wking:
This is like the `k8s.io/apimachinery/pkg/util/wait` `Poll`, except it takes a `Context` argument to allow canceling for any reason.  The k8s utility only cancels the poller on timeouts.  This PR includes an initial commit adding the new package, and then provides follow-up commits porting portions of the repository over to the new package.  Hopefully this makes the full change easier to review, because the individual commit messages can go into more detail about smaller ports.  I'm also happy to split this up into per-commit PRs, or squash the PR down to a single commit, if that would be easier.

### DIFF
--- a/e2e/internal/e2eutil/testworkload/nginx_workload.go
+++ b/e2e/internal/e2eutil/testworkload/nginx_workload.go
@@ -1,10 +1,12 @@
 package testworkload
 
 import (
+	"context"
 	"fmt"
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/kubernetes-incubator/bootkube/pkg/poll"
 	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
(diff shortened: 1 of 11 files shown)


wking:
I've fixed a number of nits, and now with 9f51545 the tests are green :).

ericchiang:
What does this PR improve beside code refactoring? Can you articulate that a bit?

wking:
> What does this PR improve beside code refactoring?

Now you can set timeouts anywhere up the stack without having to pass timeout arguments around.  And it sets the stage for graceful ctrl-C teardown in similar, since you can cancel from anywhere in the stack.

The cost you pay for the new flexibility is the new package, which I can push upstream if you're willing to wait for the extra review cycle.

e2e/internal/e2eutil/testworkload/nginx_workload.go
```diff
@@ -241,11 +240,11 @@ func (n *Nginx) newNginxService() error {
...
 }
 
-func (n *Nginx) newPingPod(reachable bool) error {
+func (n *Nginx) newPingPod(ctx context.Context, reachable bool) error {
 	name := fmt.Sprintf("%s-ping-job-%s", n.Name, utilrand.String(5))
-	deadline := int64(PollTimeoutForNginx.Seconds())
 
-	cmd := fmt.Sprintf("wget --timeout 5 %s", n.Name)
+	var deadline int64 = 6
+	cmd := fmt.Sprintf("wget --timeout %d %s", deadline-1, n.Name)
```
rphillips: deadline is always (6-1=5) here, right?

wking: > deadline is always (6-1=5) here, right?

Yup, but this way it slides if/when you adjust `deadline`.  It will get optimized down to constants during compilation, no?  I haven't actually checked, but that would be a pretty basic compile-time optimization.

ericchiang:
> Now you can set timeouts anywhere up the stack without having to pass timeout arguments around. And it sets the stage for graceful ctrl-C teardown and similar, since you can cancel from anywhere in the stack.

@wking okay it sounds like change isn't trying to fix a specific timeout or add a feature, it just adds instrumentation.

I'm going to close this to prevent a drawn out code review process that deals with hypotheticals. If you have a specific fix or feature in mind that impacts the bootkube tool, please open an issue for that, and we can discuss possible implementations for resolving it.

Also, you might want to look at [wait.PollUntil](https://godoc.org/k8s.io/apimachinery/pkg/util/wait#PollUntil) which accepts a context's Done channel.

PR #984, opened 2018-05-17
pkg/asset/k8s: Evenly weight chars in newBootstrapToken

wking:
The `256 % len(charset)` remainder unevenly biased characters listed earlier in the charset array.  For example, 256 is `7 * 36 + 4`, so a 36-character charset that naively used `int(b) % len(charset)` would include the first four characters (0 through 3) 14% (`8/7`) more frequently than the other characters.

The new integer approach is unbiased and a more efficient use of entropy, because we no longer collapse 256 possible values into 36 possible values.  With the current 36-character token charset, this reduces our read from 22 bytes to 15 bytes.

`SetBytes` converts the full input slice into a big integer.  There may be a more efficient approach to this that consumes the input incrementally, but `SetBytes` is convenient.  We can always polish more later if we want a more efficient approach.

The endian-ness doesn't matter for our current use case, because the input data is random.  I've documented it to make independently calculating the expected test values easier, but we can flip the endian-ness of the internal helper now or later on if it's convenient to do so.

### DIFF
--- a/pkg/asset/k8s.go
+++ b/pkg/asset/k8s.go
@@ -5,6 +5,8 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"fmt"
+	"math"
+	"math/big"
 	"path/filepath"
 	"text/template"
 
@@ -117,15 +119,35 @@ const validBootstrapTokenChars = "0123456789abcdefghijklmnopqrstuvwxyz"
 // https://kubernetes.io/docs/admin/bootstrap-tokens/#token-format
 func newBootstrapToken() (id string, secret string, err error) {
 	// Read 6 random bytes for the id and 16 random bytes for the token (see spec for details).
-	token := make([]byte, 6+16)
-	if _, err := rand.Read(token); err != nil {
+	tokenLength := 6 + 16
+	charsetLength := len(validBootstrapTokenChars)
+	size := int(math.Ceil(float64(tokenLength) * math.Log2(float64(charsetLength)) / math.Log2(256)))
+
+	random := make([]byte, size)
+	if _, err := rand.Read(random); err != nil {
 		return "", "", err
 	}
 
-	for i, b := range token {
-		token[i] = validBootstrapTokenChars[int(b)%len(validBootstrapTokenChars)]
-	}
+	token := make([]byte, tokenLength)
+	encodeBytes(validBootstrapTokenChars, random, token)
 	return string(token[:6]), string(token[6:]), nil
+}
+
+// encodeBytes encodes the input byte slice into encoded byte slices.
+// The input is treated as a big-endian unsigned integer, and the
+// encoded slice represents the input value in little-endian unsigned
+// base len(charset).  The size of the encoded slice should be set
+// before calling this function; after consuming input, any "extra"
+// encoded slice length will be filled with charset[0].
+func encodeBytes(charset string, input []byte, encoded []byte) {
+	charsetLength := len(charset)
+	inputInt := big.NewInt(0).SetBytes(input)
+	base := big.NewInt(int64(charsetLength))
+	remainder := big.NewInt(0)
+	for i := range encoded {
+		inputInt.QuoRem(inputInt, base, remainder)
+		encoded[i] = charset[int(remainder.Int64())%charsetLength]
+	}
 }
 
 func newKubeConfigAssets(assets Assets, conf Config) ([]Asset, error) {
--- a/pkg/asset/k8s_test.go
+++ b/pkg/asset/k8s_test.go
@@ -1,0 +1,31 @@
+package asset
+
+import (
+	"testing"
+)
+
+func TestEncodeBytes(t *testing.T) {
+	token := make([]byte, 22)
+	random := []byte{ // 1324076502674089023697701463353265679
+		255,
+		2,
+		3,
+		4,
+		5,
+		6,
+		7,
+		8,
+		9,
+		10,
+		11,
+		12,
+		13,
+		14,
+		15,
+	}
+	encodeBytes(validBootstrapTokenChars, random, token)
+	expected := "n2r8j0ug4twvqmqt53hyff"
+	if string(token) != expected {
+		t.Errorf("%s != %s", string(token), expected)
+	}
+}


Sean-Q-Sun:
Thank you for your pull request. However, this seems like a very complicated approach for a very small amount of benefit. This issue has been fixed in kubernetes/kubernetes by discarding bytes that are greater or equal to 252. Would you consider changing your code to follow that example? See https://github.com/kubernetes/kubernetes/pull/57984/files

/cc @ericchiang

wking:
> This issue has been fixed in kubernetes/kubernetes by discarding bytes that are greater or equal to 252.

I'd considered that approach, but didn't feel comfortable with re-drawing when we were already drawing more random bytes than we needed ;).  But I didn't realize there was an upstream implementation.  I'll float my approach upstream and (maybe in parallel?) update this pull request to use the upstream package.

ericchiang:
@wking I think this is an over-optimization for a method that's called once every time you initialize a cluster.

wking:
> I think this is an over-optimization for a method that's called once every time you initialize a cluster.

Over-optimzations in shared packages shouldn't be a problem, right?  Might as well have an efficient implementation if you can share it between all Go k8s initializers.  But you can go ahead and close this, and I can re-open something later on if I get this landed in an upstream package.

pkg/asset/k8s.go
```diff
@@ -116,16 +118,38 @@ const validBootstrapTokenChars = "0123456789abcdefghijklmnopqrstuvwxyz"
...
 // https://kubernetes.io/docs/admin/bootstrap-tokens/#token-format
 func newBootstrapToken() (id string, secret string, err error) {
-	// Read 6 random bytes for the id and 16 random bytes for the token (see spec for details).
-	token := make([]byte, 6+16)
-	if _, err := rand.Read(token); err != nil {
+	idBytes := make([]byte, 6)
+	secretBytes := make([]byte, 16)
+	tokenLength := len(idBytes) + len(secretBytes)
+	charsetLength := len(validBootstrapTokenChars)
+	size := int(math.Ceil(float64(tokenLength) * math.Log2(float64(charsetLength)) / math.Log2(256)))
```
ericchiang: Add a comment on what this math is doing. It's not clear why this works.

pkg/asset/k8s.go
```diff
@@ -116,16 +118,38 @@ const validBootstrapTokenChars = "0123456789abcdefghijklmnopqrstuvwxyz"
...
+
+	random := make([]byte, size)
+	if _, err := rand.Read(random); err != nil {
 		return "", "", err
 	}
 
-	for i, b := range token {
-		token[i] = validBootstrapTokenChars[int(b)%len(validBootstrapTokenChars)]
+	encodeBytes(validBootstrapTokenChars, random, idBytes, secretBytes)
```
ericchiang: Don't have this be a function call? It just renames the variable and makes it harder to follow.

pkg/asset/k8s.go
```diff
@@ -116,16 +118,38 @@ const validBootstrapTokenChars = "0123456789abcdefghijklmnopqrstuvwxyz"
...
+// encoded slices represent the input value in little-endian unsigned
+// base len(charset).  The size of the encoded slices should be set
+// before calling this function; after consuming input, any "extra"
+// encoded slice length will be filled with charset[0].
+func encodeBytes(charset string, input []byte, encoded ...[]byte) {
+	charsetLength := len(charset)
+	inputInt := big.NewInt(0).SetBytes(input)
+	base := big.NewInt(int64(charsetLength))
+	remainder := big.NewInt(0)
+	for _, enc := range encoded {
```
ericchiang: Instead of this loop, just encode to a single byte slice and split after like the old code did?

pkg/asset/k8s.go
```diff
@@ -116,16 +118,38 @@ const validBootstrapTokenChars = "0123456789abcdefghijklmnopqrstuvwxyz"
...
 
-	for i, b := range token {
-		token[i] = validBootstrapTokenChars[int(b)%len(validBootstrapTokenChars)]
+	encodeBytes(validBootstrapTokenChars, random, idBytes, secretBytes)
+	return string(idBytes), string(secretBytes), nil
+}
+
+// encodeBytes encodes the input byte slice into encoded byte slices.
+// The input is treated as a big-endian unsigned integer, and the
+// encoded slices represent the input value in little-endian unsigned
```
ericchiang: Confused why this is mentioning "little-endian"

pkg/asset/k8s_test.go
```diff
@@ -0,0 +1,36 @@
+package asset
+
+import (
+	"testing"
+)
+
+func TestEncodeBytes(t *testing.T) {
```
ericchiang: I don't think this test buys us much. Maybe just a test on "newBootstrapToken" that ensures the resulting tokens are the right length and the right character set?

pkg/asset/k8s.go
wking: > Don't have this be a function call?

It's a function call so you can test base 256 -> base `len(charset)` conversion without pulling in `randRead`.  If you don't want the test, I can inline the conversion code here.

pkg/asset/k8s_test.go
wking: > I don't think this test buys us much. Maybe just a test on "newBootstrapToken"...

The old code would have passed that test.  This test checks the advertized base 256 to base `len(charset)` conversion, which you need to show that we're not unevenly collapsing the input distribution.

pkg/asset/k8s.go
wking: > Add a comment on what this math is doing.

You want `256^size > len(token)^len(charset)`.  Taking the log2 of both sides:

    log2(256^size) > log2(len(token)^len(charset))

bringing the exponents out of the logs:

    size * log2(256) > len(charset)*log2(len(token))

dividing by `log2(256)`:

    size > len(charset)*log2(len(token)) / log2(256)

We can't read a fractional byte, so `Ceil` the whole right hand side to get the smallest integer `size` with enough data, and that's where the right hand side comes from.

I've pushed a version of that explanation into the commit message with abc1878b, where it's available to folks who want to blame the line but not distracting to folks who either understand or don't care.  Let me know if you want me to add the explanation to comments in the file as well.

pkg/asset/k8s.go
wking: > Confused why this is mentioning "little-endian"

In the test, input bytes are big-endian, so `[255, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]` is 255 * 256^14 + 2 * 256^13 + ... + 12^256^3 +  13 * 256^2 + 14 * 256 + 15 = 1324076502674089023697701463353265679.

The output encoded value is little endian, so `n2r8j0ug4twvqmqt53hyff` is [23, 2, 27, 8, 19, 0, 30, 16, 4, 29, 32, 31, 26, 22, 26, 29, 5, 3, 17, 34, 15, 15] (decoding each character to decimal, e.g. `n` -> 23, `2` -> 2, ...) which is 23 + 2 * 36 + 27 * 36^2 + ... + 15 * 36^20 + 15 * 36^21 = 7431760576136592951846490044435983.  If you add on the next two unused values (which a 24-character token would have consumed), you get 7431760576136592951846490044435983 + 4 * 36^22 + 2 * 36^23 = 1324076502674089023697701463353265679, the input number.

So the input is being interpreted as a big-endian unsigned integer, and the encoded output is little-endian base `len(charset)`.  If we had big-endian output, it would have been `ffyh35tqmqvwt4gu0j8r2n` instead of `n2r8j0ug4twvqmqt53hyff`.

wking:
> Just to be clear the feedback from the reviewers is "this is hard to read" which is why we recommended kubernetes/kubernetes#57984 as an alternative. Does that make sense?

That makes sense.  And I'm obviously not a maintainer; you're free to move in whichever direction makes the most sense to you.  I've filed a Go proposal (golang/go#25531) with a start at making this sort of thing easier with the stdlib, and that's probably where I'll take the "upstream a generic solution" approach (instead of upstreaming into kubeadm, since as you point out that's more of a sibling than an upstream dependency).

I've pushed aad377e5 -> abc1878b here, adding a derivation of the `size` calculation to the commit message (hopefully addressing [this point][1]) and adjusting the helper function to only take a single encoded slice (addressing [this point][2]).  I've tried to address the other points in their sub-threads, but let me know if I missed anything.  And even if you like where this is going, I have no problem with you deciding to not merge it for whatever reasons you like.

[1]: https://github.com/kubernetes-incubator/bootkube/pull/984#discussion_r190430913
[2]: https://github.com/kubernetes-incubator/bootkube/pull/984#discussion_r190431402

rphillips:
@wking since the length of the alphabet is a power of 2 (a-z0-9, 26+10=36, sqrt(36)=6), there would not be a bias using the mod, correct?

aaronlevy:
Any updates on this? I think we're in agreement that what we have right now is not correct - and we should either use common approach with upstream/kubeadm, or proposal here?

wking:
> I think we're in agreement that what we have right now is not correct - and we should either use common approach with upstream/kubeadm, or proposal here?

My upstream Go proposals were declined ([here][1] and [here][2]) because they could be implemented outside the stdlib.  I'm in favor of creating a tiny package that implements this random-token-drawn-from-user-provided-charset functionality, possibly based on the `Choose` API I floated in golang/go#25534.  Then bootkube, kubeadm, and probably a number of other consumers could depend on that (hopefully boring and stable ;) package.  If that sounds like a useful way forward, I can get an initial proof of concept going in a personal repository, link to that from this discussion, and then we can talk about whether we should move the repo into an organization or not.  Does that sound reasonable?

[1]: https://github.com/golang/go/issues/25531#issuecomment-394481505
[2]: https://github.com/golang/go/issues/25534#issuecomment-394480901

aaronlevy:
In the interest of simplicity I'd probably lean towards implementing a fix here. If the kubeadm approach is correct-but-different, not sure trying to get a shared approach across all projects is worth the effort (but just my 2cents). If for example the kubeadm approach was acceptable, and it was a relatively short utility function, I'd probably just advocate to add it inline rather than vendor kubeadm...

PR #989, opened 2018-06-15
Use hostNetwork: true for controller and scheduler pods

### LINKED ISSUE #64: some k8s components report unhealthy status after cluster bootsrap
`curl  127.0.0.1:8080/api/v1/componentstatuses`

``` json
{
  "kind": "ComponentStatusList",
  "apiVersion": "v1",
  "metadata": {
    "selfLink": "/api/v1/componentstatuses"
  },
  "items": [
    {
      "metadata": {
        "name": "scheduler",
        "selfLink": "/api/v1/componentstatuses/scheduler",
        "creationTimestamp": null
      },
      "conditions": [
        {
          "type": "Healthy",
          "status": "False",
          "message": "Get http://127.0.0.1:10251/healthz: dial tcp 127.0.0.1:10251: getsockopt: connection refused"
        }
      ]
    },
    {
      "metadata": {
        "name": "controller-manager",
        "selfLink": "/api/v1/componentstatuses/controller-manager",
        "creationTimestamp": null
      },
      "conditions": [
        {
          "type": "Healthy",
          "status": "False",
          "message": "Get http://127.0.0.1:10252/healthz: dial tcp 127.0.0.1:10252: getsockopt: connection refused"
        }
      ]
    },
    {
      "metadata": {
        "name": "etcd-0",
        "selfLink": "/api/v1/componentstatuses/etcd-0",
        "creationTimestamp": null
      },
      "conditions": [
        {
          "type": "Healthy",
          "status": "True",
          "message": "{\"health\": \"true\"}"
        }
      ]
    }
  ]
```

JrCs:
This PR is used to mimic how kubeadm deploy the control plane (in this case only controller manager and scheduller), and allow to fix some issues where components report unhealthy status.
* Fix bad report of `kubectl get componentstatus` command on masters servers
* Fix #64

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -416,6 +416,7 @@ spec:
         command:
         - ./hyperkube
         - controller-manager
+        - --address=127.0.0.1
         - --use-service-account-credentials
         - --allocate-node-cidrs=true
         - --cloud-provider={{ .CloudProvider }}
@@ -427,8 +428,12 @@ spec:
         - --leader-elect=true
         - --root-ca-file=/etc/kubernetes/secrets/ca.crt
         - --service-account-private-key-file=/etc/kubernetes/secrets/service-account.key
+        ports:
+        - name: healthz
+          containerPort: 10252
         livenessProbe:
           httpGet:
+            host: 127.0.0.1
             path: /healthz
             port: 10252  # Note: Using default port. Update if --port option is set differently.
           initialDelaySeconds: 15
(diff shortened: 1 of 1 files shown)


aaronlevy:
Except for the componentstatus assumptions - I don't believe that there is a need to use host-networking for these components - and IMO it is preferable not to. Assuming that etcd/apiserver/cm/scheduler all all co-located and addressable via localhost doesn't map very well to common/acceptable deployment models.

Personally, I don't think componentstatus is that useful (compared to monitoring setup against all control-plane components) - so optimizing around that (and restricting self-hosted deployment models) feels like the wrong direction.

I'm going to close this, as it's very likely not a direction I think we'd want to go by default (but a user could always change on their own). But if there are other reasons for this that I'm missing, happy to re-open to discuss.

PR #996, opened 2018-07-28
AWS Quickstart: mention TAG_MASTER

flowchartsman:
Ref #311 There should be a mention of adding additional masters. I understand this is a difficult thing to place correctly, since it's important to note that boot-kube is only necessary for the first node in the cluster.  This is why I mentioned it last.  It could also be a comment line in the **Add Workers** section if you think it's too confusing down here.

### DIFF
--- a/hack/quickstart/init-node.sh
+++ b/hack/quickstart/init-node.sh
@@ -85,6 +85,9 @@ if [ "${REMOTE_HOST}" != "local" ]; then
 
     echo
     echo "Node (${REMOTE_HOST}) bootstrap complete. It may take a few minutes for the node to become ready."
+    if [ "$TAG_MASTER" = true ] ; then
+        echo "NOTE: You have configured this node as a master. You may want to set up load balancing with your cloud provider so that masters receive equal traffic and gracefully fail over."
+    fi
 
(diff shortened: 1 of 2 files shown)


flowchartsman:
Added the note above to try and find a good middle ground resolution to #684. Rather than being proscriptive, this commit will make bootkube simply issue a note when extra masters are initialized to remind the user that they may wish to load balance the master.

aaronlevy:
I feel like this might be a bit misleading to the user as it realistically entails quite a few more steps - adding a loadbalancer also means updating all of the kubelet configs to point to the LB, as well as making sure the TLS assets are generated with dns/LB location. You would still have a single etcd node (just multiple masters running control-plane components).

We could consider changing the instructions to use an ELB and put the single master behind it - but comes with increased cost for what was supposed to be a hack/test environment. Or maybe an HA quickstart that just defaults to 3 nodes, with etcd, and LB. But overall, for real production HA clusters, users should look at more featured tools like https://github.com/poseidon/typhoon (which makes use of bootkube).

The hack directory isn't really meant as a real-world environment -- but I do see that a demonstration of HA control-plane could be valuable.

Issue #779, opened 2017-11-19
Integrate support for Cilium CNI


jpds:
I'd like to use cilium.io as my CNI with a Typhoon deployment at some point, however it looks like bootkube requires support for this first before support can be added there.

dghubble:
No specific changes need to be made to `bootkube start` for the self-hosted control plane to use a particular CNI provider. You can independently swap the flannel manifests for your Cilium manifests and try to get it to work correctly.

If you're interested in adding Cilium, a good starting point would be adding the DaemonSet and sidecar (to lay down the CNI plugins on hosts) manifests to `bootkube render` and verifying you can bootstrap clusters and container networking works as expected. Separately, we'd want to decide whether its worth supporting and testing as this takes time.

jpds:
PR now at: https://github.com/poseidon/terraform-render-bootkube/pull/34

PR #999, opened 2018-08-23
Cilium CNI

Promaethius:
#779 

Add upstream Cilium CNI option.

Base templates adapted from official Cilium documentation: https://raw.githubusercontent.com/cilium/cilium/v1.1/examples/kubernetes/1.11/cilium.yaml

Cilium initContainer installs latest releast of containernetworking/cni releases required by k8s.

### DIFF
--- a/pkg/asset/asset.go
+++ b/pkg/asset/asset.go
@@ -61,6 +61,12 @@ const (
 	AssetPathCalicoGlobalNetworkSetsCRD     = "manifests/calico-global-network-sets-crd.yaml"
 	AssetPathCalicoIPPoolsCRD               = "manifests/calico-ip-pools-crd.yaml"
 	AssetPathCalicoClusterInformationsCRD   = "manifests/calico-cluster-informations-crd.yaml"
+	AssetPathCiliumCfg                      = "manifests/cilium-cfg.yaml"
+	AssetPathCilium                         = "manifests/cilium.yaml"
+	AssetPathCiliumClusterRoleBinding       = "manifests/cilium-cluster-role-binding.yaml"
+	AssetPathCiliumClusterRole              = "manifests/cilium-cluster-role.yaml"
+	AssetPathCiliumSA                       = "manifests/cilium-sa.yaml"
+	AssetPathCiliumSecret                   = "manifests/cilium-secret.yaml"
 	AssetPathAPIServerSecret                = "manifests/kube-apiserver-secret.yaml"
 	AssetPathAPIServer                      = "manifests/kube-apiserver.yaml"
 	AssetPathControllerManager              = "manifests/kube-controller-manager.yaml"
@@ -127,6 +133,7 @@ type ImageVersions struct {
 	FlannelCNI      string
 	Calico          string
 	CalicoCNI       string
+	Cilium          string
 	CoreDNS         string
 	Hyperkube       string
 	Kenc            string
@@ -189,6 +196,15 @@ func NewDefaultAssets(conf Config) (Assets, error) {
 		return Assets{}, err
 	}
 	as = append(as, cmSecret)
+	
+  // Cilium ETCD secret
+	if conf.NetworkProvider == NetworkCilium && conf.EtcdUseTLS {
+		cSecret, err := newCiliumSecretAsset(as)
+		if err != nil {
+			return Assets{}, err
+		}
+		as = append(as, cSecret)
+	}
 
 	return as, nil
 }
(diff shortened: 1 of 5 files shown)


Promaethius:
CLA signed and I corrected my primary email.

Promaethius:
Planning on bumping the Cilium configuration to v1.2.0 per https://raw.githubusercontent.com/cilium/cilium/1.2.0/examples/kubernetes/1.11/cilium.yaml

pkg/asset/internal/templates.go
```diff
@@ -1565,4 +1565,349 @@ subjects:
...
+metadata:
+  name: cilium-config
+  namespace: kube-system
+data:
+  etcd-config: |-
+    ---
+    endpoints:
+    - {{ range $i, $e := .EtcdServers }}{{ if $i }},{{end}}{{ $e }}{{end}}
+{{- if .EtcdUseTLS }}
+    ca-file: '/var/lib/etcd-secrets/etcd-client-ca.crt'
```
dghubble: Not thrilled Cilium still demands etcd client credentials. Flannel and Calico are not given direct access to the cluster's underlying etcd cluster. I'd say block until Cilium addresses it.

pkg/asset/asset.go
```diff
@@ -189,6 +196,15 @@ func NewDefaultAssets(conf Config) (Assets, error) {
 		return Assets{}, err
 	}
 	as = append(as, cmSecret)
+	
+  // Cilium ETCD secret
+	if conf.NetworkProvider == NetworkCilium && conf.EtcdUseTLS {
+		cSecret, err := newCiliumSecretAsset(as)
```
Promaethius: newCiliumSecretAsset is a missing function

pkg/asset/asset.go
```diff
@@ -189,6 +196,15 @@ func NewDefaultAssets(conf Config) (Assets, error) {
 		return Assets{}, err
 	}
 	as = append(as, cmSecret)
+	
+  // Cilium ETCD secret
```
Promaethius: Based on https://github.com/kubernetes-incubator/bootkube/pull/999/files#r212474645,

https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/tls.go
newEtcdKeyAndCert( CAEtcd, CAKeyEtcd, "cilium-agent", etcdServers)

pkg/asset/internal/templates.go
Promaethius: @dghubble thoughts on using RBAC for this particular etcd connection?

dghubble: Hey Promaethius, I'm excited for Cilium as well, but scoping down the etcd client cert doesn't alleviate the concern. Its the principle. The Kubernetes cluster's etcd is for Kubernetes alone. No other cluster components are given access to it. Not flannel, not Calico, and not Cilium. Cilium can't be special.

We can't lower the bar just for Cilium and as a distro maintainer, the requirement to mint a special etcd client cert just for Cilium seems unreasonable. Cilium could make use of CRDs or a few choice host locations like flannel and Calico have, in order to align with established expectations.

Not sure of the governance of this repo or how I'm the sole reviewer, but I can't approve this.

Promaethius: @dghubble I totally understand, thank you for your feedback.
At this point, is it out of the question to introduce new CLI options for a standalone, external ETCD specifically for Cillium? As far as I understand you, the fact that my PR used kubernetes ETCD cluster blocks this.

aaronlevy: Agree with @dghubble assessment. Direct access to etcd is a slippery slope. I've seen this convention start to erode a bit with aggregated api servers, but personally would not want to advocate for it here.

neoakris:
/reopen

Just to point out newer versions of cilium have addressed the issue of trying to reuse Kubernete's etcd.
They have a CRD option and an etcd operator option. 

Personally I'd love to see cilium become the future standard CNI:
Since it has 
*cluster wide network policies
*Layer 7 Network Policies that can block and allow based on DNS names (all others are L4 last I checked)

In theory you could make a network policy that only allowed ingress from a private docker repo which had approved containers, so if a pod became compromised they couldn't just download hacker tools from random public docker repos.

andrewrynhard:
@neoakris We are working on refactoring the builtin rendering process. I'm not quite sure what the fallout will be just yet, but we are looking at having a way to build a plugin that can/will handle the rendering process. We aren't quite sure if they will live in tree. There will most likely be a default plugin where we can we can revisit which CNI to default to once we have things fleshed out some more.

PR #1003, opened 2018-10-11
Update Kubernetes from v1.11.3 to v1.12.1

dghubble:
* Workaround https://github.com/kubernetes/kubernetes/issues/68973


aaronlevy:
These changes overall seem fine to me. On a quick scan of the issue, however, I'm not clear on how these certs are used. Is this something we should open an issue for and come back and revisit minting real certs for?

Then also - we need the tests to be running before actually merging this. @rphillips was looking into this.

aaronlevy:
ok to test

dghubble:
Yeah sorry, it was mainly intended to repro the issue. The
controller-manager generates certs and I'm thinking we may want to mint
certs

dghubble:
I'll rebase to pickup the new pod-checkpointer image once #1010 lands.

dghubble:
Concerning the controller-manager generated certs, I poked around and asked some questions. https://github.com/kubernetes/kubernetes/issues/68973#issuecomment-429638833

I kinda favor not bothering to mint real certs for these yet. The HTTPS endpoint seems to be a new and so far unused endpoint that doesn't have clients. The HTTP endpoint is still always there. Nothing even happens if the HTTPS endpoint is turned off (--secure-port=0). Seems half baked / incomplete.

If we were aiming to tell users they can now talk to the controller-manager with TLS, I think we'd need to create a controller-manager service, use a conventional DNS name, mint certs with that name, and also tell people to just ignore the HTTP endpoint for now. Not sure what the use case is here really, feels kinda fruitless. :man_shrugging:

dghubble:
I think this is ready to go. Might be valuable to get on v1.12 before the new cycle of manifest improvements.

rphillips:
Looks like we missed these.

https://github.com/kubernetes-incubator/bootkube/blob/40956680d90135cfd8edbe59e5477b96e956fba6/Makefile#L80-L81

dghubble:
The checkpointer image doesn't necessarily need to update those source files to work with v1.12.

PR #1012, opened 2018-10-19
start: fix wait loop for cluster scoped CRDs

sttts:
Cluster scoped CRDs have a different URI.


sttts:
/assign @rphillips @aaronlevy 

This is critical for our CVO crd.

sttts:
The issue is actually a different one. The whole CRD check is broken. Will open another PR.

PR #1022, opened 2018-11-02
Update README.md

xichengliudui:
Update README.md

### DIFF
--- a/pkg/checkpoint/cri/README.md
+++ b/pkg/checkpoint/cri/README.md
@@ -6,5 +6,8 @@ We copy generated CRI APIs for a couple reasons:
 
 The various versions of CRI are taken from:
 
+* v1alpha1: https://github.com/kubernetes/kubernetes/tree/release-1.8/pkg/kubelet/apis/cri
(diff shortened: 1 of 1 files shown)


dghubble:
I'm confused. The existing links look correct to me. The code copied into [pkg/checkpointer/cri/v1alpha2](https://github.com/kubernetes-incubator/bootkube/tree/v0.14.0/pkg/checkpoint/cri/v1alpha2) does match what was in [kubernetes](https://github.com/kubernetes/kubernetes/tree/v1.10.0-rc.1/pkg/kubelet/apis/cri/runtime/v1alpha2) v1.10.0-rc.1.

xichengliudui:
@dghubble  Thanks for the tip, but shouldn't we show the latest version of cri

dghubble:
The message is supposed to indicate where the code was taken from, for internal implementers to not lose track of. If there's a need to bump to the code from v1.11 or v1.12, it could be bumped and *then* this change would be accurate.

xichengliudui:
@dghubble  Do I need to turn it close?

dghubble:
Yeah, I think this can be closed

PR #1023, opened 2018-11-02
update coredns

SataQiu:
update coredns image version

### DIFF
--- a/pkg/asset/images.go
+++ b/pkg/asset/images.go
@@ -7,7 +7,7 @@ var DefaultImages = ImageVersions{
 	FlannelCNI:      "quay.io/coreos/flannel-cni:v0.3.0",
 	Calico:          "quay.io/calico/node:v3.0.3",
 	CalicoCNI:       "quay.io/calico/cni:v2.0.0",
-	CoreDNS:         "k8s.gcr.io/coredns:1.1.3",
+	CoreDNS:         "k8s.gcr.io/coredns:1.2.4",
 	Hyperkube:       "k8s.gcr.io/hyperkube:v1.12.1",
 	PodCheckpointer: "quay.io/coreos/pod-checkpointer:018007e77ccd61e8e59b7e15d7fc5e318a5a2682",
 }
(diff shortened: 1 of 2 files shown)


dghubble:
You should be able to use 1.2.4

dghubble:
I forgot to mention, the newer CoreDNS version both fixes the earlier memory leak and also requires a change in the ClusterRole. See https://github.com/poseidon/terraform-render-bootkube/commit/6a77775e52179f2ff0c71f0ad3ceea3a51822f07

dghubble:
Leaving final approval for @rphillips

SataQiu:
Dear @rphillips , all the tests were passed!

aaronlevy:
Looks like the config changed a bit as well:

Upstream added `loop` and `loadbalance`:
https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed#L63-L68

Ours (for reference):
https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L812-L815

I'm unfamiliar with CoreDNS, so I'm not sure the implication - but want to make sure we don't need these.

dghubble:
From the CoreDNS plugins list, `loop` and `loadbalance` do look like useful new plugins. I haven't tried those yet. @SataQiu want to try them out?

dghubble:
Fix s/1.2.5/1.2.4/. CoreDNS is an indpendent project and not every release is distributed via Kubernetes images. Be careful to watch upstreams and test.

dghubble:
The new CoreDNS `loop` and `loadbalance` plugins are just "nice-to-have", they're not required. CoreDNS v1.2.4 works fine without them. 

* loop sends an initial query to detect infinite forwarding loops in configured upstream DNS servers and fast exit with an error (its a fatal misconfiguration on the network that will otherwise cause resolvers to consume memory/CPU until crashing, masking the problem)
* loadbalance randomizes the ordering of A, AAAA, and MX records in responses to provide round-robin load balancing (as usual, clients may still cache responses though)

I've started using them without issue [here](https://github.com/poseidon/terraform-render-bootkube/commit/3f3ab6b5c07f606e2c0cc6a096887962172a2680).

I'm not sure why tests fail here (I don't have access either). You could try re-running without one or both plugins or spinning it up locally to get a better idea what's going on. There must be something different about the setup here.

PR #1025, opened 2018-11-05
Disable Kubelet read-only port 10255

dghubble:
`pod-checkpointer` creates an `insecureClient` and `secureClient` to the kubelet, but only makes use of the former (via 10255). Adapt `localParentPods` to try to use the secure client first, then the insecure client to support clusters that disable the kubelet read-only port.

Open questions:

* What about we just always use the kubelet secure API (instead of having a fallback)?
* Testing with the secure client, we'll need a ClusterRole with node/proxy get. Surprising since the pod-checkpointer also mounts the admin kubeconfig from the host (possibly not being used). 

## Background

Today in bootkube, the kubelet read-only port (10255) is enabled and pod-checkpointer [uses](https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/checkpoint/kubelet.go#L52) its `/pods` endpoint to get all pods (later filters to find parent pods). That all works fine.

We've come a long way toward eliminating the read-only port. Cloud load balancers can now health check the apiserver, Prometheus can [use the kubelet secure](https://github.com/poseidon/typhoon/pull/217) API to scrape metrics, and [heapster can get metrics](https://github.com/poseidon/typhoon/pull/323) from the kubelet secure API too.

Running clusters with kubelet `read-only-port=0` (disabled), the active pod-checkpointer will log:

```
failed to list local parent pods, assuming none are running: Get http://127.0.0.1:10255/pods/: dial tcp 127.0.0.1:10255: connect: connection refused
```

Interestingly, even with localParentPods calls failing on these clusters, recovery from cluster power cycling is unaffected. But I figure if we can eliminate this one last use of the read-only API, all the better.

### DIFF
--- a/Documentation/network-requirements.md
+++ b/Documentation/network-requirements.md
@@ -24,5 +24,5 @@ The information below describes a minimum set of port allocations used by Kubern
 | TCP      | 4194        | Master & Worker Nodes          | The port of the localhost cAdvisor endpoint |
 | UDP      | 4789        | Master & Worker Nodes          | flannel overlay network - *vxlan backend* |
 | TCP      | 10250       | Master Nodes                   | Worker node Kubelet API for exec and logs.                                  |
-| TCP      | 10255       | Master & Worker Nodes          | Worker node read-only Kubelet API (Heapster).                                  |
+| TCP      | 10255       | Master & Worker Nodes          | (Optional) Worker node read-only Kubelet API (Heapster).               |
 | TCP      | 30000-32767 | External Application Consumers | Default port range for [external service][https://kubernetes.io/docs/concepts/services-networking/service] ports. Typically, these ports would need to be exposed to external load-balancers, or other external consumers of the application itself. |
--- a/hack/multi-node/user-data.sample
+++ b/hack/multi-node/user-data.sample
@@ -43,6 +43,7 @@ coreos:
           --network-plugin=cni \
           --node-labels=node-role.kubernetes.io/master \
           --pod-manifest-path=/etc/kubernetes/manifests \
+          --read-only-port=0 \
           --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
           --rotate-certificates
         ExecStop=-/usr/bin/rkt stop --uuid-file=/var/cache/kubelet-pod.uuid
--- a/hack/quickstart/kubelet.service
+++ b/hack/quickstart/kubelet.service
@@ -40,6 +40,7 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --network-plugin=cni \
   --node-labels=${node_label} \
   --pod-manifest-path=/etc/kubernetes/manifests \
+  --read-only-port=0 \
   --register-with-taints=${node_taint} \
   --rotate-certificates
 ExecStop=-/usr/bin/rkt stop --uuid-file=/var/cache/kubelet-pod.uuid
--- a/hack/single-node/user-data.sample
+++ b/hack/single-node/user-data.sample
@@ -43,6 +43,7 @@ coreos:
           --network-plugin=cni \
           --node-labels=node-role.kubernetes.io/master \
           --pod-manifest-path=/etc/kubernetes/manifests \
+          --read-only-port=0 \
           --rotate-certificates
         ExecStop=-/usr/bin/rkt stop --uuid-file=/var/cache/kubelet-pod.uuid
         Restart=always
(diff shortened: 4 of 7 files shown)


dghubble:
I expect I'll need to split the checkpointer change and user-data change for tests to pass.

But manually testing on clusters with `--read-only-port=0`, I was able to resolve the errors I posted above, using this PR as `quay.io/dghubble/pod-checkpointer:14b2af0e8acc40cabdbd2bad5218ae94730c9ee1` and adding a ClusterRole:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: pod-checkpointer
rules:
  - apiGroups: [""]
    resources:
      - nodes
    verbs:
      - get
```

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: pod-checkpointer
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: pod-checkpointer
subjects:
- kind: ServiceAccount
  name: pod-checkpointer
  namespace: kube-system
```

rphillips:
Thanks Dalton for looking into this. We just had a conversation about this exact issue. I am worried about the certificate rotation needed for the checkpointer -> kubelet secure connection. There is an edge case where if a node is down for a period of time, comes back, the cert might be expired. Not to mention needing to maintain the cert rotation on that connection.

We are definitely on board to try and remove the insecure port requirement.

dghubble:
Ah right. bootkube supports using kubelet TLS bootstrap so the concern is if there's no apiserver (e.g. the cluster is powered off) when the certificate expires, then on startup, kubelet starts pod-checkpointer which may not make it far enough* to start the bootstrap-apiserver, to issue the new cert so the kubelet can register. So given the choice, we prefer to at least fallback to 10255.

I can post a standalone PR for the checkpointer to try 10250, then fallback to 10255.

*I'm not sure how that can happen, since clusters that disable read-only today, pod-checkpointer can't contact anything. Yet control plane recovery succeeds.

dghubble:
Added the ClusterRole and ClusterRoleBinding to give checkpointer permission to perform requests previously done via the kubelet read-only API. Requires #1027 and a checkpointer release before this can pass tests.

rphillips:
@dghubble can you update the checkpointer hash to: `83e25e5968391b9eb342042c435d1b3eeddb2be1`

dghubble:
Updated, and opened https://github.com/kubernetes-incubator/bootkube/pull/1031 with just the image change, since we might consider it separate from the feature change of disabling read-only in test / example clusters.

Release Note:

```
* Disable Kubelet read-only port 10255 (may affect applications)
  * Configure any added apps using the insecure API to use the Kubelet secure API (e.g. heapster, prometheus)
  * If using Kubelet certificate rotation, you may wish to leave read-only port on to mitigate any [chance of interruption](https://github.com/kubernetes-incubator/bootkube/pull/1025#issuecomment-437038376).
```

rphillips:
```
19:17:03 Error: loading manifests: parse file /home/core/assets/manifests/pod-checkpointer-cluster-role.yaml: invalid manifest: yaml: line 7: found unexpected end of stream
```

rphillips:
coreosbot run e2e calico

Documentation/network-requirements.md
```diff
@@ -24,5 +24,4 @@ The information below describes a minimum set of port allocations used by Kubern
 | TCP      | 4194        | Master & Worker Nodes          | The port of the localhost cAdvisor endpoint |
 | UDP      | 4789        | Master & Worker Nodes          | flannel overlay network - *vxlan backend* |
 | TCP      | 10250       | Master Nodes                   | Worker node Kubelet API for exec and logs.                                  |
-| TCP      | 10255       | Master & Worker Nodes          | Worker node read-only Kubelet API (Heapster).                                  |
```
rphillips: We should leave this network requirement in but label it 'optional'. otherwise this lgtm and tests are passing.

dghubble: Updated

dghubble:
Did we want to add just the checkpointer bump via #1031, then rebase this to follow it? Since the two are somewhat independent. :man_shrugging: either way works for me

rphillips:
@dghubble Thanks! We can rebase this PR now.

dghubble:
Updated

rphillips:
coreosbot run e2e calico

PR #1034, opened 2018-12-18
Minor docs fix to re-run master tests

dghubble:
Mainly intended to validate that master is in a decent state since there are some failing PRs.

PR #1030, opened 2018-11-20
start: wait until pods ready

sttts:
Minimal alternative to https://github.com/kubernetes-incubator/bootkube/pull/1029.

Processes that are accessed via a service are not serving requests until they are "ready". In the case of the kube-apiserver being ready is reported through the `/healthz` endpoint.

Without this PR, bootkube switches to a self-hosted kube-apiserver as soon as the pod is marked running. It tears down the bootstrap control plane immediately, leaving the cluster without an API server, possibly for a number of seconds.

The ready condition in the Pod is meant exactly for this purpose, i.e. to report when a pod is ready to serve. We should use it.

PR #1035, opened 2018-12-18
start: remove defaulted required pods

rphillips:
The bootstrap node's api server does not enter a ready state.

Note: waiting on jenkins to test, but this is reproducible with the vagrant environment.

/cc @sttts @dghubble @aaronlevy


rphillips:
coreosbot run e2e

aaronlevy:
What I've seen is that it just takes a very long time for the status to settle with the checkpointed pivot to be ready (but eventually it will). I have a feeling that https://github.com/kubernetes-incubator/bootkube/pull/1030 might be related to this.

I think https://github.com/kubernetes-incubator/bootkube/pull/1030 is technically correct - so if that is the culprit - may need to think through a more deterministic way of determining completeness (that is safe for current checkpoint behavior). But I don't think just removing the checks for all pods is likely the correct way of going about this.

dghubble:
Well it seems like master is in a pretty flaky state even for NoOp changes (https://github.com/kubernetes-incubator/bootkube/pull/1034) and blocking getting to v1.12.3. Do we want to revert recent changes or relax this requirement in the interests of getting things moving again?

I'm fine to revert #1034 too, if it gets changes moving again.

PR #1048, opened 2019-10-09
updates cri v1alpha2

### LINKED ISSUE #1047: update CRI in checkpointer
The CRI in checkpointer is outdated and requires an update 

https://github.com/kubernetes-incubator/bootkube/tree/master/pkg/checkpoint/cri

rmenn:
Fixes  #1047


rmenn:
Ouch looks like protobuf/proto needs an update as well, i'd prefer to fix and merge this once the go modules PR has been merged

rmenn:
Shows checkpointer as functional and running in cluster
```
kube-system   pod-checkpointer-jld6v                     1/1     Running   0          2m45s
kube-system   pod-checkpointer-jld6v-172.17.4.101        1/1     Running   0          35s
```
Image using the updated version of checkpointer
```
multi-node|cri_update ⇒ kubectl --kubeconfig=cluster/auth/kubeconfig get ds pod-checkpointer -o=jsonpath='{$.spec.template.spec.containers[:1].image}' -n kube-system
quay.io/rmenn/checkpoint:4af128538feda93c386e45ddb6f73399e0ce9b6a
```

pkg/checkpoint/cri/README.md
```diff
@@ -7,4 +7,4 @@ We copy generated CRI APIs for a couple reasons:
 The various versions of CRI are taken from:
 
 * v1alpha1: https://github.com/kubernetes/kubernetes/tree/v1.9.6/pkg/kubelet/apis/cri
-* v1alpha2: https://github.com/kubernetes/kubernetes/tree/v1.10.0-rc.1/pkg/kubelet/apis/cri
+* v1alpha2: https://github.com/kubernetes/cri-api/tree/release-1.16/pkg/apis
```
andrewrynhard: IMHO its safer to be explicit about the version since, in theory, we could build two different binaries (e.g. one with 1.16.1, and another with 1.16.2) on two different days without ever changes the code. I propose we use `https://github.com/kubernetes/cri-api/tree/kubernetes-1.16.2`

rmenn: Ill change the readme, the branches are the same, presently, does that work ?

andrewrynhard: For sure, if `release-1.16` was `kubernetes-1.16.2` when you pulled these in then 👍. I think in the long run we will import go code instead of copying the protobuf definitions, so this should be fine for now.

PR #1049, opened 2019-10-15
build,image: support building multi-arch container images

dongsupark:
Allow users to build container images for multiple architectures, like `arm64`.

Also add Makefile targets to customize build parameters.

For example:

```
make release-bootkube IMAGE_REPO_BOOTKUBE=quay.io/kinvolk/bootkube GOARCH=arm64 VERSION=v0.14.0
make release-podcheckpointer IMAGE_REPO_PODCHECKPOINTER=quay.io/kinvolk/pod-checkpointer GOARCH=arm64 VERSION=83e25e5968391b9eb342042c435d1b3eeddb2be1
```

To do that, we need to also fix hidden build issues in `make cross` target, so that the default release target can cross-build binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,10 @@ export CGO_ENABLED:=0
 export GOARCH:=amd64
 export PATH:=$(PATH):$(PWD)
 
+IMAGE_REPO_BOOTKUBE?=quay.io/coreos/bootkube
+IMAGE_REPO_PODCHECKPOINTER?=quay.io/coreos/pod-checkpointer
+VERSION?=
+
 LOCAL_OS:=$(shell uname | tr A-Z a-z)
 GOFILES:=$(shell find . -name '*.go' ! -path './vendor/*')
 VENDOR_GOFILES ?= $(shell find vendor -name '*.go')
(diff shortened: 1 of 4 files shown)


andrewrynhard:
@dongsupark This project was recently moved over to this org, and we are in the process of transitioning project owners. Once things settle, we can take a closer look at this.

andrewrynhard:
@rmenn Thoughts?

andrewrynhard:
@dongsupark Are you still wanting to push this through?

invidian:
Yes, we are still interested in that :)

Makefile
```diff
@@ -2,6 +2,10 @@ export CGO_ENABLED:=0
 export GOARCH:=amd64
 export PATH:=$(PATH):$(PWD)
 
+IMAGE_REPO_BOOTKUBE?=quay.io/coreos/bootkube
+IMAGE_REPO_PODCHECKPOINTER?=quay.io/coreos/pod-checkpointer
```
andrewrynhard, line 6: These may need to change. @rmenn What is the status of the repos we have?

andrewrynhard (commented):
Seems like a good start. We have no way of running tests on this currently, but I know from experience that this will fail because we need to update the default flannel templates for arm. The official deployment is a good reference on how we might do that: https://github.com/coreos/flannel/blob/master/Documentation/kube-flannel.yml

dghubble (commented):
Not a review. I'm not involved in bootkube these days and there doesn't seem to be an option to decline a review request.

dongsupark:
@dghubble 
Sure, no problem.
I set you and Aaron as reviewers back in 2019, just because back then you folks were still maintainers. :-)
At the moment, I am waiting for reviews just from the current maintainers.

mrbobbytables:
This project is being archived, for more information see this issue:
https://github.com/kubernetes/org/issues/2824

/close

PR #1052, opened 2019-10-23
Create OWNERS_ALIASES

aaronlevy:


aaronlevy:
Adding per https://github.com/kubernetes/org/issues/1305#issuecomment-543844378

PR #1056, opened 2019-11-11
Update kubernetes to 1.16

rmenn:
The following changes are present

* ~Updates flannel to 0.11~
* ~Updates CoreDNS to 1.6.2 along with the config map~
* ~Replace of `node-role.kubernetes.io/master` with `node.kubernetes.io/master` as label/selector/taint/toleration ( https://github.com/kubernetes/kubernetes/issues/75457 )~
* Renames the hyperkube short aliases since they were removed ( https://github.com/kubernetes/kubernetes/pull/76953 ) 
* Fixes hack/* with updates for 1.16
* Updates client-go with associated changes in bootkube and checkpointer
* Work around for kubelet self labeling/tainting in `hack/*`

Fixes : #1044


pkg/asset/images.go
```diff
@@ -3,11 +3,11 @@ package asset
 // DefaultImages are the defualt images bootkube components use.
 var DefaultImages = ImageVersions{
 	Etcd:            "quay.io/coreos/etcd:v3.3.12",
-	Flannel:         "quay.io/coreos/flannel:v0.10.0-amd64",
+	Flannel:         "quay.io/coreos/flannel:v0.11.0-amd64",
 	FlannelCNI:      "quay.io/coreos/flannel-cni:v0.3.0",
 	Calico:          "quay.io/calico/node:v3.0.3",
```
rata: Would it make sense to update calico too? We are using bootkube and calico, we just upgraded to calico 3.10.1 (https://docs.projectcalico.org/v3.10/release-notes/#v3.10.1)

We are templating this vars outside of bootkube (to keep bootkube on the bootkube functionality), but maybe it makes sense to update so it is clear that is supported and there is no reason to use an old calico version.

What do you think?

pkg/asset/internal/templates.go
```diff
@@ -806,13 +806,15 @@ data:
     .:53 {
         errors
         health
-        kubernetes cluster.local {{ .ServiceCIDR }} {
+        log . {
+            class error
+        }
+        kubernetes cluster.local in-addr.arpa ip6.arpa {
```
rata: Did not follow this. Can you please elaborate a little bit? Not sure why serviceCIDR was needed and is not needed anymore

hack/multi-node/user-data.sample
```diff
@@ -29,7 +29,6 @@ coreos:
         ExecStartPre=/bin/mkdir -p /var/lib/kubelet/pki
         ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/cache/kubelet-pod.uuid
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
-          --allow-privileged \
```
rata: Maybe it is worth explaining why this is not needed anymore? In the commit message, for example?

rmenn: Sorry, its a deprecated flag which is no longer available

pkg/asset/images.go
rmenn: well, we are still discussing how to enable multiple CNI's which is why i didnt want to update it at this time, but i can open an issue so that CNI's can be discussed there

hack/multi-node/user-data.sample
rata: :+1:

pkg/asset/internal/templates.go
rmenn: Previously it was only resolving records for service IPs but not pod IPs hence moving it to support all PTR requests

pkg/asset/images.go
rata: Not sure I follow. There are several CNIs already, isn't the plan to just keep them as they are at least until the way to have many is revisited? Why not change the version number, then? Am I missing something?

Also, it would be great if you can create an issue or document in some way what you mean with still discussing with how to enable different CNIs (like what are the problems you see now, what you would like in a solution, or what are you at in the current disucssion)

rmenn: Calico in its canal implementation is still fixed with the `experimental` flag, since the jenkins which ran the e2e tests are no longer present ( they were run by coreos ) we still have to figure out where/how to run the e2e `flannel-e2e`, `calico-e2e` & `conformance` needs to run, it is a part of the roadmap to get e2e running. If its just a version bump+config i guess without e2e, i can send a separate PR for it since its still experimental

rata: Oh, okay. Let me know if I can help :)

pkg/asset/internal/templates.go
rata: Oh, okay. IMHO, it would be nice to be in the commit message (my personal taste is different commits, but that is just me).

Also, just curious: what pod IPs don't resolve? Pod IPs in stateful set?

hack/multi-node/bootkube-up
```diff
@@ -25,7 +25,7 @@ if [ ! -d "cluster" ]; then
   ../../_output/bin/${local_os}/bootkube render --asset-dir=cluster --api-servers=https://172.17.4.101:6443 ${etcd_render_flags} ${cnp_render_flags}
   cp user-data.sample cluster/user-data-worker
   cp user-data.sample cluster/user-data-controller
-  sed -i -e '/node-role.kubernetes.io\/master/d' cluster/user-data-worker
+  sed -i -e '/node.kubernetes.io\/master/d' cluster/user-data-worker
```
andrewrynhard: We shouldn't do this. The `node-role.kubernetes.io` label should be used since `node-role.kubernetes.io/master` is "special" for security reasons. The `node.kubernetes.io` label used for node status/conditions. I know this puts the burden of labeling the joining master nodes on the user, but I don't we can do any better than that without a much bigger change.

pkg/asset/images.go
```diff
@@ -3,11 +3,11 @@ package asset
...
 	CalicoCNI:       "quay.io/calico/cni:v2.0.0",
-	CoreDNS:         "k8s.gcr.io/coredns:1.2.6",
-	Hyperkube:       "k8s.gcr.io/hyperkube:v1.13.3",
+	CoreDNS:         "k8s.gcr.io/coredns:1.6.2",
+	Hyperkube:       "k8s.gcr.io/hyperkube:v1.16.2",
```
andrewrynhard: I feel like the flannel and CoreDNS changes should be in a separate PR since they are not strictly required in the upgrade to  Kubernetes 1.16.

hack/multi-node/bootkube-up
rmenn: so since coreos uses the kubelet to self lebel/taint itself this is no longer possible with the NodeRestriction, im open to suggestions
* How about i leave the `node-role` in and have it removed in the hack/* so that `node-role` remains in render and the hack/* can do what it requires.

rmenn: Based on https://github.com/kubernetes/kubernetes/issues/75457#issuecomment-532609381, we will keep the `node-role` in all the templates and its upto the tooling with uses bootkube to manage the labels, so the `hack/*` will handle the label wrt to this repo

rmenn:
Tested the checkpointer changes as well
```
kube-system   pod-checkpointer-7z2xl                     1/1     Running   0          5m40s
kube-system   pod-checkpointer-7z2xl-172.17.4.101        1/1     Running   0          32s

multi-node|f_update_1.16⚡ ⇒ kubectl --kubeconfig=cluster/auth/kubeconfig get ds pod-checkpointer -o=jsonpath='{$.spec.template.spec.containers[:1].image}' -n kube-system
quay.io/rmenn/checkpoint:caf73a76f659fe867b96d3a46aa1b35b21f10816
```

PR #1061, opened 2019-11-11
update coredns and config for 1.6.2

rmenn:
Fixes : #1059

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -806,13 +806,16 @@ data:
     .:53 {
         errors
         health
-        kubernetes cluster.local {{ .ServiceCIDR }} {
+        ready
+        log . {
+            class error
+        }
+        kubernetes cluster.local in-addr.arpa ip6.arpa {
             pods insecure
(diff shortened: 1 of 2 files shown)


pkg/asset/internal/templates.go
```diff
@@ -806,13 +806,16 @@ data:
     .:53 {
         errors
         health
-        kubernetes cluster.local {{ .ServiceCIDR }} {
+        ready
```
andrewrynhard: Do we need to update the liveness and readiness probes to make use of this?

pkg/asset/internal/templates.go
```diff
@@ -806,13 +806,16 @@ data:
...
+        log . {
+            class error
```
andrewrynhard: Log errors only?

rmenn: yes

pkg/asset/internal/templates.go
rmenn: isnt it better to have that in a separate PR ?

PR #1064, opened 2019-11-16
feat: enable aggregation layer

andrewrynhard:
This enables the aggregation layer according to https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer. This is required in order to create conformant Kubernetes clusters. Without this, sonobuoy fails with:

```
[sig-api-machinery] Aggregator Should be able to support the 1.10 Sample API Server using the current Aggregator [Conformance]
```


pkg/asset/tls.go
```diff
@@ -22,6 +22,40 @@ func newTLSAssets(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey, altNames
...
+	aggregatorCAPrivKey, err := tlsutil.NewPrivateKey()
+	if err != nil {
+		return assets, err
+	}
+
+	cfg := tlsutil.CertConfig{
+		CommonName:         "string",
+		Organization:       []string{},
+		OrganizationalUnit: []string{},
+		AltNames:           tlsutil.AltNames{},
```
andrewrynhard: I might have just missed it in the docs, but looking for guidance here.

rmenn:
@andrewrynhard can we put this behind a flag ? I would like that this is optional and the user knows what is being enabled the default should be that the aggregation be disabled and be rendered when `--enable-aggregation` is passed to render, having it in a flag will ensure the flag can be passed for conformance and it be avoided if the user chooses not to enable it since it increases the security footprint

andrewrynhard:
Being that this is required to pass conformance I feel this should be enabled by default. IMHO bootkube should be conformant out of the box. Would a middle ground be to have a flag and default to true?

rmenn:
My concern is that by enabling aggregation by default is that it enables extensions by default, while as an operator i feel that all extensions should go thru a review process prior to being enabled. The flag will help in keeping the general users of bootkube in a better security scope and the flag will enable bootkube to pass compliance. 

Also depending on further changes in the k/k upstream we can revisit this, we can supply documents presently to this repo with instruction on conformance and replicate the same flag driven behavior in the conformance repo

rmenn:
i have had a chance to look at the usefulness of the aggregation and have changed my stance on this

1 ) all of the other installers have this enabled
2) The metrics server which is an aggregation layer is quite useful and most users will eventually end up using the metrics server along with an hpa or something similar

pkg/asset/tls.go
```diff
@@ -22,6 +22,34 @@ func newTLSAssets(caCert *x509.Certificate, caPrivKey *rsa.PrivateKey, altNames
...
+		return assets, err
+	}
+
+	frontProxyPrivKey, err := tlsutil.NewPrivateKey()
+	if err != nil {
+		return assets, err
+	}
+
+	cfg = tlsutil.CertConfig{
+		CommonName: "front-proxy-client",
```
rmenn: nit : is this a standard naming convention ? `front-proxy-*`

pkg/asset/internal/templates.go
```diff
@@ -222,6 +222,13 @@ spec:
     - --authorization-mode=Node,RBAC
     - --bind-address=0.0.0.0
     - --client-ca-file=/etc/kubernetes/secrets/ca.crt
+    - --requestheader-client-ca-file=/etc/kubernetes/secrets/aggregator.crt
```
rmenn: nit : can this be renamed to `aggregation-ca` ?

rmenn:
Post merge this will require a fix for conformance tests

pkg/asset/tls.go
andrewrynhard: It is. At least in kubeadm.

pkg/asset/internal/templates.go
andrewrynhard: Even better, I aligned with the standard naming convention and called it `front-proxy-ca.crt`.

PR #1066, opened 2019-11-21
enable patching for rendered manifests

rmenn:
The idea is to allow a limited level of patching of the rednered output which is static
There are two levels of users for bootkube
1) Brings their own manifests
2) Uses the manifests available 

This feature is to bridge the gap between both so that if someone needs a change they do not need to go to option 1 but is able to patch the available resource. 

there are examples of patching in the test cases, but more robust examples will be made available , the current code does a simple match with the apiVersion/kind/metadata.name and then proceeds with a normal jsonpatch

### DIFF
--- a/cmd/bootkube/render.go
+++ b/cmd/bootkube/render.go
@@ -47,6 +47,7 @@ var (
 		serviceCIDR         string
 		cloudProvider       string
 		networkProvider     string
+		patchDir            string
 	}
 
 	imageVersions = asset.DefaultImages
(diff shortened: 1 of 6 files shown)


rmenn:
added a mather type to Assets its mainly due to the fact to keep the patches simple an compliant ith what exists, its a simple yaml file which provided in a patch directory combined with the matcher will result in a jsonpactched resource manifest, no validation of a kubernetes manifest will occur

aaronlevy:
Overall I am +1 for the idea of implementing something of a "bootkube delivered base + user provided customizations" model -- but thinking we might get a lot of benefit by using the https://github.com/kubernetes-sigs/kustomize model rather than coming up with our own. Possibly even use the codebase directly. Maybe this has already been discussed?

rmenn:
@aaronlevy the initial implementation was using kustomize but the dependency's that were pulled in was the reason to go a different way

andrewrynhard:
There are larger discussions going on as well. There is a clear split between helm and kustomize in the community. Nothing solid yet, but we were trying to find maybe some kind of plugin mechanism by which people can bring in their own rendering logic. It is a very rough idea.

aaronlevy:
Sounds good.

For what it's worth, this is a big reason for the clear separation between `render` and `start`. The "plugin mechanism" is really just an expected directory structure. Bootkube shouldn't/doesn't care how you generate them. E.g.

```
$ ./my-render-tool --out=./manifests
$./bootkube start --asset-dir=./manifests
```

And another reason for the [design principle](https://github.com/kubernetes-sigs/bootkube/blob/master/Documentation/design-principles.md) that `start` should not require any knowledge/flags that are "known" during any render step (native or completely custom).

My general suggestion would be that we don't invent a new method of rendering. We either output static manifests, or decide that the render step uses another community supported rendering tool/process.

andrewrynhard:
@rmenn with our recent talks about plugins, should we close this?

PR #1068, opened 2019-11-27
fix: remove node selector from static manifests

andrewrynhard:
Due to changes in Kubernetes 1.16, the node-role.kubernetes.io label can
no longer be automatically applied by the kubelet. Due to this, we had
to remove it from the `--node-labels` flag. This caused the static
pod-checkpointer to fail to run after a complete loss of the control
plane (e.g. all control plane nodes rebooted) because it had a
`nodeSelector` inherited from the `DaemonSet` that used the
`node-role.kubernetes.io` label. This PR adds logic to strip off all
`node-role.kubernetes.io` related metadata from the static manifest, which
allows the scheduling to occur.

### DIFF
--- a/pkg/checkpoint/manifest.go
+++ b/pkg/checkpoint/manifest.go
@@ -8,14 +8,14 @@ import (
 	"strings"
 
 	"github.com/golang/glog"
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
 // getFileCheckpoints will retrieve all checkpoint manifests from a given filepath.
-func getFileCheckpoints(path string) map[string]*v1.Pod {
-	checkpoints := make(map[string]*v1.Pod)
+func getFileCheckpoints(path string) map[string]*corev1.Pod {
+	checkpoints := make(map[string]*corev1.Pod)
 
 	fi, err := ioutil.ReadDir(path)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/checkpoint/manifest.go
```diff
@@ -8,7 +8,7 @@ import (
 	"strings"
 
 	"github.com/golang/glog"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
rmenn: can we use `corev1` instead of `v1`. `corev1` is present in other parts of the codebase already and helps with differentiating with `meta/v1` etc

pkg/checkpoint/pod.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strings"
 
 	"github.com/golang/glog"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
rmenn: as above

pkg/checkpoint/manifest.go
andrewrynhard: Done.

pkg/checkpoint/pod.go
```diff
@@ -66,14 +66,17 @@ func sanitizeCheckpointPod(cp *v1.Pod) *v1.Pod {
 	// Remove affinity
 	cp.Spec.Affinity = nil
 
+	// Remove node selector
+	cp.Spec.NodeSelector = nil
+
```
andrewrynhard: This is the fix.

rmenn: whoops i didnt realize the number of changes that would be required.

PR #1069, opened 2019-12-01
Enables Kubelet Auth

rmenn:
* Changes api-server cert org to system:masters
* This allows the api-server to maintain admin access and also ensure that kubectl logs/exec etc remain functional 
* Changes to `hack/*` to move the kubelet from `AllwaysAllow` to `WebHook`authn mode
* Fixes broken `sed` command.

PR #1070, opened 2019-12-01
adds rbac for checkpointer secure client

rmenn:
* Adds manifests for checkpointer cluster role and cluster role binding which is required by the secure client. 

to be merged after #1069


andrewrynhard:
@rmenn I'll leave the `/hold`. Remove once you are ready.

PR #1071, opened 2019-12-02
CoreDNS Updates

rmenn:
- Adds Pod Anti Affinity
- Updates replicas to 2
- Adds lameduck plugin to health ( it allows 5 seconds wait time after its unhealthy to shutdown )
- Fixes Labels kube-dns to coredns
- Adds Readiness Probe
- Updates CoreDNS to 1.6.5

### DIFF
--- a/pkg/asset/internal/templates.go
+++ b/pkg/asset/internal/templates.go
@@ -843,7 +843,9 @@ data:
   Corefile: |
     .:53 {
         errors
-        health
+        health {
+          lameduck 5s
+        }
         ready
         log . {
             class error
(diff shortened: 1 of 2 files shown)


pkg/asset/internal/templates.go
```diff
@@ -871,24 +873,33 @@ metadata:
...
-  # 3. Will be tuned in real time if DNS horizontal auto-scaling is turned on.
+  replicas : 2
   strategy:
     type: RollingUpdate
     rollingUpdate:
       maxUnavailable: 1
   selector:
     matchLabels:
-      k8s-app: kube-dns
+      k8s-app: coredns
```
andrewrynhard: FWIW, https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/addons/dns/manifests.go#L224. 

I think that is to align it with the labels and the kubernetes DNS service. Do we want to stick to the same?

pkg/asset/internal/templates.go
```diff
@@ -871,24 +873,33 @@ metadata:
...
   template:
     metadata:
       labels:
-        k8s-app: kube-dns
+        k8s-app: coredns
       annotations:
         seccomp.security.alpha.kubernetes.io/pod: 'docker/default'
     spec:
+      affinity:
```
andrewrynhard: Should we also look into https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/addons/dns/manifests.go#L239 in this PR? Looks like they are doing a number of things I like for security too: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/addons/dns/manifests.go#L287. We can certainly add those in a follow up.

rmenn: Security Context is already present in the manifest. Priority Classes is on the roadmap, i thought i'd prefer to do that all together for all the manifests

pkg/asset/internal/templates.go
rmenn: Not sure of the upside or downside of keeping it as `kube-dns`

rmenn: looks like it was changed to coredns but got changed back to kubedns in this PR https://github.com/kubernetes-sigs/bootkube/pull/1038 - the particular commit https://github.com/kubernetes-sigs/bootkube/commit/daf844ac8e9aea459688bb9611aaf937fd44c027

andrewrynhard: I have a feeling we should leave as `kube-dns` but we can go with this in the mean time.

PR #1073, opened 2019-12-10
update for kubernetes 1.17

rmenn:
updates kubernetes to 1.17

### DIFF
--- a/pkg/asset/images.go
+++ b/pkg/asset/images.go
@@ -8,6 +8,6 @@ var DefaultImages = ImageVersions{
 	Calico:          "quay.io/calico/node:v3.0.3",
 	CalicoCNI:       "quay.io/calico/cni:v2.0.0",
 	CoreDNS:         "k8s.gcr.io/coredns:1.6.2",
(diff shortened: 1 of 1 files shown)


pkg/asset/images.go
```diff
@@ -8,6 +8,6 @@ var DefaultImages = ImageVersions{
 	Calico:          "quay.io/calico/node:v3.0.3",
```
swade1987: This is quite an old image and idea why we are tracking so far behind?

rmenn: calico updates have not been considered

rata: That is being addressed here: https://github.com/kubernetes-sigs/bootkube/pull/1074 :-D (draft PR for now :))

aaronlevy:
There's a few other places that k8s version usually gets bumped as well. As an example, see: https://github.com/kubernetes-sigs/bootkube/commit/5abe4790c190293759468d6a296dd51bcc9b8dfb

Otherwise all the examples / tests will have mis-matched kubelet vs control-plane versions.

PR #1074, opened 2019-12-11
Update calico to 3.10.2

rata:
This is not ready for review yet, I need to test and split in proper commits to explain the changes.

This is based on templates from https://github.com/kinvolk/terraform-render-bootkube. That project has been used in production and took manifests generated by bootkube initially. Just they were tuned with calico updates there.

Do you think I need to explain here all the changes done through the "years" or just bump it up to match the current version is enough as a start for calico that is supported in a experimental way?

NOTE: This is not updating canal. IIUC, the plan is to remove it, so I'll leave canal as it is.

Fixes: #1058


rata:
I've done some progress. I think one CRD is missing (need to check which one) and see how to handle the IPPool for the pod cidr.

PR #1078, opened 2020-02-13
Remove dghubble from OWNERS list

dghubble:
Hey folks, for my own clusters, I no longer use bootkube and haven't for a while. Its been a good few years, but I think its best I take myself off the list and leave it in the capable hands of those who are actively using it.


andrewrynhard:
Thanks for all the work to get the project to where it is @dghubble!

PR #1083, opened 2020-04-15
Add support for installing helm charts using helm 3 as a library during “bootkube start”

### LINKED ISSUE #1063: Add support for installing helm charts using helm 3 as a library during “bootkube start”
## Description

Helm 3 has been released a few days ago and it removes the requirement of having Tiller running on the cluster. This makes it usable during cluster bootstrapping. It would be great if bootkube would allow to create helm releases from helm charts, in addition to applying static YAML files (manifests). Having bootstrap manifests installed as a helm release allows easy management of further updates, configuration changes, etc.

This could be added as an experimental feature, in addition to existing manifests applying mechanism, so there would be no breaking changes for the existing users.

We, at [Kinvolk](https://github.com/kinvolk) have developed a PoC for this and will be more than happy to create a pull request and act further on this to be integrated seamlessly with upstream bootkube.

## Background:

Currently [lokomotive-kubernetes](https://github.com/kinvolk/lokomotive-kubernetes) uses [terraform-render-bootkube ](https://github.com/kinvolk/terraform-render-bootkube/) terraform module, which renders all bootkube manifests using Terraform `template_dir` resource and copies it to the controller node, which then runs bootkube and applies those manifests.

As we think it is difficult to manage and update the manifests installed in such a way, we are exploring options to use Helm for templating and managing those. With the new Helm 3 release, this became significantly easier, as Tiller is no longer a requirement.

With bootkube supporting installing charts in addition to the manifests, we would install Kubernetes Control Plane, Calico and CoreDNS using helm charts, so we can update them later using regular “helm upgrade” command.

## Benefits 

We believe there exists a mutual synergy in the simplicity of using bootkube to create a self hosted kubernetes cluster and the management of control plane components and applications via helm charts.

- We get all the benefits of using helm to control the releases, even after bootkube has long finished its task of setting up a self hosted cluster.

- Seamless upgrades and downgrades of the control plane components, as well as configuration changes

- Additional dependencies such as CoreDNS, CNI plugins can be deployed as dependent charts, thereby delegating maintenance responsibilities to upstream charts.

While we cannot talk about benefits without neglecting possible complexities, below are the reasons listed as well

- Security stigma attached with Helm, w.r.t to tiller, even though the proposal talks about Helm 3, which does not require it anymore

- Introducing Helm 3 as a library into bootkube, creates a coupling that may not be desired, although this is not a major issue as we remain committed to not change the existing functionality, and that the feature starts its life as an experimental feature.

## Proposed Implementation

1. Execution of the command `bootkube render --asset-dir=<dir-name>` to create an empty directory named `charts` inside the assets directory

2. Bootkube user can add any dependent chart in that directory should it be desired by the user.
The chart can also be a chart of kubernetes control-plane components, in which case the user understands that  `manifests` and `charts` directory should not contain duplicated resources.

3. In `charts` directory, each helm chart will be stored as a directory. The name of the directory will be used as a helm’s release name. For user-supplied values, YAML file in `charts` directory with the same name as the chart directory will be used. Alternatively, the directory could be replaced with the chart tarball.

4. The command `bootkube start --asset-dir=<dir-name>` to apply manifests from `manifests` directory, and then install charts from`charts` directory.

## Further steps

If introduced changes gets positive feedback from the community and maintainers, changing the way `bootkube render` creates assets could be considered. Bootkube could fetch and untar kubernetes helm chart, rather than storing all manifests in go templates. This would move the maintenance cost to chart maintainer (which is undecided yet, bootkube could maintain it’s own version of the chart). However, this is out of scope for this issue.

Cc - @invidian

invidian:
Closes #1063

CC @ipochi @andrewrynhard


andrewrynhard:
Thanks @invidian . As we discussed on slack, we need to figure out how to integrate this with the new plugin model. Give me a bit to work on that interface, and then we will circle back to this once we are ready.

invidian:
/remove-lifecycle stale

I need to rebase...

invidian:
Not stale

mrbobbytables:
This project is being archived, for more information see this issue:
https://github.com/kubernetes/org/issues/2824

/close

PR #1088, opened 2020-04-17
feat: add support for watching node status

andrewrynhard:
This introduces support for waiting for all nodes to report back as Ready.
This fixes an edge case where self hosted pods come up on other control
plane nodes but not on the bootstrap node and all control plane nodes are
pointed to the bootstrap node.

Depends on #1087.


rmenn:
* This does increase the time required for the bootkube start process to exit, it is negligible but i believe it leaves the cluster in a better state since it waits for node status prior to exiting

/lgtm

PR #1090, opened 2020-04-28
refactor: use updated etcd package

andrewrynhard:


andrewrynhard:
@rmenn When you get a chance, PTAL.

PR #1093, opened 2020-05-19
build/build-release.sh improvements and fix failing tests

### LINKED ISSUE #1092: Don't use rkt for building images
https://github.com/rkt/rkt is EoL and it seems `build/build-release.sh` still require it. I think it should be replaced with Docker.

invidian:
This PR is mainly replacing `rkt run` with `docker run` in `build/build-release.sh` to address #1092, as `rkt` is EoL and it is more likely that users will have `docker` installed.

I also introduced bunch of improvements to the script itself, along the way, including updating to the newer Go version.

While verifying the build, I found that many tests are failing because of updated dependencies, so I fixed syntax error there as well.

Closes #1092


invidian:
Can this be merged then?

PR #1097, opened 2020-05-29
pod-checkpointer: allow setting kubelet API address via HOST_IP env variable

ipochi:
Currently pod-checkpointer has hardcoded 127.0.0.1 as IP address of
kubelet API on the node. This means, it will be able to talk to the
kubelet only if the kubelet listens on 0.0.0.0 address, as kubelet must
be exposed to kube-apiserver.

In case where kubelet listens on some specific IP address, e.g. on
node's private IP, pod-checkpointer will no longer be able to talk to
it.

This PR allows setting kubelet IP via HOST_IP environment variable,
which can be set in the pod template using the following snippet:

```yaml
env:
- name: HOST_IP
  valueFrom:
    fieldRef:
      fieldPath: status.hostIP
```

If the environment variable is not defined, it will fallback to use
127.0.0.1, to preserve backward compatibility.

Signed-off-by: Imran Pochi <imran@kinvolk.io>

### DIFF
--- a/pkg/checkpoint/kubelet.go
+++ b/pkg/checkpoint/kubelet.go
@@ -3,6 +3,7 @@ package checkpoint
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


andrewrynhard:
@rmenn Thoughts?

ipochi:
@rmenn @andrewrynhard 
Can we merge this ?

